PR #45, opened 2016-10-19
Move algo finding to cudnnFind*Ex

slayton58:
Move from cudnnFind\* to cudnnFind*Ex for algorithm choosing.

Properly gate finding dgrad algorithm to when dgrad is actually needed to be computed.

This is a) better and b) fixes some potential deadlock issues in multi-GPU runs relating to freeing of device memory.


Yangqing:
Thanks Simon :)

PR #46, opened 2016-10-19
Fix BN in test phase

slayton58:
It seems like this fix got nuked in the last sync - re-issuing


Yangqing:
Oops, I see what you meant now - the Net class is trying to play overly smart by returning a blob reference instead of an iterable. That is probably bad (my fault of course, it makes things inconsistent) but yeah, this should fix it :)

https://github.com/caffe2/caffe2/blob/master/caffe2/python/core.py#L1103

PR #47, opened 2016-10-19
Threaded decode + GPU transform

slayton58:
2 major additions:
1. Multi-threaded JPEG decode + crop is now done using a thread pool
2. Optionally, the NHWC->NCHW transpose, scale, mean subtraction and int8->fp32 conversions can now be performed on the GPU


caffe2/image/image_input_op.h
```diff
@@ -192,80 +364,57 @@ bool ImageInputOp<Context>::Prefetch() {
...
+  if (gpu_transform_) {
+    // we'll transfer up in int8, then convert later
+    prefetched_image_.mutable_data<uint8_t>();
+  } else {
+    prefetched_image_.mutable_data<float>();
+  }
   prefetched_label_.mutable_data<int>();
   // TODO(jiayq): Handle this prefetching with a real thread pool. Currently,
-  // with 4 threads we should be able to get a decent sheed for AlexNet type
+  // with 4 threads we should be able to get a decent speed for AlexNet type
```
Yangqing, line 393: It's sheed :P!

I think we can remove this todo now?

caffe2/image/transform_gpu.cu
```diff
@@ -0,0 +1,81 @@
...
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ **/
```
Yangqing, line 9: I wonder if we can move this to the general LICENSE file so that we avoid putting them in every source file. Is that OK with nv?

caffe2/image/image_input_op.h
slayton58: Ultimate sheed! Yes, I think the TODO can be removed :)

caffe2/image/transform_gpu.cu
slayton58: "Newly-created files **must** have both a copyright and license statement in the header" -- this is a guideline from our team, which is what I followed (It's a copy of the relevant part of the LICENSE file in the project root) -- I'll ask for guidance about if we can rely on the project-wide file / give a link / whatever.

slayton58: Yep, and I have a cleaner solution I'll push in the morning, just needed to clear it :)

PR #48, opened 2016-10-19
Add priority streams for NCCL ops

slayton58:
Helps prevent potential deadlocks in NCCL calls


Yangqing:
Thanks Simon - this seem to fix issues for concurrent kernels. I wonder if we will need a backup for commodity gpus who do not have priority stream :) Also, I am thinking about having a mutex in CUDAContext to guard against alloc / deallocs: nccl op can then lock it to avoid other caffe2 ops to accidentally invoke a sync. Would that be a reasonable idea?

PR #50, opened 2016-10-27
Make bias optional in cuDNN conv op

slayton58:
@Yangqing This seems to work for me, not sure if it's implemented in the right way for you to accept :)

Allows user to specify "no_bias" as an option for convolution layers (only cuDNN at this point), so that the bias associated with that operator is not allocated or computed. This is useful in particular for conv + BatchNorm combinations (such as ResNets), as the bias term can be handled by both conv and Batch Norm, wasting memory and computation.


bwasti:
import

bwasti:
import

Yangqing:
Landed - thanks Simon :)

PR #51, opened 2016-11-22
Top-k accuracy operator on host

slayton58:
Automatically copy from device -> host if necessary.

Thanks to @pooyadavoodi for the host top-k code.


Yangqing:
Looks pretty cool, thanks Simon @slayton58 !

@bwasti Bram - maybe this could be a good pull request to test our sync scripts?

caffe2/operators/accuracy_op.cc
```diff
@@ -14,21 +14,43 @@ bool AccuracyOp<float, CPUContext>::RunOnDevice() {
   DCHECK_EQ(label.dim32(0), N);
   Y->Resize(vector<TIndex>());
   const auto* Xdata = X.data<float>();
-  const auto* labeldata = label.data<int>();
+  const auto* labelData = label.data<int>();
+  const int top_k = top_k_;
```
akyrola: This code depends on top_k <= D. So I think we should add 

top_k = std::min(top_k, D)

akyrola: this local variable is not needed?

caffe2/python/cnn.py
```diff
@@ -521,6 +521,22 @@ def Iter(self, blob_out, **kwargs):
             **kwargs)
         return self.net.Iter(blob_out, blob_out, **kwargs)
 
+    def Accuracy(self, blob_in, blob_out, **kwargs):
```
akyrola: can we move this to model_helper.py instead, as this is not CNN specific function?

caffe2/operators/accuracy_op.cc
Yangqing: Or better, do CAFFE_ENFORCE(top_k <= D, "Hey dude"); to guard against incorrect arguments.

pooyadavoodi: That's right, we can use `top_k_` instead.

pooyadavoodi: Done

pooyadavoodi: Done

caffe2/python/cnn.py
pooyadavoodi: I guess this should be done in different PR?

slayton58: I can do this, but I won't be able to get to it for a little bit.

@pooyadavoodi it should be done in this PR, no sense in adding functionality and then immediately moving it.

slayton58: @akyrola Moved to model_helper.py and rebased from master

PR #52, opened 2016-11-30
Fix incorrect range in CreateCommonWorld

slayton58:
NumInputs should be (0,1) to account for optional first argument. Fixes MPI tests.


Yangqing:
Looks right :)

cc @bwasti for a test on pulling diffs.

PR #54, opened 2016-12-06
Initial CMake building with deps

slayton58:
Based very heavily on Caffe1's cmake build system (and incorporating multiple of their files unchanged for now).

Builds a test binary using shared libs with some external dependency resolution (cuDNN, NCCL etc.)

Pushing so @bwasti and @Yangqing can see + build on it.


neil1899:
hi @slayton58 

when i use your branch for build caffe2, it occurs error either, can't find pthread.h

i just use the command by:
mkdir build & cmake ../ & make

slayton58:
@neil1899 this is a very experimental and currently incomplete branch - it's not expected to work yet.

That said, it shouldn't have any issues finding pthread.h

caffe2/CMakeLists.txt
```diff
@@ -77,11 +77,23 @@ add_subdirectory(proto)
...
+LIST(APPEND CUDA_NVCC_FLAGS -Xcompiler -std=c++11)
+LIST(APPEND CUDA_NVCC_FLAGS -gencode arch=compute_52,compute=sm_52)
+CUDA_ADD_LIBRARY(Caffe2_GPU SHARED ${Caffe2_GPU_SRCS})
+target_link_libraries(Caffe2_GPU Caffe2_PROTO glog gflags)
+
 # Compile test binaries.
 add_executable(Caffe2_CPU_TEST ${Caffe2_CPU_TEST_SRCS})
 
-target_link_libraries(Caffe2_CPU_TEST Caffe2_CPU gtest)
+target_link_libraries(Caffe2_CPU_TEST Caffe2_CPU gtest glog atlas cblas)
```
Yangqing: Could you add a TODO for us to later make the blas backend interchangeable?

caffe2/proto/CMakeLists.txt
```diff
@@ -4,5 +4,6 @@ file(GLOB Caffe2_PROTOBUF_FILES "${CMAKE_CURRENT_SOURCE_DIR}/*.proto")
 
 PROTOBUF_GENERATE_CPP(Caffe2_PROTO_SRCS Caffe2_PROTO_HEADERS ${Caffe2_PROTOBUF_FILES})
 
-add_library(Caffe2_PROTO ${Caffe2_PROTO_HEADERS} ${Caffe2_PROTO_SRCS})
+add_library(Caffe2_PROTO SHARED ${Caffe2_PROTO_HEADERS} ${Caffe2_PROTO_SRCS})
```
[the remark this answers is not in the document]
slayton58: I was wondering about this - I'd normally build everything as shared and link it, but if you'd prefer static I can do that (and I realise that's what you do with BREW)

caffe2/CMakeLists.txt
slayton58: Yup, done locally

slayton58:
Ok, so this is now compiling the vast majority of the project - enough so that I can successfully run a python example.

I was compiling with:

```
cd build;
cmake ../ -DCUDNN_ROOT=<path to cudnn> -DNCCL_ROOT_DIR=<path to NCCL root> -DCMAKE_INSTALL_PREFIX=<path to install to>;
make -j8 install
```

slayton58:
Added different BLAS backends (per @Yangqing request), some refactoring, fixed a couple of issues (not building NCCL correctly when it isn't found in system directories for example)

caffe2/CMakeLists.txt
```diff
@@ -1,41 +1,70 @@
-# Excludes
+# ---[ Excludes
 set(Caffe2_DIR_EXCLUDES
-        "/binaries"
-        "/contrib"
-        "/distributed"
-        "/mpi"
-        "/image"
```
bwasti, line 7: some of the files in here have openMP requirements, breaking the build if you don't exclude them by default. They should only be included if openmp is available.

cmake/Modules/FindOpenBLAS.cmake
```diff
@@ -0,0 +1,64 @@
+
+
+SET(Open_BLAS_INCLUDE_SEARCH_PATHS
+  /usr/include
+  /usr/include/openblas
+  /usr/include/openblas-base
+  /usr/local/include
+  /usr/local/include/openblas
+  /usr/local/include/openblas-base
+  /opt/OpenBLAS/include
```
bwasti, line 10: I'm on OS X (and installed openblas with homebrew).  I had to add

    /usr/local/opt/openblas/include

cmake/Modules/FindOpenBLAS.cmake
```diff
@@ -0,0 +1,64 @@
...
+SET(Open_BLAS_LIB_SEARCH_PATHS
+        /lib/
+        /lib/openblas-base
+        /lib64/
+        /usr/lib
+        /usr/lib/openblas-base
+        /usr/lib64
+        /usr/local/lib
+        /usr/local/lib64
+        /opt/OpenBLAS/lib
```
bwasti, line 24: and

    /usr/local/opt/openblas/lib

bwasti: This probably applies to some other search paths, I'll do a more aggressive test in the future.

caffe2/CMakeLists.txt
```diff
@@ -46,42 +75,93 @@ exclude(Caffe2_DIRS "${Caffe2_DIRS}" ${Caffe2_DIR_EXCLUDES})
...
+if(NOT USE_MPI)
+  message(STATUS "Excluding MPI binaries")
+  prepend(tmp "${CMAKE_CURRENT_SOURCE_DIR}/binaries/" "${Caffe2_MPI_BINARY_SRC}")
+  exclude(Caffe2_BINARY_SRCS "${Caffe2_BINARY_SRCS}" "${tmp}")
+endif()
+
+foreach(binary_src ${Caffe2_BINARY_SRCS})
+  get_filename_component(bin_name ${binary_src} NAME_WE)
+  add_executable(${bin_name} ${binary_src})
+  target_link_libraries(${bin_name} -Wl,--whole-archive ${Caffe2_LINK_LIBS} glog atlas cblas cnmem ${Caffe2_LINKER_LIBS} -Wl,--no-whole-archive)
```
bwasti: `--*whole-archive` is not available in OS X

bwasti: `atlas` and `cnmem` are optional, so they may not even be in this compilation path

slayton58: cnmem is not optional for the GPU path, but it's being refactored now.

cmake/Modules/FindOpenBLAS.cmake
slayton58: This should be handled by setting the OpenBLAS_HOME environment variable outside of cmake.

cmake/Modules/FindOpenBLAS.cmake
slayton58: Again, these should be handled by using existing cmake or environment variables.

caffe2/CMakeLists.txt
slayton58: They shouldn't be excluded wholesale either. If you can give me a list of OMP-requiring files I can build an explicit exclusion list for them

cmake/Modules/FindOpenBLAS.cmake
bwasti: Is there any documentation on how this is handled?  Must be a bug in the package I installed because that environment variable definitely was not set appropriately

slayton58: Not that I've seen - it looks to be up the the provider of the Find* package - in this case from Caffe. Otherwise online docs for built-in packages will document what options can be passed in to suggest paths

bwasti:
Should it require an installed protobuf if the source files for protobuf are in third_party?  Am I misreading this error/cmake files?  I'm testing on a fresh install of OS X right now.

````
CMake Error at /usr/local/Cellar/cmake/3.7.1/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:138 (message):
  Could NOT find Protobuf (missing: Protobuf_LIBRARIES Protobuf_INCLUDE_DIR)
Call Stack (most recent call first):
  /usr/local/Cellar/cmake/3.7.1/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  /usr/local/Cellar/cmake/3.7.1/share/cmake/Modules/FindProtobuf.cmake:393 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  cmake/ProtoBuf.cmake:4 (find_package)
  cmake/Dependencies.cmake:5 (include)
  CMakeLists.txt:44 (include)
````

slayton58:
Hmm, I didn't add that functionality - let me do that now

slayton58:
Took a while longer than expected (cmake's external_project is not particularly pleasant to work with in my opinion), but I've pushed support - based on the USE_SYSTEM_PROTOBUF option.

Not as clean as I'd like, but I (or someone else) can go back and refactor as needed.

bwasti:
looks good!

PR #56, opened 2016-12-15
Gpu transform

slayton58:
Adds a thread pool for image decode, and optional GPU-based data conversion, mean subtraction and std division

### DIFF
--- a/caffe2/image/image_input_op.h
+++ b/caffe2/image/image_input_op.h
@@ -1,7 +1,6 @@
 #ifndef CAFFE2_IMAGE_IMAGE_INPUT_OP_H_
 #define CAFFE2_IMAGE_IMAGE_INPUT_OP_H_
 
-#include <omp.h>
 #include <opencv2/opencv.hpp>
 
 #include <iostream>
@@ -9,6 +8,7 @@
 #include "caffe/proto/caffe.pb.h"
 #include "caffe2/core/db.h"
 #include "caffe2/utils/math.h"
+#include "caffe2/utils/thread_pool.h"
 #include "caffe2/operators/prefetch_op.h"
 
 namespace caffe2 {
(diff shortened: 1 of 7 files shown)


caffe2/image/image_input_op.h
```diff
@@ -192,80 +364,55 @@ bool ImageInputOp<Context>::Prefetch() {
...
   }
-  #pragma omp parallel for num_threads(4)
+
   for (int item_id = 0; item_id < batch_size_; ++item_id) {
     std::bernoulli_distribution mirror_this_image(0.5);
-    std::mt19937& randgen = randgen_per_thread[omp_get_thread_num()];
-    float* image_data = prefetched_image_.mutable_data<float>()
-        + crop_ * crop_ * channels * item_id;
-    string key, value;
+    std::mt19937& randgen = randgen_per_thread[num_decode_threads_];
```
akyrola, line 383: I think here is a bug, since randgen_per_thread has only num_decode_threads_ elements. My code crashes here.

slayton58:
Damnit, we had a race between when the PR was imported and when I found + fixed that issue: https://github.com/slayton58/caffe2/commit/50a8505a6a89a492a96900c54e505245e3bc03f8

caffe2/image/transform_gpu.cu
```diff
@@ -0,0 +1,63 @@
...
+  }
+}
+
+}
+
+template <typename T_IN, typename T_OUT, class Context>
+bool TransformOnGPU(Tensor<Context>& X, Tensor<Context> *Y, T_OUT mean, T_OUT std, Context *context) {};
+
+template <>
+bool TransformOnGPU<uint8_t, float, CUDAContext>(Tensor<CUDAContext>& X, Tensor<CUDAContext> *Y, float std, float mean, CUDAContext *context)
```
Yangqing, line 49: @slayton58 Simon, @dutran caught inconsistency - seems that we are having std and mean instead of the other order round, is this intended? Or is it a bug actually?

slayton58: This was a bug, and was fixed previously (see my comment on 01/09 in fact!)

It was fixed by me post-importing, but pre-merging -- I think that's a corner case for your import routines

Yangqing: T_T

slayton58: This reminds me, need to add a test for this code...

PR #60, opened 2016-12-19
Skip sparse tests if operators not available

slayton58:
Only tests for SparseFunHash for now


Yangqing:
Looks totally reasonable, thanks Simon :)

PR #65, opened 2016-12-20
Delete testFetchFeedBlobTypeAsserts

pooyadavoodi:
Delete testFetchFeedBlobTypeAsserts because FeedBlob now accepts float64 blobs for CUDA.

Note that there is still no CUDA ops for float64.


Yangqing:
(For some reason, this should already have been merged but the fb tool did not close the issue, manually closing. Thanks @pooyadavoodi !

PR #78, opened 2017-01-04
Warn against unsatisfied but optional dependencies

bwasti:


Yangqing:
Oops, my recent cmake changes might have caused a bunch of conflicts, sorry in advance :)

PR #85, opened 2017-01-06
Cudnn v6

slayton58:
Add cudnn v6 support, including testing support for dilated convolution.
Add a check to ensure that the versions of cuDNN used to compile Caffe2 and run it are compatible

### DIFF
--- a/caffe2/operators/conv_op_cudnn.cc
+++ b/caffe2/operators/conv_op_cudnn.cc
@@ -56,9 +56,12 @@ class CudnnConvOpBase : public ConvPoolOpBase<CUDAContext> {
         pad_l_ == pad_r_,
         "The current padding scheme leads to unequal padding on the left "
         "and right, which is not supported by cudnn.");
+    // dilated convolution supported by some algorithms in cuDNN v6
+#if !(CUDNN_VERSION_MIN(6,0,0))
     OPERATOR_NEEDS_FEATURE(
         dilation_h_ == 1 && dilation_w_ == 1,
         "The cudnn convolution does not support dilation yet.");
+#endif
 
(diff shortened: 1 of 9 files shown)


caffe2/operators/conv_op_cudnn.cc
```diff
@@ -270,9 +273,15 @@ bool CudnnConvOp<T>::RunOnDevice() {
...
     // Set the convolution descriptor
+#if CUDNN_VERSION_MIN(6,0,0)
+    CUDNN_CHECK(cudnnSetConvolution2dDescriptor(
+          conv_desc_, pad_t_, pad_l_, stride_h_, stride_w_, dilation_h_, dilation_w_,
+          CUDNN_CROSS_CORRELATION, cudnnTypeWrapper<T>::type));
+#else
     CUDNN_CHECK(cudnnSetConvolution2dDescriptor(
           conv_desc_, pad_t_, pad_l_, stride_h_, stride_w_, 1, 1,
           CUDNN_CROSS_CORRELATION));
+#endif
```
Yangqing: Simon - we don't need this if-else, right? we can simply use dilation_h_ and dilation_w_ because in the case of cudnn5, the OPERATOR_NEEDS_FEATURE would have guaranteed that they are 1.

caffe2/operators/conv_op_cudnn.cc
```diff
@@ -450,9 +459,15 @@ bool CudnnConvGradientOp<T>::RunOnDevice() {
...
```
Yangqing: ditto - see line 284

caffe2/operators/conv_op_cudnn.cc
Yangqing: Oh sorry, scratch that, I did not notice the type input.

PR #92, opened 2017-01-16
[WIP] Param grad coalescing

slayton58:
First look at adding parameter (and parameter gradient) coalescing to reduce the overhead of communication and param updates.

Support added via the core.ParamGroupScope(name) construct, and through the data_parallel_model abstraction class.


Yangqing:
@slayton58 - are we still planning to merge this? Just curious.

slayton58:
@Yangqing Some version of this I intend to merge at some point, but probably not this one. The code is somewhat messy, and I never got the performance boosts I expected.

With shape inference available in python, I feel a graph transform (consume the original graph, output a graph with coalesced parameters + gradients) is a better approach for this, or directly modifying the generated prototxt file.

pietern:
Let's close this since it won't be merged in the current form.

PR #93, opened 2017-01-16
[WIP] Refactor CuDNNReluOp for multi-precision

slayton58:
One way of simplifying the fp16 / multi-precision operators -- remove the explict OpName / OpNameFP16 divide, dispatch the correct calls at runtime based on the contents of the input tensor(s).

### DIFF
--- a/caffe2/operators/pool_op_cudnn.cc
+++ b/caffe2/operators/pool_op_cudnn.cc
@@ -20,6 +20,20 @@ class CuDNNPoolOp : public ConvPoolOpBase<CUDAContext> {
     } else {
       LOG(FATAL) << "Unsupported pooling method: " << def().type();
     }
+
+    // Choose function body for given math type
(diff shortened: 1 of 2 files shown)


Yangqing:
Thanks Simon! There is actually one functionality implemented in Caffe2 exactly for this purpose, called DispatchHelper:

https://github.com/caffe2/caffe2/blob/bea8ed126baa98ddaa313e4d255b9c1000374865/caffe2/core/operator.h#L227

which essentially uses templates to factor out the need of an explicit if loop. For one example, check out the reshape op:

https://github.com/caffe2/caffe2/blob/beb5141f5676e996bf7ef9f26a356beb902e8d71/caffe2/operators/utility_ops.h#L1047-L1062

Essentially, instead of running RunOnDevice you use that as a wrapper and put the real meat in the DoRunWithType() function.

Do you mind refactoring to that pattern?

PR #95, opened 2017-01-17
Prevent concurrent memory and NCCL ops

slayton58:
Use a mutex to prevent simultaneous memory alloc / free and NCCL ops

### DIFF
--- a/caffe2/contrib/nccl/cuda_nccl_gpu.cc
+++ b/caffe2/contrib/nccl/cuda_nccl_gpu.cc
@@ -126,19 +126,23 @@ void runNCCL(const NCCLExecution& ex, F&& f) {
     CUDA_CHECK(cudaEventRecord(context->master_event_, ex.stream));
   }
 
-  for (auto i = 0; i < ex.elements.size(); ++i) {
-    auto& ctx = ex.elements[i];
-    DeviceGuard g(ctx.device);
-    auto& comm = comms[i];
-    auto& stream = streams[i];
-    auto& event = events[i];
+  {
+    // lock out alloc / free while NCCL launches
+    std::lock_guard<std::mutex> lock(CUDAContext::mutex());
+    for (auto i = 0; i < ex.elements.size(); ++i) {
+      auto& ctx = ex.elements[i];
+      DeviceGuard g(ctx.device);
+      auto& comm = comms[i];
+      auto& stream = streams[i];
+      auto& event = events[i];
 
-    DCHECK_EQ(ctx.device, GetGPUIDForPointer(ctx.src->raw_data()));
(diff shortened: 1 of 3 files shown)


caffe2/core/context_gpu.h
```diff
@@ -177,6 +178,10 @@ class CUDAContext final {
 
   static void Delete(void* data);
 
+
+  // Get the read-write mutex
```
akyrola: But this is not really a read-write mutex but a basic mutex?

caffe2/contrib/nccl/cuda_nccl_gpu.cc
```diff
@@ -126,19 +126,23 @@ void runNCCL(const NCCLExecution& ex, F&& f) {
...
-
-    DCHECK_EQ(ctx.device, GetGPUIDForPointer(ctx.src->raw_data()));
-    CUDA_CHECK(cudaStreamWaitEvent(stream, context->master_event_, 0));
-    f(ctx, comm, stream);
-    // Record an event on each children stream that we have finished
-    // our computation
-    CUDA_CHECK(cudaEventRecord(event, stream));
+  {
+    // lock out alloc / free while NCCL launches
+    std::lock_guard<std::mutex> lock(CUDAContext::mutex());
```
akyrola, line 131: I guess read/write lock would be better, so that NCCL could run in parallel? But of course currently we don't run NCCL in parallel (in data_parallel_model).

caffe2/core/context_gpu.cu
```diff
@@ -268,8 +274,13 @@ CUDAContext::CUDAContext(const DeviceOption& option)
   DCHECK_EQ(option.device_type(), CUDA);
 }
 
+std::mutex& CUDAContext::mutex() {
+  return gCUDAContextMutex();
```
Yangqing: Out of curiosity, can we save 3 lines of code by putting gCUDAContextMutex() into this function?

slayton58:
It was a read-write mutex before I realised what versions of c++ they needed :) I'll change the comment

slayton58:
3 lines is 3 lines :) yes I'll do that

slayton58:
PR updated based on comments - thanks @Yangqing and @akyrola!

Note: For posterity, @akyrola is correct, this should really be a reader-write mutex, but unfortunately the mechanisms needed in standard C++ (shared_mutex, shared_lock) won't both be available until C++17. Boost could be used, but that's a heavy dependency to add for just this. The only (potential) downside of this implementation is that NCCL calls will be serialized, but in my experience that's actually best for performance (each NCCL call is designed to maximize the use of available bandwidth, multiple calls just adds contention to the links)

PR #103, opened 2017-01-23
Comment out NHWC Alexnet test for now

slayton58:
Relies on NHWC implementation of group conv which doesn't exist right
now


Yangqing:
Thanks Simon :) Importing.

PR #106, opened 2017-01-25
Remove non-docs items from the documentation branch

JoelMarcey:
Let's not cruft up the documentation branch with non-documentation items.
Also let's move the actual docs to the root of this branch as well.

This will make it easier to get started with incorporating a template. These
files will move according to what the template dictates.


Yangqing:
Sounds good - cc @bwasti for double-check on merging to a non-master branch, I thin this should be safe?

JoelMarcey:
Yeah, we have done direct merges of some of @aaronmarkham's pull requests directly too.

PR #127, opened 2017-02-02
Fix git URL in README

lukeyeager:
URL was changed in https://github.com/caffe2/caffe2/commit/f5169438411737b49265b2e319c2ef588428c46b

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Caffe2 is released under the [BSD 2-Clause license](https://github.com/Yangqing/
 
 [![Build Status](https://travis-ci.org/caffe2/caffe2.svg?branch=master)](https://travis-ci.org/caffe2/caffe2)
 
-    git clone --recursive https://github.com/bwasti/caffe2.git
+    git clone --recursive https://github.com/caffe2/caffe2.git
     cd caffe2
     
 #### OS X


Yangqing:
Thanks Luke! Do you mind filling the CLA so we can pull it? (NV has a corporate CLA with us already FYI)

lukeyeager:
Yeah, we're trying to figure out how to add me to the list...

lukeyeager:
Dang, y'all are really slamming Travis. I wonder if the nvcc-compatible version of ccache that the PyTorch guys wrote would be helpful? https://github.com/pytorch/pytorch/pull/614#issuecomment-275802467

lukeyeager:
> Dang, y'all are really slamming Travis. I wonder if the nvcc-compatible version of ccache that the PyTorch guys wrote would be helpful? pytorch/pytorch#614 (comment)

FWIW, I looked into this briefly. Using ccache ([like this](http://stackoverflow.com/a/24305849/2404152)) doesn't seem to provide any speed-up at all:

Without ccache: `3m13.261s`
Using ccache with a cached build: `3m14.060s` (slower!)

I suspect this is because `caffe.pb.h` gets regenerated every time and most files include it. Oh well.

PR #130, opened 2017-02-04
Use MDB_NOLOCK.

pooyadavoodi:
- Do not lock LMDB.
- This avoids failure when multiple readers try to read the same LMDB.
- This also can cause a race if a process tries to write into the LMDB that is being read by another process. Because this commit removes the locking mechanism. 
- Note that we already use MDB_RDONLY when reading LMDB.
- It seems that LMDB does not provide any method of locking the database to avoid writes while allowing reads.


Yangqing:
Thanks! This sounds much reasonable, I believe MDB write/read concurrency isn't that common in Caffe2 use cases :)

PR #135, opened 2017-02-08
caffe_translator fix: HasField actually throws exceptions.

Yangqing:
@aaronmarkham this should solve your issue.


Yangqing:
Abandoning - will commit internally.

PR #141, opened 2017-02-09
[TravisCI] Don't install the full CUDA toolkit

lukeyeager:
Should speed up the build process slightly since far few packages will be installed.

Matches the Caffe1 builds: https://github.com/BVLC/caffe/blob/rc4/scripts/travis/install-deps.sh#L79-L109

### DIFF
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -59,8 +59,8 @@ else
   #################
 
   # Found here:
-  # https://github.com/NVIDIA/nvidia-docker/blob/master/ubuntu-16.04/cuda/8.0/devel/cudnn5/Dockerfile#L11-L16
-  CUDNN_DOWNLOAD_SUM=a87cb2df2e5e7cc0a05e266734e679ee1a2fadad6f06af82a76ed81a23b102c8
+  # https://gitlab.com/nvidia/cuda/blob/ff2d7c34fe/8.0/devel/cudnn5/Dockerfile
+  CUDNN_DOWNLOAD_SUM=c10719b36f2dd6e9ddc63e3189affaa1a94d7d027e63b71c3f64d449ab0645ce
   CUDNN_URL="http://developer.download.nvidia.com/compute/redist/cudnn/v5.1/cudnn-8.0-linux-x64-v5.1.tgz"
   curl -fsSL ${CUDNN_URL} -O
   echo "$CUDNN_DOWNLOAD_SUM  cudnn-8.0-linux-x64-v5.1.tgz" | sha256sum -c --strict -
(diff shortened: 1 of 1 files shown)


lukeyeager:
Don't merge this yet. When inspecting a local build, it looks like the driver is linked?
```sh
$ ldd install/lib/libCaffe2_GPU.so | grep -i cud
	libcurand.so.8.0 => /usr/local/cuda-8.0/targets/x86_64-linux/lib/libcurand.so.8.0 (0x00007f1954c42000)
	libcublas.so.8.0 => /usr/local/cuda-8.0/targets/x86_64-linux/lib/libcublas.so.8.0 (0x00007f1952292000)
-->     libcuda.so.1 => /usr/lib/x86_64-linux-gnu/libcuda.so.1 (0x00007f19518ac000)
	libnvrtc.so.8.0 => /usr/local/cuda-8.0/targets/x86_64-linux/lib/libnvrtc.so.8.0 (0x00007f195048f000)
	libcudnn.so.5 => /usr/lib/x86_64-linux-gnu/libcudnn.so.5 (0x00007f194b220000)
	libcudart.so.8.0 => /usr/local/cuda-8.0/targets/x86_64-linux/lib/libcudart.so.8.0 (0x00007f1941843000)
	libicudata.so.55 => /usr/lib/x86_64-linux-gnu/libicudata.so.55 (0x00007f1930415000)
```
Does Caffe2 use the driver API in addition to the CUDA API? If so, we may need to install more of the toolkit.

However, [a Travis build](https://travis-ci.org/lukeyeager/caffe2/builds/199832958) passed without installing the driver...

Yangqing:
Hi @lukeyeager - thanks so much for this. Yeah, we are using the driver api in the cuda_rtc path here: 

https://github.com/caffe2/caffe2/blob/master/caffe2/cuda_rtc/common_rtc.h

lukeyeager:
OK, I think that ought to do it. I'm not totally sure whether all of `cuda-drivers` is strictly necessary, but it should be sufficient.

lukeyeager:
(Blocked by #142)

flx42:
`cuda-driver-dev-8-0` is probably what you need, it provides the `libcuda.so` stub at:
```
/usr/local/cuda-8.0/targets/x86_64-linux/lib/stubs/libcuda.so
```

lukeyeager:
@flx42 I was trying to make that work, but it looks like I need libcuda1-367 to get the actual .so. The simplest way I found to install that package was to install the whole cuda-drivers package.
```
$ dpkg -S /usr/lib/x86_64-linux-gnu/libcuda.so.1
libcuda1-375: /usr/lib/x86_64-linux-gnu/libcuda.so.1
```

PR #146, opened 2017-02-13
Windows build and contbuild stuff

Yangqing:
TODO: right now cuda is still not supported, just sending this so we can get Ruben starting to test on Windows more easily.


ezineo:
@Yangqing, do you have built Caffe2_CPU.lib in windows successfully? I failed to build it in VS2015. Is it necessary to add or change something in source code of your windows branch?

Yangqing:
seems that import is failing, I am submitting internally with @bwasti . Closing.

Yangqing:
@ezineo I actually don't - this is actually only WIP and there are a few
Linux-specific things that we need to sort out before complete windows
support is ready.

Yangqing

PR #147, opened 2017-02-13
Added local build and apple fix for generating .so files

bwasti:


Yangqing:
lgtm :)

bwasti:
broke builds :( I'll look into it

cmake/ProtoBuf.cmake
```diff
@@ -15,16 +14,29 @@ if (ANDROID OR IOS OR WIN32)
...
+endfunction()
+
+if (ANDROID OR IOS OR WIN32)
+  custom_protobuf_find()
 else()
-  find_package( Protobuf REQUIRED )
-  list(APPEND Caffe2_DEPENDENCY_LIBS ${PROTOBUF_LIBRARIES})
-  include_directories(SYSTEM ${PROTOBUF_INCLUDE_DIR})
+  find_package( Protobuf )
+  if ( NOT Protobuf_FOUND )
```
Yangqing: @bwasti so this is the issue: in earlier versions of cmake (like 3.0.0) PROTOBUF_FOUND is used, and for some reason in the newest versions cmake started using Protobuf_FOUND, see:

https://github.com/Kitware/CMake/blob/v3.0.0/Modules/FindProtobuf.cmake

vs 

https://github.com/Kitware/CMake/blob/v3.7.2/Modules/FindProtobuf.cmake

and as a result, if we are only checking one, we end up missing the found message, ending up adding the custom protobuf find path.

I honestly think this is a cmake careless mistake, but I guess it means that we will need to guard both PROTOBUF_FOUND and Protobuf_FOUND. Maybe adding more message(WARNING "...") messages could help. My feeling is that the Protobuf package is not found, we are better off sending a warning message to the user saying "this is unusual, we will try to build with a local protobuf, but you should be noted that this path is flaky and you are better off installing your own protobuf and manually makes sure that c++ and python are using the same protobuf version".

Yangqing (changes requested):

Yangqing:
(btw, ignore the appveyor build fail right now, it is still WIP)

PR #148, opened 2017-02-14
Delete SerializeToString() call in class Model(), workspace.py

ezineo:
.In Tutorial, I found it not correct when calling Model(). After that changing, It works.


Yangqing:
Thanks for the fix! cc @aaronmarkham: this might fix the problem as we just chatted 15 mins ago.

PR #150, opened 2017-02-15
Add python predictor

bwasti:
Python wrapped predictor class


Yangqing:
Dang, so the appveyor build is really annoying right now because it keeps showing failure - I'll turn it off until when it succeeds on my local branch.

bwasti:
deprecated, closing

PR #151, opened 2017-02-15
windows build: getting there

Yangqing:
This clears up a bunch of windows build errors, but there are still 12 errors mostly relating to
- template keywords
- initializer list
- pthreadpool

that are not readily available on windows. Also, cuda build is being disabled right now.

Current error can be found here: https://ci.appveyor.com/project/Yangqing/caffe2-w2ucm


Yangqing:
I also found that the exp and log op are not optimized, and this PR fixed it too.

PR #152, opened 2017-02-15
Fix ProtoBuf.cmake to use PROTOBUF_LIBRARY as well

bwasti:


bwasti:
github is somewhat broken
https://patch-diff.githubusercontent.com/raw/caffe2/caffe2/pull/152.patch
is not the same as the actual patch (which is here:
https://patch-diff.githubusercontent.com/raw/caffe2/caffe2/pull/152.diff )
so the import isn't working

PR #156, opened 2017-02-16
Added model downloader (hard coded to point to bwasti/models at the moment)

bwasti:


bwasti:
{PR URL}.patch is broken on this pull request as well, reopening

PR #157, opened 2017-02-17
Update caffe_translator.py

ezineo:
Add "PReLU" TranslatorRegistry in caffe_translator.py


bwasti:
would you mind splitting out the BatchNorm changes into a different pull request?

bwasti (changes requested):
Thanks! Please separate the BatchNorm change into a different pull request.

ezineo:
There are still some problems. The BatchNorm in Caffe must follow with a Scale layer, so that it can be equal to Caffe2's implementation.

ezineo:
OK, I see...

PR #170, opened 2017-02-24
use CMAKE_SYSTEM_NAME instead of LINUX

Yangqing:


Yangqing:
Verified via contbuild - landing.

cc @slayton58 FYI

PR #171, opened 2017-02-25
Add PReLU translator

ezineo:


caffe2/python/caffe_translator.py
```diff
@@ -522,6 +522,16 @@ def TranslateROIPooling(layer, pretrained_blobs, is_test):
     return caffe_op, []
 
 
+@TranslatorRegistry.Register("PReLU")
+def TranslateRelu(layer, pretrained_blobs, is_test):
```
Yangqing: Change name to

def TranslatePRelu

?

ezineo: Sry, it's a mistake.

Yangqing:
Merged - thanks so much!

PR #184, opened 2017-03-06
Add _aligned_free if defined _MSC_VER in context.h

ezineo:
In windows, it is necessary to use `_aligned_free` instead of `free` when using `_aligned_malloc` before.


Yangqing:
Thanks! Travis build failure should have been handled in #185 .

PR #187, opened 2017-03-07
Bugfix: type not being set when inferring types+shapes

lukeyeager:
/cc @akyrola 

I basically just copied all the `ShapeCall` stuff as `TypeCall`. Is there a better way?

### DIFF
--- a/caffe2/core/context_gpu.cu
+++ b/caffe2/core/context_gpu.cu
@@ -136,6 +136,11 @@ static void Caffe2InitializeCuda() {
   // Restore the current device.
   CUDA_ENFORCE(cudaSetDevice(init_device));
 
+  RegisterTypeCallFunction(
+    TypeMeta::Id<Tensor<CUDAContext>>(),
+    GetTensorType<CUDAContext>
+  );
+
   RegisterShapeCallFunction(
     TypeMeta::Id<Tensor<CUDAContext>>(),
     GetTensorShape<CUDAContext>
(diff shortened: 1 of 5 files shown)


caffe2/core/context_gpu.cu
```diff
@@ -136,6 +136,11 @@ static void Caffe2InitializeCuda() {
   // Restore the current device.
   CUDA_ENFORCE(cudaSetDevice(init_device));
 
+  RegisterTypeCallFunction(
```
akyrola, line 139: We need to do the same to Tensor<CPUContext>? I think RegisterShapeCallFunction for TensorCPU is done in tensor.cc

lukeyeager: It's in the constructor for the registry

akyrola: Oh yeah, sorry!

akyrola (approved):

lukeyeager:
@Yangqing GitHub reports a merge conflict, but @facebook-github-bot says you already imported it. I'm not sure what it means when you import a PR, exactly. Do you need me to rebase this?

Yangqing:
Thanks @lukeyeager - yeah, if you don't mind I'll just merge it internally, and also resolve some of our internal tests that depends on the test interface change. does that work?

lukeyeager:
Ah got it. Didn't realize I was breaking your stuff. Ok, I'll leave this un-rebased.

Yangqing:
Actually, @lukeyeager , my apologies, I mistook this for #199 - could you rebase?

(that PR is landing in a minute btw)

PR #190, opened 2017-03-07
Updated and new tutorials

aaronmarkham:
Many changes and additions, the major one removing the MNIST database so we're not pushing a large binary file.


aaronmarkham:
landing problems.... trying another route.

PR #191, opened 2017-03-08
Fix binary file reading bug for MSC compiler

ezineo:
For MSC compiler binary flag needs to be specified


Yangqing:
Thanks for the catch!

PR #193, opened 2017-03-09
[tests] CastOp shape inference

lukeyeager:
Tests all types which are valid in numpy and which don't currently throw errors.


lukeyeager:
Trying to test the full matrix of CastOp ins and outs fails because I need the fix in https://github.com/caffe2/caffe2/pull/187. I'll just close this and add this extra test to that PR.

PR #196, opened 2017-03-13
fix minor typo in math_{cpu.cc,gpu.cu}

thinxer:


thinxer:
CLA signed.

Yangqing:
Thanks so much for the catch :)

PR #199, opened 2017-03-14
Allow test discovery in caffe2/python/

lukeyeager:
These are all essentially no-op changes which allow for nose-style (or pytest-style) test discovery.

With this patch, you can use any of these methods to discover and run tests under `caffe2/python`:
```
python -m unittest discover caffe2/python/ '*test*.py'
python -m nose caffe2/python/
python -m pytest caffe2/python/
```

Future work:

* Get all of the tests to pass
  * Some seem to be testing operations which don't have GPU implementations
  * I get a segfault unless I set `CUDA_VISIBLE_DEVICES=0`
  * Some tests are flaky
* Allow test discovery throughout the whole project (e.g. the `experiments/` dir)

### DIFF
--- a/caffe2/python/caffe_translator_test.py
+++ b/caffe2/python/caffe_translator_test.py
@@ -16,8 +16,6 @@ import unittest
 @unittest.skipIf(not os.path.exists('data/testdata/caffe_translator'),
                  'No testdata existing for the caffe translator test. Exiting.')
 def setUpModule():
-    raise RuntimeError
-
     # We will do all the computation stuff in the global space.
     caffenet = caffe_pb2.NetParameter()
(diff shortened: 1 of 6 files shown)


caffe2/python/caffe_translator_test.py
```diff
@@ -13,6 +13,39 @@
 import unittest
 
 
+@unittest.skipIf(not os.path.exists('data/testdata/caffe_translator'),
+                 'No testdata existing for the caffe translator test. Exiting.')
+def setUpModule():
+    raise RuntimeError
```
Yangqing: Is this intended?

lukeyeager: Oops, no.

Issue #194, opened 2017-03-12
build_ios.sh script failing


nicodjimenez:
I successfully installed caffe2 on my unbuntu machine and on my macbook! However ./scripts/build_ios.sh is not working for me.  I get the following error:

```
In file included from /Users/nicojimenez/git/caffe2/caffe2/image/image_input_op.cc:1:
/Users/nicojimenez/git/caffe2/caffe2/image/image_input_op.h:470:21: error: no member named 'runTask'
      in 'caffe2::ThreadPool'
      thread_pool_->runTask(std::bind(
      ~~~~~~~~~~~~  ^
/Users/nicojimenez/git/caffe2/caffe2/image/image_input_op.h:482:21: error: no member named 'runTask'
      in 'caffe2::ThreadPool'
      thread_pool_->runTask(std::bind(
      ~~~~~~~~~~~~  ^
/Users/nicojimenez/git/caffe2/caffe2/image/image_input_op.h:493:17: error: no member named
      'waitWorkComplete' in 'caffe2::ThreadPool'
  thread_pool_->waitWorkComplete();
  ~~~~~~~~~~~~  ^
3 errors generated.
```

What's happening is that the ThreadPool class from ThreadPool.h is being used instead of the intended ThreadPool class from utils/thread_pool.h.  I was able to resolve the build error by changing variable names to remove the duplication (renamed ThreadPool to TaskThreadPool in utils/thread_pool.h).  Happy to make a PR if this is an acceptable solution.

Yangqing:
Thanks @nicodjimenez - please send a PR at your convenience.

Yangqing:
PR merged - thanks for the fix!

PR #200, opened 2017-03-14
Fix build_ios.sh bug (#194) due to name collision.

nicodjimenez:


Yangqing:
Thanks so much!

cc @wickedfoo - Jeff, would this affect your local changes?

wickedfoo:
@Yangqing This should be fine, if I attempt to compile again at some point, the compiler will complain and I can fix.

wickedfoo:
This doesn't look like the threadpool I wrote in any case?

Yangqing:
@wickedfoo Ah yeah, you are right, this is from @bwasti and is probably a contribution from our friend @slayton58 :) Looks safe.

slayton58:
Yes, this looks like mine :) It's used to provide threaded decode / transform in the ImageInputOp

PR #202, opened 2017-03-15
Explicitly pass CXX to NCCL Makefile

pietern:
Necessary if CXX isn't set when cmake is called. The CXX variable will then be
empty which prevents make from using its own default.


Yangqing:
cc Mr @slayton58 by the way - in case NV finds errors in other nccl clients.

slayton58:
@Yangqing Thanks! Good catch @pietern!

PR #208, opened 2017-03-17
CaffeMaxPoolOp

slayton58:


Yangqing:
@slayton58 Simon, quick check - would you like to merge this PR?

(If so, do you want to rename it MaxPoolWithIndex instead of CaffeMaxPool?)

slayton58:
@Yangqing No, this was just there so you guys could look internally :) I do want to merge a version of this at some point, so I'll rename the Op before I do that.

PR #210, opened 2017-03-19
[build] bugfix for Windows, esp. VS 2017

Yangqing:
@aaronmarkham this solves your Windows build issue. Basically:

(1) VS 2017 does not have CUDA support yet, and we will be waiting on NVidia to do so.

(2) VS 2015 and 2017 need different cmake generator strings.

This PR shows how to determine those and also updates appveyor to do contbuild guard for the following 3 settings:
- VS2015 without cuda
- VS2017 without cuda
- VS2015 with cuda

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,12 +4,27 @@ environment:
   matrix:
     - USE_CUDA: OFF
       CMAKE_BUILD_TYPE: Release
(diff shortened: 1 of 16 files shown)


Yangqing:
cc @thatguymike FYI - this is not high priority, but I wonder if NV will support VS2017 soon :)

PR #212, opened 2017-03-21
Use version-specific CUDA packages 16.04

pietern:
So you can just run `BUILD_CUDA=ON .travis/install.sh` on a 16.04 machine and have it install the right packages.


Yangqing:
Sounds great :)

PR #215, opened 2017-03-22
[iOS] Metal backend

ajtulloch:
Putting this up here while we figure out the right syncing and building approach. No need to merge.


Yangqing:
(closing since we just need a link address :))

cancan101:
The [docs state](https://caffe2.ai/docs/mobile-integration.html) :
>surprisingly, ARM CPUs outperform the on-board GPUs (our NNPACK ARM CPU implementation outperforms Apple’s MPSCNNConvolution on devices older than iPhone 6s

Is this benchmark published anywhere?

PR #218, opened 2017-03-23
CMake support for Gloo dependency

pietern:
This also requires a change to cmake/External/nccl.cmake to use the
static NCCL binary instead of the shared object. When the Caffe2/Gloo
build uses the bundled NCCL version it should be packaged up in the
resulting libraries and not cause another runtime dependency on a
library that has to be installed separately.


Yangqing:
@slayton58 Simon, does this work for you? I'll land if this makes sense.

slayton58:
Yes, this should work for us -- it would work the same as NCCL does now

Yangqing (approved):

PR #221, opened 2017-03-24
mnist dataset downloader and squeezenet model loader fixes

aaronmarkham:


caffe2/python/tutorials/MNIST.ipynb
```diff
@@ -90,7 +90,7 @@
     "if not os.path.exists(data_folder):\n",
     "    os.makedirs(data_folder)\n",
     "if not os.path.exists(label_file_train):\n",
-    "    DownloadDataset(\"https://s3.amazonaws.com/caffe2/datasets/mnist/mnist.zip\", data_folder)\n",
+    "    DownloadDataset(\"https://fb.me/caffe2-datasets-mnist.zip\", data_folder)\n",
```
JoelMarcey, line 93: Are we changing our storage location for everything?

aaronmarkham: Yes, the fb url seemed to work for a while but then started throwing ssl errors, so I just switched to using the s3. Also Git LFS was really limiting on how you could access the files, so we'll migrate over to s3.

PR #225, opened 2017-03-28
Fix test_random_seed_behavior for multi-GPU

lukeyeager:
```
E0327 17:33:12.775998 15629 context_gpu.h:126] Encountered CUDA error: an illegal memory access was encountered
F0327 17:33:12.776208 15629 operator.h:176] Computation on device returned error in operator
output: "Y" name: "" type: "XavierFill" arg { name: "shape" ints: 2 } device_option { device_type: 1 cuda_gpu_id: 0 }
```


lukeyeager:
To test before and after:
```
python -m unittest caffe2.python.hypothesis_test.TestOperators.test_random_seed_behaviour
```

PR #227, opened 2017-03-28
[tests] Fixes for ops without a CUDA backend

lukeyeager:
All of these tests fail with some variant of `Cannot create operator of type 'X' on the device 'CUDA'` (see commit messages).


dzhulgakov:
I wonder whether we leverage unittest.skipTest to do it intelligently. I.e. silence the error if an operator is not implemented on gpu without the need for manual modification.

lukeyeager:
If we threw something other than a `RuntimeError`, we might be able to turn a certain exception class into a skip when running tests?

PR #238, opened 2017-03-31
Fix protobuf build to properly include directories

Yangqing:
@aaronmarkham - I think this should fix the oss build.
@slayton58 FYI
@kmatzen FYI


Yangqing:
cc @ajtulloch - this fixes the build.

PR #241, opened 2017-04-03
Add fp16 dispatch to main cuDNN operators

slayton58:
Adds support for fp16 to main cuDNN ops (conv, relu, pool, BatchNorm).

Done via. runtime dispatch, not using DispatchHelper at this point to allow for more complex dispatch logic in the future if necessary. Using separate template  for all input / output types is for these reasons also - it's easier to add the functionality now and never use it, than need to add it later.

### DIFF
--- a/caffe2/operators/conv_op_cudnn.cc
+++ b/caffe2/operators/conv_op_cudnn.cc
@@ -420,8 +420,9 @@ bool CudnnConvOp::RunOnDevice() {
                          // float16,      // Math
                          float16>();   // Y
   } else {
-    CAFFE_THROW("Unsupported inputs");
+    LOG(FATAL) << "Unsupported type inputs";
   }
+  return true;
 }
 
 template <typename T_X, typename T_DY, typename T_W, typename T_B,
@@ -753,8 +754,9 @@ bool CudnnConvGradientOp::RunOnDevice() {
                          float16,    // dW
                          float16>(); // db
   } else {
-    CAFFE_THROW("Unsupported inputs");
+    LOG(FATAL) << "Unsupported input types";
   }
+  return true;
 }
 
 REGISTER_CUDNN_OPERATOR(Conv, CudnnConvOp);
--- a/caffe2/operators/conv_op_cudnn.cc
+++ b/caffe2/operators/conv_op_cudnn.cc
@@ -417,7 +417,6 @@ bool CudnnConvOp::RunOnDevice() {
                          float16,      // W
                          float16,      // B
                          float,      // Math
-                         // float16,      // Math
                          float16>();   // Y
   } else {
     LOG(FATAL) << "Unsupported type inputs";
@@ -749,7 +748,6 @@ bool CudnnConvGradientOp::RunOnDevice() {
                          float16,    //  W
                          float16,    //  b
                          float,    // Math
-                         // float16,    // Math
                          float16,    // dX
                          float16,    // dW
                          float16>(); // db
--- a/caffe2/operators/conv_op_cudnn.cc
+++ b/caffe2/operators/conv_op_cudnn.cc
@@ -140,13 +140,16 @@ class CudnnConvOpBase : public ConvPoolOpBase<CUDAContext> {
   size_t cudnn_state_;
 };
 
-template <typename T>
+
 class CudnnConvOp final : public CudnnConvOpBase {
  public:
   CudnnConvOp(const OperatorDef& operator_def, Workspace* ws)
       : CudnnConvOpBase(operator_def, ws)  {}
 
   ~CudnnConvOp() {}
+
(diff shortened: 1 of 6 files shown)


Yangqing:
This is awesome, thanks @slayton58 !

(I am tagging a 0.6.0 version for pre-release but I am sure we will have it in by mid April.

caffe2/operators/conv_op_cudnn.cc
```diff
@@ -386,21 +393,41 @@ bool CudnnConvOp<T>::RunOnDevice() {
...
+    return DoRunWithType<float16,      // X
+                         float16,      // W
+                         float16,      // B
+                         float,      // Math
+                         // float16,      // Math
+                         float16>();   // Y
+  } else {
+    CAFFE_THROW("Unsupported inputs");
+  }
+}
```
asaadaldien: That breaks the build. add return true here.

asaadaldien: Please apply this to all Run** methods

caffe2/operators/relu_op_cudnn.cc
```diff
@@ -205,7 +162,12 @@ class CuDNNReluGradientOp final : public Operator<CUDAContext> {
     auto* dX = Output(0);
     dX->ResizeLike(Y);
 
-    return (this->*body_)();
+    if (Y.IsType<float>()) {
+      return DoRunWithType<float,float>();
+    } else if (Y.IsType<float16>()) {
+      return DoRunWithType<float16,float>();
+    }
+    // return (this->*body_)();
```
asaadaldien: else LOG(FATAL) ? Also please return true at the end.

asaadaldien (changes requested):

slayton58:
@asaadaldien Done and updated

asaadaldien (changes requested):
Looks good to me Thanks @slayton58 for working on this. few nits.

caffe2/operators/relu_op.cc
```diff
@@ -29,8 +29,9 @@ bool ReluOp<float, CPUContext>::RunOnDevice() {
 
 template <>
 bool ReluGradientOp<float, CPUContext>::RunOnDevice() {
-  auto& Y = Input(0);
-  auto& dY = Input(1);
+  auto& X = Input(0);
```
asaadaldien: Remove unused variable X

caffe2/operators/conv_op_cudnn.cc
```diff
@@ -648,80 +682,84 @@ bool CudnnConvGradientOp<T>::RunOnDevice() {
...
   }
-};
-REGISTER_GRADIENT(ConvFp16, GetConvFp16Gradient);
+  else if (Input(0).IsType<float16>()) {
+    return DoRunWithType<float16,    //  X
+                         float16,    // dY
+                         float16,    //  W
+                         float16,    //  b
+                         float,    // Math
+                         // float16,    // Math
```
asaadaldien: Remove this line

caffe2/operators/relu_op.cc
slayton58: X is used. Strictly, Relu gradient requires X, Y, dY

caffe2/operators/conv_op_cudnn.cc
slayton58: Done

caffe2/operators/relu_op.cc
asaadaldien: Relu gradient needs only Y and dY also the variable X isn't  used in the code section bellow :)

slayton58: Ok, I'll be very explicit here. Relu gradient *should* be using X, the definition of it requires X. However, your CPU implementation doesn't, so I'll comment out this line, although it'll look weird starting with Input(1).

ezineo: Sry, it's my fault. `Xvec` bellow actually means `dXvec`. Do you mind to fix it?

Yangqing:
Simon, a quick question - this PR actually changes the signature of the ReluGradient function by *requiring* the input for the backward pass. In this case we won't be able to do in-place computation and old protos won't be able to run. Is it OK for us to revert to the old implementation where only the output is being provided for the backward pass? I think in any case, the relu backward pass should not need to use the input.

Yangqing:
@slayton58

slayton58:
Sure - I originally added it as an error checking for cuDNN functions (as cudnnActivationBackwards takes X, Y, dY, but it doesn't seem to be 100% required

Yangqing:
Thanks @slayton58 for checking :) The fix is committed in https://github.com/caffe2/caffe2/commit/4a1713b833b9cd719b477bbea77d16b0c86a3fac and a unit test is in place to guard the numerical correctness.

PR #242, opened 2017-04-03
Conv no dx

slayton58:
Based on a discussion with @Yangqing, optionally disables the calculation of dX for a convolution op (i.e. conv1 in Alexnet), where the data gradient is not needed.


Yangqing:
I assume we might need to change the other ConvGradient implementations too?

slayton58:
From my quick check, they all had the logic to do this inside the op, the definition coming in from the net just had to be adjusted -- I could run C2's (non-cuDNN) conv implementation with this change without issue.

PR #246, opened 2017-04-03
fp16 support for FullyConnected op

slayton58:
Includes math lib support, removal of double-precision.


caffe2/operators/utility_ops_gpu.cc
```diff
@@ -5,6 +5,28 @@
...
+
+template <>
+bool SumOp<CUDAContext>::RunOnDevice() {
+  if (Input(0).IsType<float>()) {
+    return DoRunWithType<float,float>();
+  } else if (Input(0).IsType<float16>()) {
+    return DoRunWithType<float16,float16>();
+  } else {
+    CAFFE_THROW("Unsupported inputs");
+  }
```
asaadaldien, line 27: return false;

asaadaldien (commented):
Thanks @slayton58 for working on this. There is some conflicts. Can you please rebase?

slayton58:
rebased + return false changes made

asaadaldien (commented):
LGTM! Thanks @slayton58 . few nits. can you also rebase before I import this :).

caffe2/utils/conversions.h
```diff
@@ -0,0 +1,187 @@
...
+#else
+#define CONVERSIONS_DECL inline
+#endif
+
+namespace caffe2 {
+
+namespace convert {
+
+namespace {
+inline float16 cpu_float2half_rn(float f)
```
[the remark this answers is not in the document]
slayton58: Fixed just now - I rebased just before your comment, so I hope that's still good :)

caffe2/operators/elementwise_op.cu
```diff
@@ -222,6 +220,56 @@ bool SumReduceLikeOp<CUDAContext>::DoRunWithType() {
...
 }
 
+namespace {
+
 REGISTER_CUDA_OPERATOR(SumReduceLike, SumReduceLikeOp<CUDAContext>);
 
+template <typename T, typename M>
+__global__
+void binary_add_kernel(const int N, const T* a, const T* b, T* r) {
+  for (int idx=threadIdx.x+blockIdx.x*blockDim.x; idx < N; idx+=blockDim.x*gridDim.x) {
```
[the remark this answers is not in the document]
slayton58: Missed this before - fixed

asaadaldien (commented):
@slayton58 
Can you please update the signature for Gemm & Gemv
Here:
caffe2/caffe2/utils/math_cpu.cc:261
caffe2/caffe2/utils/math_cpu.cc:292
caffe2/caffe2/contrib/nervana/nervana_math_gpu.cc:13 

Also update the call to FullyConnectedOp here
caffe2/caffe2/contrib/nervana/nervana_fc_op_gpu.cc:8

caffe2/operators/fully_connected_op_gpu.cc
```diff
@@ -2,9 +2,51 @@
...
+                         float,    // Y
+                         float>(); // Math
+  } else if (Input(0).IsType<float16>()) {
+    return DoRunWithType<float16,    // X
+                         float16,    // W
+                         float16,    // B
+                         float16,    // Y
+                         float>(); // Math
+  }
+}
```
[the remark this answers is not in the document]
asaadaldien: else {
CAFFE_THROW("Unsupported type");
return false;
}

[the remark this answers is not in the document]
slayton58: I assume the nervana change was meant to be caffe2/contrib/nervana/nervana_math_gpu.cc

Changes made.

caffe2/utils/conversions.h
```diff
@@ -0,0 +1,185 @@
...
+
+namespace caffe2 {
+
+namespace convert {
+
+namespace {
+inline float16 cpu_float2half_rn(float f) {
+    float16 ret;
+
+    unsigned x = *((int*)(void*)(&f));
```
Yangqing: Just for safety... could you do

    static_assert(sizeof(unsigned int) == sizeof(float),
        "Programming error: sizeof(unsigned int) != sizeof(float).");
    unsigned x = *(reinterpret_cast<unsigned int*>(&f));

asaadaldien: Unfortunately both `*((int*)(void*)(&f))` and `*(reinterpret_cast<unsigned int*>(&f))` will break builds with -fstrict-aliasing (error: dereferencing type-punned pointer will break strict-aliasing rules)
To fix that we can just use the pointer itself: 
`unsigned int* x = reinterpret_cast<unsigned int*>(&f);`
 and in line 101 as well.

slayton58: I don't understand how this is meant to work if I can't dereference that pointer -- I need the actual value of x as originally done.

asaadaldien: @slayton58 You can dereference the pointer x later but the compiler with -fstrict-aliasing
is not happy with this `*(reinterpret_cast<unsigned int*>(&f))`

slayton58: ok, so I can do `unsigned *xp = reinterpret_cast<unsigned int*>(&f); unsigned x = *xp;`

slayton58:
Updated: Note that I've reverted the dgemv commit, if we need to support double precision (I was under the impression we no longer did) then there's more work to be done in terms of function signatures.

caffe2/utils/conversions.h
asaadaldien: Yes that should work. You can do this or dereference x in line 33.

dragonxlwang:
Hey, am wondering what's the reason behind the decision to remove double-precision?

slayton58:
@dragonxlwang It complicates the API and no-one uses double precision.

PR #247, opened 2017-04-04
Add Load, Save functions to CNNModelHelper

lukeyeager:
I'm trying to make some examples more succinct, and I think something like these CNNModelHelper functions will be helpful.

* Do you like the approach of adding `model.solverstate_params`?
* Would you prefer another name to "solverstate" (borrowed from C1)?
* Is there anything besides ITER, LR and momentum that needs to be saved for "solver state"?
  * I expect things may need to be added on a case-by-case basis, as with `params` and `computed_params`

### DIFF
--- a/caffe2/python/cnn.py
+++ b/caffe2/python/cnn.py
@@ -5,7 +5,7 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-from caffe2.python import core, scope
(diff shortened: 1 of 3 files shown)


lukeyeager:
Closing due to merge conflicts and because this isn't a great approach anyway. Still curious if anyone has any comments or suggestions for improvements in this direction.

PR #248, opened 2017-04-04
Add Adagrad with update params on GPU.

pooyadavoodi:
- AdagradOp was supported on GPU but only for computing gradients (without updating params). This PR adds updating parameters on GPU.
- SparseAdagradOp is still only supported on CPU.

### DIFF
--- a/caffe2/sgd/adagrad_op.h
+++ b/caffe2/sgd/adagrad_op.h
@@ -32,12 +32,12 @@ void adagrad_compute(
     float* nw,
     float* nh,
     float epsilon,
-    float lr,
+    const float* lr,
     Context* context) {
   for (auto i = 0; i < N; ++i) {
     float gi = g[i];
     float hi = nh[i] = h[i] + gi * gi;
-    nw[i] = w[i] + lr * gi / (std::sqrt(hi) + epsilon);
+    nw[i] = w[i] + lr[0] * gi / (std::sqrt(hi) + epsilon);
   }
 }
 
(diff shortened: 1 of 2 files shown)


caffe2/sgd/adagrad_op.h
```diff
@@ -32,12 +32,12 @@ void adagrad_compute(
     float* nw,
     float* nh,
     float epsilon,
-    float lr,
+    const float* lr,
```
salexspb, line 35: so why do you have to make it a const float* instead of float?

caffe2/sgd/adagrad_op_gpu.cu
```diff
@@ -36,6 +36,40 @@ void adagrad_update<CUDAContext>(
...
 }
 
+__global__ void AdagradCompute(
+    int N,
+    const float* w,
+    const float* g,
+    const float* h,
+    float* nw,
+    float* nh,
+    float epsilon,
```
salexspb, line 46: normally outputs passed via  a pointer go in the end (according to google code style which we generally use)

pooyadavoodi: @salexspb Do you mean the order of the arguments, i.e. the output arguments should appear at the end of the list?

caffe2/sgd/adagrad_op.h
pooyadavoodi: Just because `RunOnDevice` calls `adagrad_compute` for both CPU and CUDA context. In case of CUDA context, we can't dereference the device pointer on the host. So we need to either send a pointer to `adagrad_compute` or first do a `cudaMemcpy` to copy the content of LR to the host. I picked the first option because `cudaMemcpy` is expensive.

pooyadavoodi:
I need to rebase the branch because it has conflicts.

caffe2/sgd/adagrad_op_gpu.cu
salexspb: Yeah, thats the google code style for c++ functions.

pooyadavoodi:
Closing this PR because it's already implemented by https://github.com/caffe2/caffe2/pull/254.

PR #249, opened 2017-04-04
Fix number of indices and block_size in SparseAdam 

pooyadavoodi:
Fix number of indices and block_size in SparseAdam to support gradients of any dimension.


lukeyeager:
To test, you could refactor this:
https://github.com/caffe2/caffe2/blob/v0.7.0/caffe2/python/operator_test/adam_test.py#L83

To be more like this (from https://github.com/caffe2/caffe2/pull/254):
https://github.com/caffe2/caffe2/blob/v0.7.0/caffe2/python/operator_test/momentum_sgd_test.py#L74-L79

PR #250, opened 2017-04-04
ReversePackedSegsOp optimized GPU code

slayton58:
Removes the need for all the Copy calls, in one of our apps reduced time from ~40ms to < 200us

### DIFF
--- a/caffe2/operators/reverse_packed_segs_op.cu
+++ b/caffe2/operators/reverse_packed_segs_op.cu
@@ -2,6 +2,91 @@
 #include "reverse_packed_segs_op.h"
 
 namespace caffe2 {
+
+namespace {
+
+template <typename T, typename LengthType>
+__global__
+void ReversePackedSegments_kernel(
+      size_t max_length,
+      size_t batch_size,
+      size_t block_size,
+      const LengthType* lengths_ptr,
+      const T* data_ptr,
+      T* rev_data_ptr) {
+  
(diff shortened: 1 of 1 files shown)


caffe2/operators/reverse_packed_segs_op.cu
```diff
@@ -2,6 +2,91 @@
...
+template <typename T, typename LengthType>
+__global__
+void ReversePackedSegments_kernel(
+      size_t max_length,
+      size_t batch_size,
+      size_t block_size,
+      const LengthType* lengths_ptr,
+      const T* data_ptr,
+      T* rev_data_ptr) {
+
```
pietern: Mega nit, trailing whitespace. Internal linter trips over this. Could you configure your editor to automatically strip this?

slayton58: Sure, I'll look into that (I've had it setup on vim before) - I'll update the PR in a few minutes

akyrola (approved):
Looks good to me, assuming you run the unit test for it:

caffe2/python/operator_test/sequence_ops_test.py

caffe2/operators/reverse_packed_segs_op.cu
slayton58: Took longer than expected, but updated

slayton58:
And yes, I ran this through the unit test

PR #251, opened 2017-04-04
Support MomentumSGD, Adagrad, and Adam in Seq2Seq

pooyadavoodi:
- Previously only regular SGD was supported through WeightedSum and ScatterWeightedSum.
- This PR relies on PR https://github.com/caffe2/caffe2/pull/248 and https://github.com/caffe2/caffe2/pull/249 to work.

### DIFF
--- a/caffe2/python/examples/seq2seq.py
+++ b/caffe2/python/examples/seq2seq.py
@@ -579,41 +579,114 @@ class Seq2SeqModelCaffe2:
     def _apply_norm_ratio(
         self, norm_ratio, model, params, learning_rate, scope, ONE
     ):
-        for param in params:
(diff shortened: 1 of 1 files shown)


jawaid2018:
25.04.2017, 00:39, "lem14789" <notifications@github.com>:@lem14789 approved this pull request.

salexspb:
Have you considered using these? https://github.com/caffe2/caffe2/blob/master/caffe2/python/optimizer.py

Also creating of the optimizer could be abstracted out. You could add a method CreateOptimizer to optimizers.py which takes a string and creates an optimizer for you.

pooyadavoodi:
@salexspb I agree. It's better to use optimizer.py. I close the PR for now.

PR #253, opened 2017-04-05
Add __builtin_cpu_supports function def in windows

ezineo:


Yangqing:
Thanks! Right now we don't have a windows verification for avx and non-avx paths, so I'll just blindly merge it :)

PR #254, opened 2017-04-06
Add tests and GPU impls for sparse optimizers

lukeyeager:
These GPU paths are probably even buggier than the CPU paths for sparse gradients with duplicate indices. Both paths cause multiple momentum updates in a single iteration, but only the GPU path is non-deterministic. Depending on how we decide to address the issues on the CPU path, @pooyadavoodi has a good idea for how to match dense behavior with the sparse GPU ops.

### DIFF
--- a/caffe2/python/operator_test/adagrad_test.py
+++ b/caffe2/python/operator_test/adagrad_test.py
@@ -5,7 +5,7 @@ from __future__ import unicode_literals
 
 import functools
 
-from hypothesis import given
+from hypothesis import given, strategies as st
 import numpy as np
 
 from caffe2.python import core
@@ -21,14 +21,15 @@ class TestAdagrad(hu.HypothesisTestCase):
         param_out = param_in + grad_adj
         return (param_out, mom_out)
 
-    @given(inputs=hu.tensors(n=3), **hu.gcs)
-    def test_adagrad(self, inputs, gc, dc):
-        # Inputs
+    @given(inputs=hu.tensors(n=3),
+           lr=st.floats(min_value=0.01, max_value=0.99,
+                        allow_nan=False, allow_infinity=False),
+           epsilon=st.floats(min_value=0.01, max_value=0.99,
+                             allow_nan=False, allow_infinity=False),
+           **hu.gcs)
+    def test_adagrad(self, inputs, lr, epsilon, gc, dc):
         param, momentum, grad = inputs
-        lr = np.random.rand(1).astype(np.float32)
-
-        # Op parameters
-        epsilon = np.random.rand(1).astype(np.float32)[0]
+        lr = np.array([lr], dtype=np.float32)
 
         op = core.CreateOperator(
             "Adagrad",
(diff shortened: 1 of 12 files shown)


caffe2/python/operator_test/adagrad_test.py
```diff
@@ -0,0 +1,77 @@
...
+        return (param_out, mom_out)
+
+    @given(inputs=hu.tensors(n=3), **hu.gcs)
+    def test_adagrad(self, inputs, gc, dc):
+        # Inputs
+        param, momentum, grad = inputs
+        lr = np.random.rand(1).astype(np.float32)
+
+        # Op parameters
+        epsilon = np.random.rand(1).astype(np.float32)[0]
```
dzhulgakov: Please make them arguments of the function with hypothesis inputs. Using randomness inside the test confuses hypothesis and makes it run forever

caffe2/sgd/adagrad_op.h
```diff
@@ -81,6 +62,26 @@ class SparseAdagradOp final : public Operator<Context> {
         epsilon_(OperatorBase::GetSingleArgument<float>("epsilon", 1e-5f)) {}
 
   bool RunOnDevice() override {
+    // Enforce types
+    CAFFE_ENFORCE(OperatorBase::InputIsType<Tensor<Context>>(PARAM));
+    CAFFE_ENFORCE(OperatorBase::InputIsType<Tensor<Context>>(MOMENT_1));
+    CAFFE_ENFORCE(OperatorBase::InputIsType<Tensor<Context>>(INDICES));
+    CAFFE_ENFORCE(OperatorBase::InputIsType<Tensor<Context>>(GRAD));
+    CAFFE_ENFORCE(OperatorBase::InputIsType<Tensor<Context>>(LR));
```
dzhulgakov: no need to do those explicitly, error message would be generated anyway (and there was another PR recently that adds the input name to error message)

caffe2/sgd/adagrad_op.h
```diff
@@ -81,6 +62,26 @@ class SparseAdagradOp final : public Operator<Context> {
...
+
+    // Enforce shapes
+    CAFFE_ENFORCE(Input(PARAM).size() == Input(MOMENT_1).size());
+    CAFFE_ENFORCE(Input(LR).size() == 1);
+    CAFFE_ENFORCE(Input(PARAM).size_from_dim(1) ==
```
dzhulgakov: please use CAFFE_ENFORCE_EQ in these comparisons

caffe2/sgd/adagrad_op.h
```diff
@@ -81,6 +62,26 @@ class SparseAdagradOp final : public Operator<Context> {
...
+            Input(GRAD).size_from_dim(Input(INDICES).ndim()));
+
+    // These must be in-place for the sparse op. If out-of-place is required,
+    // we need to copy input to output before running.
+    CAFFE_ENFORCE_EQ(&Input(PARAM), Output(OUTPUT_PARAM),
```
dzhulgakov: you can use `EnforceInplace` in the OpSchema definition in .cc file. Then you don't really need this. Also you can remove ResizeLike in the DoRunWithType in this case.

caffe2/sgd/adam_op.h
```diff
@@ -115,6 +115,31 @@ class SparseAdamOp final : public Operator<Context> {
         epsilon_(OperatorBase::GetSingleArgument<float>("epsilon", 1e-5f)) {}
 
   bool RunOnDevice() override {
+    // Enforce types
```
dzhulgakov: similar comments to SparseAdaGrad

caffe2/sgd/momentum_sgd_op_gpu.cu
```diff
@@ -57,9 +57,72 @@ void momentum_sgd_update<CUDAContext>(
...
+    const float *lr)
+{
+  const float LR = lr[0];
+  CUDA_1D_KERNEL_LOOP(i, N)
+  {
+    const size_t gradIdx = i;
+    const SIndex index = indices[i / sz];
+    const size_t paramIdx = index * sz + (i % sz);
+
+    if (!nesterov)
```
dzhulgakov, line 81: is it better to have two separate kernels instead?

dzhulgakov (changes requested):
Looks good overall, some minor comments. Thanks!

caffe2/python/operator_test/adagrad_test.py
lukeyeager: Good point. I had copied this from some other test :wink:

caffe2/sgd/adagrad_op.h
lukeyeager: Ok

caffe2/sgd/adagrad_op.h
lukeyeager: Ok

caffe2/sgd/adagrad_op.h
lukeyeager: Neat!

caffe2/sgd/momentum_sgd_op_gpu.cu
lukeyeager: No - there's no perf drawback to this approach

PR #268, opened 2017-04-16
Cmake mobile build improvements

Yangqing:
(1) integrate gcc compatible nnpack
(2) speed up the ios travis ci.


cmake/External/nnpack.cmake
```diff
@@ -35,9 +35,89 @@ endif()
 # (2) Mobile platform - direct build
 ##############################################################################
 
+# We know that NNPACK is only supported with clang due to some builtin
```
Yangqing, line 38: (this part is for old nnpack clang version. After verifying that gcc works we will remove this.)

PR #269, opened 2017-04-17
Mkl changes

rgomathi:
All MKL related files are added here after incorporating review comments, major changes include
1. usage of Clang-format(Linter) with --style = Google
2. usage of macros for checking input and filter dimension in the mkl operators
3. merged Max and Average pooling functions
4. created a new folder for mkl related python scripts in Python folder and moved them there
5. there is no mkl_alexnet_test.py as that was redundant when convnet_benchmark.py does the same thing


Yangqing:
@githubGomathi - I assume that this is now obsolete since #270 is a newer version?

rgomathi:
Yes Yangqing…

Thanks
Gomathi

PR #293, opened 2017-04-19
Add six to Tegra X1 install script

HBadertscher:
When compiling Caffe2 on a Jetson TX2 using JetPack 3.0, the compilation with the Tegra X1 build script runs through perfectly fine. However, when running

    from caffe2.python import workspace

the following error shows up:

> ImportError: No module named six

After installing `six` manually using

    sudo pip install six

this works fine. I thus added the `six` module to the install script.

I assume this will also be required for the `build_raspbian.sh` script, however as I could test this, I didn't add it (yet).


Yangqing:
Thanks so much for the fix!

PR #296, opened 2017-04-19
[FIX] build error in context_gpu_test.cc

farrell236:
caffe2/caffe2/core/context_gpu_test.cc:97:31: error: implicit instantiation of undefined template 'std::__1::array<CUstream_st *, 2>' std::array<cudaStream_t, 2> temp = {0};

(fixes build issue on macOS 10.11.6)


Yangqing:
Thanks for the fix!

PR #320, opened 2017-04-20
Add BatchNorm translator for caffe implementation

ezineo:
This is not an elegent implement, but it works.

### DIFF
--- a/caffe2/python/caffe_translator.py
+++ b/caffe2/python/caffe_translator.py
@@ -92,36 +92,50 @@ class TranslatorRegistry(object):
                 'only accepts new style layers that are stored in the '
                 'layer field.'
             )
-        for layer in caffe_net.layer:
-            if not _ShouldInclude(net_state, layer):
+        i = 0
+        pretrained_layers = [l for l in pretrained_net.layer] +\
+                            [l for l in pretrained_net.layers]
+        while i < len(caffe_net.layer):
+            if not _ShouldInclude(net_state, caffe_net.layer[i]):
                 log.info('Current net state does not need layer {}'
-                            .format(layer.name))
+                            .format(caffe_net.layer[i].name))
                 continue
-            log.info('Translate layer {}'.format(layer.name))
+            log.info('Translate layer {}'.format(caffe_net.layer[i].name))
             # Get pretrained one
-            pretrained_layers = (
-                [l for l in pretrained_net.layer
-                 if l.name == layer.name] + [l
-                                             for l in pretrained_net.layers
-                                             if l.name == layer.name]
-            )
-            if len(pretrained_layers) > 1:
+            layers_index = [l for l in xrange(len(pretrained_layers))
+                 if pretrained_layers[l].name == caffe_net.layer[i].name]
+            is_bn = False
+            if len(layers_index) > 1:
                 raise ValueError(
                     'huh? more than one pretrained layer of one name?')
-            elif len(pretrained_layers) == 1:
-                pretrained_blobs = [
-                    utils.CaffeBlobToNumpyArray(blob)
-                    for blob in pretrained_layers[0].blobs
-                ]
+            elif len(layers_index) == 1:
+                if pretrained_layers[layers_index[0]].type == "BatchNorm":
+                    # A Scale layer should follow BatchNorm layer
+                    # according to paper https://arxiv.org/abs/1502.03167.
+                    assert pretrained_layers[layers_index[0]+1].type == "Scale"
+                    pretrained_blobs = [utils.CaffeBlobToNumpyArray(blob)
+                    for blob in pretrained_layers[layers_index[0] + 1].blobs] +\
+                        [utils.CaffeBlobToNumpyArray(blob)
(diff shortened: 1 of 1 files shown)


caffe2/python/caffe_translator.py
```diff
@@ -92,36 +92,50 @@ def TranslateModel(
...
-            elif len(pretrained_layers) == 1:
-                pretrained_blobs = [
-                    utils.CaffeBlobToNumpyArray(blob)
-                    for blob in pretrained_layers[0].blobs
-                ]
+            elif len(layers_index) == 1:
+                if pretrained_layers[layers_index[0]].type == "BatchNorm":
+                    # A Scale layer should follow BatchNorm layer
+                    # according to paper https://arxiv.org/abs/1502.03167.
+                    assert pretrained_layers[layers_index[0]+1].type == "Scale"
```
cbalint13, line 115: Not necessary to be followed by ```Scale``` layer. See e.g.: https://github.com/BVLC/caffe/blob/master/examples/cifar10/cifar10_full_sigmoid_train_test_bn.prototxt . I believe its user responsibility to add (or not) ```Scale```. I personaly would prefer a tanslator that translates 1:1 original prototxt from old caffe.
 
  * I believe BatchNorm and Scale would be translated as separate things (just as they exists or not).

caffe2/python/caffe_translator.py
```diff
@@ -92,36 +92,50 @@ def TranslateModel(
                 'only accepts new style layers that are stored in the '
                 'layer field.'
             )
-        for layer in caffe_net.layer:
-            if not _ShouldInclude(net_state, layer):
+        i = 0
+        pretrained_layers = [l for l in pretrained_net.layer] +\
+                            [l for l in pretrained_net.layers]
+        while i < len(caffe_net.layer):
```
cbalint13, line 98: ```for i in range(0, len(caffe_net.layer))``` perhaps is better. I don't got why original ``` for layer in caffe_net.layer:``` wasn't good enough for this iteration. I believe you wanted assert that layer+1 is a ```Scale``` layer, but i would't enforce that (see previous example).

caffe2/python/caffe_translator.py
milewis1: Actually, this is correct. In Caffe the BatchNorm layer can't handle a bias, where as in Caffe2, the bias is required for the SpatialBN operator. Also, since Caffe2's Scale operator can't handle a bias we can only do the conversion if a Scale layer with a bias term follows the batch norm layer.

milewis1:
While this change to the translator, allows the translator to succeed, the translated networks will not load due to a protobuf violation.

bandabouse:
@ezineo 
Thanks for add BatchNorm to caffe_translate. But it doesn't work for me when I tried to translate my .prototxt and caffemodel to caffe2. here is the prompt:
INFO:caffe_translator:Translate layer conv1/7x7_s2
INFO:caffe_translator:Translate layer conv1/relu_7x7
INFO:caffe_translator:Translate layer pool1/3x3_s2
INFO:caffe_translator:Translate layer batchNorm-1
Traceback (most recent call last):
  File "/usr/lib64/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib64/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "caffe2/build/caffe2/python/caffe_translator.py", line 646, in <module>
    caffenet, caffenet_pretrained, is_test=True, input_mean=input_mean
  File "caffe2/build/caffe2/python/caffe_translator.py", line 157, in TranslateModel
    return TranslatorRegistry.TranslateModel(*args, **kwargs)
  File "caffe2/build/caffe2/python/caffe_translator.py", line 128, in TranslateModel
    assert pretrained_net.layer[pretrained_layers_index[0]+1].type == "Scale"
AssertionError

And when I remove the Assert, as @cbalint13 said, it doesn't work either
INFO:caffe_translator:Translate layer batchNorm-1
Traceback (most recent call last):
  File "/usr/lib64/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib64/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "caffe2/build/caffe2/python/caffe_translator.py", line 646, in <module>
    caffenet, caffenet_pretrained, is_test=True, input_mean=input_mean
  File "caffe2/build/caffe2/python/caffe_translator.py", line 157, in TranslateModel
    return TranslatorRegistry.TranslateModel(*args, **kwargs)
  File "caffe2/build/caffe2/python/caffe_translator.py", line 146, in TranslateModel
    caffe_net.layer[i], pretrained_blobs, is_test)
  File "caffe2/build/caffe2/python/caffe_translator.py", line 65, in TranslateLayer
    layer, pretrained_blobs, is_test)
  File "caffe2/build/caffe2/python/caffe_translator.py", line 580, in TranslateBatchNorm
    bias = utils.NumpyArrayToCaffe2Tensor(pretrained_blobs[3], output + '_bias')
IndexError: list index out of range

PR #344, opened 2017-04-20
Request for adding new face CNNs to wiki page

iacopomasi:
added PAMs, ResFace-101 and Resnet-101 for regressing 3DMMs


Yangqing:
Wow this is nice - thanks so much!

iacopomasi:
@Yangqing Thank you so much for adding this!

I just figured out that the caffe icon is not visible properly [here](https://caffe2.ai/docs/zoo.html) Can you replace it with the correct link below? (I don't want to open another pull request).

[../static/images/download-c1.png](../static/images/download-c1.png)

Another possibility is that I get access to write the page, then I won't bother you anymore.

Thank you!

PR #361, opened 2017-04-21
Added Python_ADDITIONAL_VERSIONS to cmake so python 2 is default.

kochie:
When installing on systems such as Arch Linux where the default python version is 3 the build will fail. To fix this instead of changing the python link in the shell it is more efficient to set the default python version allowed by cmake.

### DIFF
--- a/cmake/Dependencies.cmake
+++ b/cmake/Dependencies.cmake
@@ -182,6 +182,7 @@ add_definitions(-DEIGEN_MPL2_ONLY)
 
 # ---[ Python + Numpy
 if (BUILD_PYTHON)
+  set(Python_ADDITIONAL_VERSIONS 2.8 2.7 2.6)
   find_package(PythonInterp 2.7)
   find_package(PythonLibs 2.7)
   find_package(NumPy REQUIRED)


Yangqing:
Oh I did not know about this cmake functionality, thanks for finding this!

hgaiser:
I'm not sure this is the best way of resolving this issue. Addtionally, if I use this on Arch it still finds python3 binaries and libraries. Currently I am building with this command:

```
cmake .. \
	-DCUDA_HOST_COMPILER=/usr/bin/g++-5 \
	-DCUDA_PROPAGATE_HOST_FLAGS=off \
	-DUSE_NNPACK=off \
	-DPYTHON_EXECUTABLE:FILEPATH=/usr/bin/python2.7 \
	-DPYTHON_LIBRARY=/usr/lib/libpython2.7.so \
	-DPYTHON_INCLUDE_DIR=/usr/include/python2.7 \
	-DNCCL_ROOT_DIR=/opt/cuda/
```

ps. I'll try soon to see how far I'll get with getting Caffe2 to work with python3.. that'd be the real fix for this.

PR #381, opened 2017-04-22
Fix not showing histogram in Basics.ipynb

jpowie01:
Histogram of Gaussian random variables doesn't show while running "Basics.ipynb" iPython notebook.

There are two ways to resolve this issue:
1. Add "%matplotlib inline" after importing matplotlib.
2. Or just add "pyplot.show()" to show this plot.

I prefer the first solution, so I've prepared a change for that. If you prefer the second one - just give me a feedback and I'll update this PR.


aaronmarkham:
Looks good.

aaronmarkham (approved):

PR #382, opened 2017-04-22
Check system dependencies first

hgaiser:
This PR changes the cmake of Caffe2 to look for system dependencies before resorting to the submodules in `third-party`. Only googletest should logically be in third-party, the other libraries should ideally be installed as system dependencies by the user. This PR adds system dependency checks for Gloo, CUB, pybind11, Eigen and benchmark, as these were missing from the cmake files.

In addition it removes the execution of `git submodule update --init` in cmake. This seems like bad behavior to me, it should be up to the user to download submodules and manage the git repository.


Yangqing:
This looks really nice - yeah, we did a best effort bases to produce a stable cmake build system, but would definitely love the community to double check this with us.

As you mentioned in https://github.com/facebookincubator/gloo/pull/30 - once the CLA gets sorted out we'll merge the PR.

Yangqing:
(respectfully sending back to your queue for CLA)

hgaiser:
> This looks really nice - yeah, we did a best effort bases to produce a stable cmake build system, but would definitely love the community to double check this with us.
> 
> As you mentioned in facebookincubator/gloo#30 - once the CLA gets sorted out we'll merge the PR.

CLA should be signed now.

I noticed I had a copy-paste error in the CMake modules so I force pushed over the commits. For some reason github shows there are 6 commits now, but my git history shows 3 (which is correct). Something to keep an eye on when merging perhaps.

Also as a note, FindGloo.cmake searches for `gloo_cuda`. Should we split this so that FindGloo.cmake still works if CUDA is not enabled? What is the convention with a situation like this?

Yangqing:
Yeah, I think the commits should be fine - when we pull we do an extensive testing (including numerical tests) and combine the commits into one, so I'll monitor the correctness.

RE FindGloo - I think the best is also to guard USE_CUDA, and then search for gloo if cuda is not used?

hgaiser:
> Yeah, I think the commits should be fine - when we pull we do an extensive testing (including numerical tests) and combine the commits into one, so I'll monitor the correctness.
> 
> RE FindGloo - I think the best is also to guard USE_CUDA, and then search for gloo if cuda is not used?

Meaning a FindGloo.cmake and a FindGlooCUDA.cmake ?

EDIT: I think adding a GLOO_USE_CUDA variable would make most sense. If that is set to true, FindGloo.cmake will search for `gloo_cuda`, if set to false it will search for `gloo`, if not set it will try `gloo_cuda` and fall back to `gloo`. I will work on this soon, I'll let you know when I committed those changes.

hgaiser:
I updated the PR to use a variable called `Gloo_USE_CUDA`. If set it will enforce that setting. If not set it will search for `gloo_cuda` and fall back to `gloo` if the cuda version is not found.

hgaiser:
@Yangqing as far as I'm concerned this one is good to merge.

Yangqing:
Sounds great :) I am back from several weeks of other duties, mind doing another rebase?

Yangqing:
Actually the conflict seems minimal so I went ahead and resolved it - will wait for contbuild and then merge when I am back in office.

hgaiser:
@Yangqing I rebased again because your rebase was incorrect, it re-added the `git submodule update --init` command.

ps. out of curiosity, how could you push to the fork? I couldn't find out how you had rights to push there ;)

Yangqing:
Ah, it seems that if you edit it in the browser that gets pushed to your fork... Maybe it is a github backdoor?

Yangqing:
Build error should have been addressed by #640

hgaiser:
> Ah, it seems that if you edit it in the browser that gets pushed to your fork... Maybe it is a github backdoor?

Ah, weird github feature if you ask me :)

PR #396, opened 2017-04-24
New cudnn ops

slayton58:
cuDNN versions of dropout and LRN (for native fp16 support), port of Caffe's max pooling algo that uses an explicit mask to store locations (also supports fp16 storage)

### DIFF
--- a/caffe2/operators/dropout_op_cudnn.cc
+++ b/caffe2/operators/dropout_op_cudnn.cc
@@ -1,0 +1,234 @@
+#include "caffe2/core/common_cudnn.h"
+#include "caffe2/core/context_gpu.h"
+#include "caffe2/core/operator.h"
+#include "caffe2/core/types.h"
+
+namespace caffe2 {
+
+class CuDNNDropoutOp final : public Operator<CUDAContext> {
+ public:
+  USE_OPERATOR_FUNCTIONS(CUDAContext);
+
+  CuDNNDropoutOp(const OperatorDef& operator_def, Workspace* ws)
+      : Operator<CUDAContext>(operator_def, ws),
+        cudnn_wrapper_(&context_),
+        ratio_(OperatorBase::GetSingleArgument<float>("ratio", 0.5)),
+        is_test_(OperatorBase::GetSingleArgument<int>("is_test", 0)) {
+    DCHECK_GE(ratio_, 0);
+    DCHECK_LT(ratio_, 1);
+    CUDNN_ENFORCE(cudnnCreateTensorDescriptor(&data_desc_));
+
+    CUDNN_ENFORCE(cudnnCreateDropoutDescriptor(&dropout_desc_));
+    CUDNN_ENFORCE(cudnnDropoutGetStatesSize(cudnn_wrapper_.inline_cudnn_handle(), &states_size_in_bytes_));
+  }
+
+  ~CuDNNDropoutOp() {
+    CUDNN_ENFORCE(cudnnDestroyTensorDescriptor(data_desc_));
+    CUDNN_ENFORCE(cudnnDestroyDropoutDescriptor(dropout_desc_));
+  }
+
+  template <typename T, typename M>
+  bool DoRunWithType();
+
+  bool RunOnDevice() override;
+
+ protected:
+  CuDNNWrapper cudnn_wrapper_;
+  cudnnTensorDescriptor_t data_desc_;
+  cudnnDropoutDescriptor_t dropout_desc_;
+
+  vector<TIndex> cudnn_input_dims_;
+
+  float ratio_;
+  bool is_test_;
+
+  size_t states_size_in_bytes_, reserve_space_size_in_bytes_;
+  // Input: X, Output: Y, mask, states
+};
+
+class CuDNNDropoutGradientOp final : public Operator<CUDAContext> {
+ public:
+  USE_OPERATOR_FUNCTIONS(CUDAContext);
+  CuDNNDropoutGradientOp(const OperatorDef& operator_def, Workspace* ws)
+      : Operator<CUDAContext>(operator_def, ws),
+        cudnn_wrapper_(&context_),
+        ratio_(OperatorBase::GetSingleArgument<float>("ratio", 0.5)),
+        is_test_(OperatorBase::GetSingleArgument<int>("is_test", 0)) {
+    DCHECK_GE(ratio_, 0);
+    DCHECK_LT(ratio_, 1);
+    CUDNN_ENFORCE(cudnnCreateTensorDescriptor(&data_desc_));
+
+    CUDNN_ENFORCE(cudnnCreateDropoutDescriptor(&dropout_desc_));
+    CUDNN_ENFORCE(cudnnDropoutGetStatesSize(cudnn_wrapper_.inline_cudnn_handle(), &states_size_in_bytes_));
+  }
+
+  ~CuDNNDropoutGradientOp() {
+    CUDNN_ENFORCE(cudnnDestroyTensorDescriptor(data_desc_));
+  }
+
+  template <typename T, typename M>
+  bool DoRunWithType();
+
+  bool RunOnDevice() override;
+
+ protected:
+  CuDNNWrapper cudnn_wrapper_;
+  cudnnTensorDescriptor_t data_desc_;
+  cudnnDropoutDescriptor_t dropout_desc_;
+
+  vector<TIndex> cudnn_input_dims_;
+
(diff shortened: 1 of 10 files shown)


caffe2/operators/max_pool_with_index.cu
```diff
@@ -0,0 +1,197 @@
...
+                       0,
+                       context_.cuda_stream()>>>(
+      X.size(), dY.data<T>(), mask.data<int>(), X.dim32(0), X.dim32(1), X.dim32(2), X.dim32(3),
+      dY.dim32(2), dY.dim32(3), kernel_h(), kernel_w(), stride_h(), stride_w(),
+      pad_t(), pad_l(), dX->template mutable_data<T>());
+  return true;
+}
+
+bool MaxPoolWithIndexGradientOp::RunOnDevice() {
+  auto& X = Input(0);
```
Yangqing, line 145: Simon, since now the pooling op supports arbitrary dimensions but this is still just 2d, would you mind adding an enforce here to make sure that X is 4-dimensional?

caffe2/operators/max_pool_with_index.h
```diff
@@ -0,0 +1,48 @@
+#include <cfloat>
```
Yangqing, line 1: So... I assume there is no CPU implementation for this op? It is probably fine since we don't really need to do maxpool with index if we do training (which is most likely going to be GPUs), but I wonder if this is something we should have for completeness.

Yangqing:
@harouwu could you take a look at the python interface? There seems to be a conflict, and you are refactoring the python side so please check.

harouwu:
@Yangqing I checked it and found that the `brew` renaming has not synced to github. So the name is still model_helpers now. That is the main reason for conflict

harouwu:
My suggestion is rebase after github is synced. But I am not very familiar with syncing process. Do you think it is feasible? @Yangqing @slayton58

slayton58:
I can certainly rebase - it was a clean merge as of when I submitted yesterday, so these conflicts have come in overnight :(

Let me know when master is ready and I'll rebase.

caffe2/operators/max_pool_with_index.h
slayton58: No, I didn't port the CPU implementation over (mainly because I didn't need it)

caffe2/operators/max_pool_with_index.cu
slayton58: Doing this now - I'll push as soon as the test compile is done

slayton58:
Rebased and should now use the brew interface

caffe2/operators/dropout_op_cudnn.cc
```diff
@@ -0,0 +1,234 @@
...
+          cudnnTypeWrapper<T>::type,
+          size_prod,
+          1,
+          1,
+          1));
+    // get the reserve space we need
+    CUDNN_ENFORCE(cudnnDropoutGetReserveSpaceSize(
+          data_desc_, &reserve_space_size_in_bytes_));
+    // store both reserve and states in the same tensor
+    vector<int> state_size{(reserve_space_size_in_bytes_ + states_size_in_bytes_ + sizeof(T)) / sizeof(T)};
```
asaadaldien: @slayton58 This is breaking windows build. (conversion from 'std::size_t' to 'int' requires a narrowing conversion) can you change the type of reserve_space_size_in_bytes_ & states_size_in_bytes_ to int ?

asaadaldien (changes requested):

caffe2/operators/dropout_op_cudnn.cc
slayton58: Done, and rebased again to be sure :)

caffe2/operators/dropout_op_cudnn.cc
```diff
@@ -0,0 +1,234 @@
...
+          data_desc_,
+          GetCudnnTensorFormat(StorageOrder::NCHW),
+          cudnnTypeWrapper<T>::type,
+          size_prod,
+          1,
+          1,
+          1));
+    // get the reserve space we need
+    CUDNN_ENFORCE(cudnnDropoutGetReserveSpaceSize(
+          data_desc_, &reserve_space_size_in_bytes_));
```
asaadaldien: `reinterpret_cast<size_t*>(&reserve_space_size_in_bytes_)`

caffe2/operators/dropout_op_cudnn.cc
```diff
@@ -0,0 +1,234 @@
...
```
asaadaldien: reinterpret_cast<size_t*>(&reserve_space_size_in_bytes_)

caffe2/operators/dropout_op_cudnn.cc
```diff
@@ -0,0 +1,234 @@
...
+      : Operator<CUDAContext>(operator_def, ws),
+        cudnn_wrapper_(&context_),
+        ratio_(OperatorBase::GetSingleArgument<float>("ratio", 0.5)),
+        is_test_(OperatorBase::GetSingleArgument<int>("is_test", 0)) {
+    DCHECK_GE(ratio_, 0);
+    DCHECK_LT(ratio_, 1);
+    CUDNN_ENFORCE(cudnnCreateTensorDescriptor(&data_desc_));
+
+    CUDNN_ENFORCE(cudnnCreateDropoutDescriptor(&dropout_desc_));
+    CUDNN_ENFORCE(cudnnDropoutGetStatesSize(cudnn_wrapper_.inline_cudnn_handle(), &states_size_in_bytes_));
```
asaadaldien: reinterpret_cast<size_t*>(&states_size_in_bytes_)

caffe2/operators/dropout_op_cudnn.cc
```diff
@@ -0,0 +1,234 @@
...
```
asaadaldien: reinterpret_cast<size_t*>(&states_size_in_bytes_)

asaadaldien (changes requested):
Still breaking the build. we need to explicitly cast int* -> size_t.

caffe2/operators/dropout_op_cudnn.cc
```diff
@@ -0,0 +1,236 @@
...
+          GetCudnnTensorFormat(StorageOrder::NCHW),
+          cudnnTypeWrapper<T>::type,
+          size_prod,
+          1,
+          1,
+          1));
+    // get the reserve space we need
+    CUDNN_ENFORCE(cudnnDropoutGetReserveSpaceSize(
+          data_desc_, reinterpret_cast<size_t*>(&reserve_space_size_in_bytes_)));
+    // store both reserve and states in the same tensor
```
asaadaldien, line 116: Unfortunately VS2015 is not happy about (int + size_t) :( 
Can you replacing sizeof(T) as the following :
```
int elem_size = static_cast<int>(sizeof(T));
vector<int> state_size{(reserve_space_size_in_bytes_ + states_size_in_bytes_ + elem_size) / elem_size};
```

Same for the bellow

asaadaldien (changes requested):
Sorry for the asking to change it again @slayton58.

caffe2/operators/dropout_op_cudnn.cc
slayton58: Done

caffe2/operators/dropout_op_cudnn.cc
```diff
@@ -126,7 +127,7 @@ bool CuDNNDropoutOp::DoRunWithType() {
           dropout_desc_,
           cudnn_wrapper_.inline_cudnn_handle(),
           ratio_,
-          reserve_data + reserve_space_size_in_bytes_ / sizeof(T),
+          reserve_data + reserve_space_size_in_bytes_ / elem_size;
```
asaadaldien: typo:
s/elem_size;/elem_size,

asaadaldien (changes requested):

caffe2/python/operator_test/pooling_test.py
```diff
@@ -145,6 +146,34 @@ def test_pooling_3d(self, stride, pad, kernel, size, input_channels,
            size=st.integers(7, 9),
            input_channels=st.integers(1, 3),
            batch_size=st.integers(1, 3),
+           **hu.gcs_gpu_only)
```
asaadaldien: Because MaxPoolWithIndex doesn't have a CPU implementation we need to run this only when GPU is supported can you add the following before line 143
`@unittest.skipIf(not workspace.has_gpu_support, "No gpu support")`

PR #400, opened 2017-04-24
Support fp16 output from ImageInputOp

slayton58:
Refactors out option -> datatype cast
Allows native fp16 output from GPU transformation path
Expanded CastOp to allow fp16 in/out on GPU

### DIFF
--- a/caffe2/image/image_input_op.h
+++ b/caffe2/image/image_input_op.h
@@ -61,11 +61,13 @@ class ImageInputOp final
   bool is_test_;
   bool use_caffe_datum_;
   bool gpu_transform_;
-  TensorProto_DataType output_type_;
 
   // thread pool for parse + decode
   int num_decode_threads_;
   std::shared_ptr<TaskThreadPool> thread_pool_;
+
+  // Output type for GPU transform path
+  TensorProto_DataType output_type_;
 };
 
 
--- a/caffe2/image/image_input_op.h
+++ b/caffe2/image/image_input_op.h
@@ -7,6 +7,7 @@
 
 #include "caffe/proto/caffe.pb.h"
 #include "caffe2/core/db.h"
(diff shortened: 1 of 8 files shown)


caffe2/image/transform_gpu.cu
```diff
@@ -62,4 +61,19 @@ bool TransformOnGPU<uint8_t, float, CUDAContext>(Tensor<CUDAContext>& X, Tensor<
   return true;
 }
 
+template <>
+bool TransformOnGPU<uint8_t, float16, CUDAContext>(Tensor<CUDAContext>& X, Tensor<CUDAContext> *Y, float std, float mean, CUDAContext *context)
```
asaadaldien: The code is the same as above. What about moving the implementation to the templated version (line:45) ?

asaadaldien (changes requested):
Looks good thanks @slayton58.
Just one comment :)

caffe2/image/transform_gpu.cu
slayton58: I swear there was a reason I didn't do that before, but I can't remember, and I fixed it this time, so it obviously wasn't too serious :)

Updated.

caffe2/image/image_input_op.h
```diff
@@ -60,6 +61,7 @@ class ImageInputOp final
   bool is_test_;
   bool use_caffe_datum_;
   bool gpu_transform_;
+  TensorProto_DataType output_type_;
```
asaadaldien: Move this line bellow line:68 (breaks initialization order rule).

asaadaldien (changes requested):

caffe2/image/image_input_op.h
slayton58: Done.

PR #418, opened 2017-04-25
Fix compiler warning in utils/conversions.h

lukeyeager:
`/data/caffe2/caffe2/utils/conversions.h(137): warning: variable "r" was declared but never referenced`

Comes from https://github.com/caffe2/caffe2/commit/0fccf7a4f7bb930c884aa70894f087d85e870b0a

/cc @asaadaldien


lukeyeager:
Looks like this was fixed in https://github.com/caffe2/caffe2/pull/400.

PR #420, opened 2017-04-25
reverting readme to previous version

aaronmarkham:
slimmed down (instructions for install are on website instead of both places) + this has the feedback survey link


aaronmarkham:
Already patched elsewhere.

PR #430, opened 2017-04-26
Translator for BatchNorm to SpatialBN

milewis1:
Based on the translator from @ezineo, but more flexible -- although possibly less efficient. This forces the bias parameter to the SpatialBN operator to be zero allowing the translator to handle Caffe networks that don't have a Scale layer following their BatchNorm layer. However, if the Caffe network has a Scale layer (with bias term) after the BatchNorm layer this method will wind up creating three Caffe2 operations from the original 2, whereas with some smarter translation the bias term could be incorporated into the SpatialBN operator and actually reduce the number of operations from 2 to 1.


MarcelSimon:
Thanks for the PR! While the conversion works without errors, using the models for prediction fails with all kinds of funny errors.

For example, I tried to convert https://github.com/cvjena/cnn-models Each model fails to convert and shows different error messages when using it for prediction. For example, the AlexNet BN model fails with `Error in `/home/simon/bin/conda/envs/py2/bin/python': munmap_chunk(): invalid pointer ...`

milewis1:
I'll take a look at that conversion and see if I can figure out what's
wrong. Thanks for the heads up.

-- Mike

MarcelSimon:
Thanks! The ResNet10 models fails with `RuntimeError: [enforce fail at operator.cc:69] schema->Verify(operator_def). Operator def did not pass schema checking: input: "conv1" input: "conv1_scale" input: "conv1_bias" input: "conv1_mean" input: "conv1_var" output: "conv1" type: "SpatialBN" arg { name: "is_test" i: 1 } arg { name: "epsilon" f: 1e-05 } arg { name: "order" s: "NCHW" } `, if this is more of a help

milewis1:
@MarcelSimon I've updated the PR with a change that should get the AlexNet BN model you pointed out working, although I don't have a good way to test it. The way Caffe stored the scale blob and the way Caffe2 wants it are different.

The ResNet10 model won't work though, since it requires the BatchNorm layer to be in place, which Caffe2 doesn't currently allow -- although I have. PR out for that too: #469

tlhchutang:
@milewis1 Thanks for sharing the solution, however, I found there maybe a small issue in this PR:
In Caffe, the BN operation is conducted as (x - mean / scale) / (sqrt(var / scale)),
In Caffe2, the SpatialBN is conducted as (x * sqrt(1 / var) * scale) + (bias - mean * sqrt(1 / var) * scale), where bias is normally 0, so the formation could be written as (x - mean) * scale / sqrt(var), which is not generally equivlent to that of Caffe unless the mean and var  parameters of SpatialBN operation are normalized by the scale parameter, i.e.,
mean_caffe2 = mean_caffe / scale_caffe, 
var_caffe2 = var_caffe / scale_caffe, 
scale_caffe2 = scale_caffe / scale_caffe = 1.0
   
So, I think the following could be added before they are converted to Caffe2 tensors
pretrained_blobs[0] = pretrained_blobs[0] / pretrained_blobs[2]
pretrained_blobs[1] = pretrained_blobs[1] / pretrained_blobs[2]
pretrained_blobs[2] = pretrained_blobs[2] / pretrained_blobs[2]

milewis1:
I didn’t even consider that the calculated would be that different. I will
take a look though and update this as necessary.

milewis1:
I've made this adjust as you suggests and updated the PR.

orionr:

PR #443, opened 2017-04-28
Add ScatterWeightedSum for GPU.

pooyadavoodi:
- Adding ScatterWeightedSumOp for CUDA.
- This version does not support input weight (weight0). In other words, the input weight has to be 1.0, otherwise the op exits.
- To check the value of weight0, we copy its value from device to host at: https://github.com/caffe2/caffe2/pull/443/files#diff-2a77f80797072e8443f4867cb709fb40R244

### DIFF
--- a/caffe2/operators/utility_ops.cu
+++ b/caffe2/operators/utility_ops.cu
@@ -8,6 +8,7 @@
 #include "utility_ops.h"
 
 namespace caffe2 {
+CAFFE_KNOWN_TYPE(const float*);
 
 __global__ void NanCheckKernel(int N, const float* X, bool* result) {
   bool has_nan = false;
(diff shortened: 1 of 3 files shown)


caffe2/operators/utility_ops.cu
```diff
@@ -181,8 +181,101 @@ bool GatherOp<CUDAContext>::DoRunWithType() {
...
+  // in device memory, copy pointers to a host vector and then copy back into a device array.
+  vector<const float*> x_data_host(B);
+  vector<const float*> weights_host(B);
+  for (int inp = 3; inp < InputSize(); inp += 2) {
+    x_data_host [(inp-3)/2] = static_cast<const float*>(Input(inp).raw_data());
+    weights_host[(inp-3)/2] = static_cast<const float*>(Input(inp+1).raw_data());
+  }
+  const float** x_data_device;
+  const float** weights_device;
+  x_data_device = (const float**) context_.New(B * sizeof(const float*));
```
akyrola: Doing this mallocs will be costly and also require acquiring the mutex.

It would be best if we could avoid the whole host-device copy business, but if it is not avoid, it is better to store x_data_device and weights_device as class fields and tensors, similarly to for example losses_ in SoftmaxWithLossOp. Then the allocations happen only if the buffer needs to be increased.

caffe2/operators/utility_ops.cu
```diff
@@ -181,8 +181,101 @@ bool GatherOp<CUDAContext>::DoRunWithType() {
...
+             const TIndex slice_size,
+             const float** alpha,
+             const float** X,
+             const T_INDEX* Indices, 
+             float* Y,
+             const TIndex M) {
+  for (int i = blockIdx.x; i < N; i += gridDim.x) {
+    T_INDEX idx = Indices[i];
+    float* y_offset = Y + (idx * slice_size);
+    for (int b = 0; b < B; b++) {
```
akyrola, line 208: To get better parallelism, is it possible to combine these loops?

```
for(int k = threadIdx.x; k < B * slice_size; k += blockDim.x) {
    const int b = k / slice_size;
    const int j = k % slize_size;
    const float* x_offset = X[b] + (i * slice_size);
    atomicAdd(&y_offset[j], (*alpha[b]) * x_offset[j]);
} 

```
(Not sure if correct)

caffe2/operators/utility_ops.cu
```diff
@@ -181,8 +181,101 @@ bool GatherOp<CUDAContext>::DoRunWithType() {
...
+      const float* x_offset = X[b] + (i * slice_size);
+      for (int j = threadIdx.x; j < slice_size; j += blockDim.x) {
+        atomicAdd(&y_offset[j], (*alpha[b]) * x_offset[j]);
```
akyrola, line 211: is the atomicAdd needed because of duplicate indices?

akyrola (changes requested):
Some comments, particularly should avoid the New()/Deletes().

caffe2/operators/utility_ops.cu
pooyadavoodi: Yes, because each index is managed by a CUDA block, and thus duplicated indices will cause different blocks write to the same memory cells.

pooyadavoodi:
Fixed the New/Delete issue by storing the arrays as class fields, and do the allocation only if the size changes.

caffe2/operators/utility_ops.cu
```diff
@@ -181,8 +181,107 @@ bool GatherOp<CUDAContext>::DoRunWithType() {
...
+  context_.Copy<float,CUDAContext,CPUContext>(1, weight0.template data<float>(), &w0);
+  OPERATOR_NEEDS_FEATURE(
+    w0 == 1.0,
+    "ScatterWeightedSumOp only supports weight_0=1 on CUDAContext");
+
+  const TIndex B = (InputSize()-3)/2;
+
+  // In order to have all device pointers of x_i (and weight_i similarly) consecutively 
+  // in device memory, copy pointers to a host vector and then copy back into a device array.
+  if(x_data_host.size() != B) {
```
akyrola: It is better to allocate only if the size is smaller than B?

Would you be able to use a Tensor<CUDAContext> here instead of x_data_device? Then just call mutable_data<float*>?

pooyadavoodi: Actually allocate if the size is larger than B?
Because if it's smaller we can reuse the buffer.

pooyadavoodi:
- Now using tensors for both host and device arrays.
- I was thinking of implementing ScatterScaleOp which then will be used in ScatterWeightedSum to support weight0 != 1.0. Then we won't need these host->device copies.

caffe2/operators/utility_ops.cu
pooyadavoodi: It's possible to change. Both are correct, however I compared the performance for large sizes and saw no diffs.

akyrola:
I guess B is usually small so it does not matter. Will accept this when online

Sent from my iPhone

On Apr 28, 2017, at 18:08, Pooya Davoodi <notifications@github.com<mailto:notifications@github.com>> wrote:


@pooyadavoodi commented on this pull request.

________________________________

In caffe2/operators/utility_ops.cu<https://github.com/caffe2/caffe2/pull/443#discussion_r114041915>:

> + */
+template<typename T_INDEX>
+__global__ void
+AxpySliceKernel(
+             const TIndex N,
+             const TIndex B,
+             const TIndex slice_size,
+             const float** alpha,
+             const float** X,
+             const T_INDEX* Indices,
+             float* Y,
+             const TIndex M) {
+  for (int i = blockIdx.x; i < N; i += gridDim.x) {
+    T_INDEX idx = Indices[i];
+    float* y_offset = Y + (idx * slice_size);
+    for (int b = 0; b < B; b++) {


It's possible to change. Both are correct, however I compared the performance for large sizes and saw no diffs.

caffe2/operators/utility_ops.cu
```diff
@@ -181,8 +182,108 @@ bool GatherOp<CUDAContext>::DoRunWithType() {
...
+  context_.Copy<float,CUDAContext,CPUContext>(1, weight0.template data<float>(), &w0);
+  OPERATOR_NEEDS_FEATURE(
+    w0 == 1.0,
+    "ScatterWeightedSumOp only supports weight_0=1 on CUDAContext");
+
+  const TIndex B = (InputSize()-3)/2;
+
+  // In order to have all device pointers of x_i (and weight_i similarly) consecutively 
+  // in device memory, copy pointers to a host vector and then copy back into a device array.
+  if(x_data_host_.size() < B) {
```
akyrola: shouldn't we actually always Resize to B? Note that the tensors won't be reallocated if B is smaller than the capacity?

pooyadavoodi: That's right. `tensor.h` will take care of the check: https://github.com/caffe2/caffe2/blob/master/caffe2/core/tensor.h#L279

Will fix and update the PR.

pooyadavoodi: @akyrola Looks like `Resize` changes `size_` even if `capacity_` is big enough. This means after `Resize(B)` is done, `size()` will return `B`, even if capacity is larger than `B`. Am I right?

akyrola:
That’s right. Capacity and size are different things, capacity is always >= size.

PR #453, opened 2017-04-29
Remove command "touch cmake".

hgaiser:
It is quite normal to rerun cmake to find new files through GLOB commands. This external command always forces cmake to run when you run make, meaning every make command takes longer than necessary. This PR removes the external command `touch CMakeLists.txt` and therefore leaves that decision up to the user when to rerun cmake and speeds up building when cmake is not required to rerun.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,9 +4,6 @@ project(Caffe2 CXX C)
 
 # TODO(bwasti): versioning
 
-# We want CMake to GLOB everything every time.
-execute_process(COMMAND find "${PROJECT_SOURCE_DIR}" -name "CMakeLists.txt" -exec touch {} \;)
-
 # We want to make sure that all submodules are being properly initialized, if
 # one is building from a git checkout.
 find_package(Git)


Yangqing:
@bwasti , I think we needed this because we do GLOB_RECURSE?

bwasti:
Yea, this was needed because of GLOB* functions, which don't automatically detect newly added files.  @hgaiser is right that the user _should_ be aware of this restriction and it definitely slows things down by having this executed every time.  With that being said, I agree we should remove "touch" (as most users will likely not be renaming/removing/adding files without some understanding of cmake's horrible GLOB).

Does anyone disagree?

bwasti (approved):

hgaiser:
@bwasti sounds good. I agree it isn't ideal to have to rerun cmake in case filenames change or files are added, but this was probably worse ;)

Out of curiosity, @facebook-github-bot mentions the PR has been "imported". Is this Facebooks method of accepting PRs? Is that to preserve proper rights?

cbalint13:
@bwasti , @hgaiser 

* As consequence of that **touch** ```make install DESTIR=somewhere``` recompile once more (annoying) everything and then only do the requested install step ...

This fix also this.

bwasti:
@hgaiser the import is actually to run internal builds/tests.  The bot will close this PR once it has landed internally (which has only been delayed by my waiting for any disagreement).  I just launched that job so you will see this closed soon.

hgaiser:
I see, thanks for the explanation, makes sense.

PR #457, opened 2017-04-29
Enable bitcode for iOS builds

abjurato:
build_ios.sh now have `-fembed-bitcode` flags for cmake and passes these flags to build_host_protoc.sh (which now accepts optional argument `--other-flags`). That allows to use output libs (libCaffe2_CPU.a, libCAFFE2_NNPACK.a, libCAFFE2_PTHREADPOOL.a and libprotobuf-lite.a, libprotobuf.a respectively) in Xcode projects with bitcode enabled.

Bitcode is enabled by default in all projects since Xcode7, is crucial for slicing and is mandatory for watchOS targets. Enabling bitcode for target requires bitcode to be enabled for all dependencies also, so Caffe2 built without bitcode forces developers to switch off bitcode for the whole app.

### DIFF
--- a/scripts/build_host_protoc.sh
+++ b/scripts/build_host_protoc.sh
@@ -15,9 +15,26 @@ mkdir -p $BUILD_ROOT/build
 
 cd $BUILD_ROOT/build
 CMAKE=$(which cmake || which /usr/bin/cmake || which /usr/local/bin/cmake)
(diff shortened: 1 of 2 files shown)


abjurato:
Sorry for so many updates, cross compilation turned out to be not as easy as I thought :)
Now all tests passed and PR can be reviewed.

Yangqing:
Thanks so much! I did not know about this before. Importing.

PR #463, opened 2017-04-30
Python3 compilation

hgaiser:
This PR does two things:

1. It allows compilation of caffe2 using python3 libraries.
2. It reverts https://github.com/caffe2/caffe2/commit/6cd340ec67d5b4a7efe92f5017b903e50359def8 (PR: https://github.com/caffe2/caffe2/pull/361) as this is not a proper fix for forcing a certain python version (in addition, it doesn't work).

Also note that this PR only enables compilation of python3 bindings. The mnist tutorial seems to mostly run, however some python tools in caffe2 need to be updated to be python2/3 compatible. I will hopefully work on that soon, but this PR can be viewed as a start on python3 compatibility.

EDIT: I'll guard the use of PyUnicode so that it is only used on python3 later today (I thought it was python2/3 compatible but apparently it isn't).

### DIFF
--- a/caffe2/python/pybind_state.cc
+++ b/caffe2/python/pybind_state.cc
@@ -16,6 +16,16 @@
 #include "caffe2/utils/windows_cpu_supports.h"
 #endif
 
+#if PY_MAJOR_VERSION >= 3
+  #define PyString_Check PyUnicode_Check
+  #if PY_VERSION_HEX < 0x03030000
+    #error "Python 3.0 - 3.2 are not supported."
+  #else
+  #define PyString_AsString(ob) \
+    (PyUnicode_Check(ob)? PyUnicode_AsUTF8(ob): PyBytes_AsString(ob))
+  #endif
+#endif
+
 namespace caffe2 {
 namespace python {
 
--- a/cmake/Dependencies.cmake
+++ b/cmake/Dependencies.cmake
@@ -182,7 +182,6 @@ add_definitions(-DEIGEN_MPL2_ONLY)
 
 # ---[ Python + Numpy
 if (BUILD_PYTHON)
-  set(Python_ADDITIONAL_VERSIONS 2.8 2.7 2.6)
   find_package(PythonInterp 2.7)
   find_package(PythonLibs 2.7)
   find_package(NumPy REQUIRED)


kashif:
@hgaiser yes I used the following in `pybind_state.cc`:

```c++
#if PY_MAJOR_VERSION >= 3
  #define PyString_Check PyUnicode_Check
  #if PY_VERSION_HEX < 0x03030000
    #error "Python 3.0 - 3.2 are not supported."
  #else
  #define PyString_AsString(ob) \
    (PyUnicode_Check(ob)? PyUnicode_AsUTF8(ob): PyBytes_AsString(ob))
  #endif
#endif
```

hgaiser:
@kashif that is more or less what I had in my mind too. I hope you don't mind that I shamelessly copied your snippet into this PR ;).

kashif:
@hgaiser with this change you can undo the `PyString_Check` change...

hgaiser:
Terribly sorry, I was aware of that and under the impression I made the correct changes. Must have not been paying attention when committing that part. I'll change it in a minute, thanks for letting me know.

bwasti:
looks like a great change! going to wait until tests pass.

hgaiser:
@bwasti seems that the build job stopped prematurely:
`The job exceeded the maximum time limit for jobs, and has been terminated.`

Shall I trigger a rebuild by closing and opening this PR ?

hgaiser:
@bwasti, it seems this PR is imported but not yet pushed to master, correct? May I ask why?

hgaiser:
It seems this PR is obsolete given https://github.com/caffe2/caffe2/commit/6530c9a035682fbe94e882653ad624ca30c26c84 from @tomdz .

@tomdz can you clarify your approach? Does the approach in this PR not work for you? It seems a lot cleaner.

tomdz:
@hgaiser As far as I know, Caffe2 core code only uses `std::string` and assumes that they contain ASCII-only strings and would be unable to deal with multi-byte strings (such as UTF-8 strings). Thus using `PyUnicode_AsUTF8` could result in weird errors and my PR instead enforces ascii strings.

hgaiser:
> @hgaiser As far as I know, Caffe2 core code only uses std::string and assumes that they contain ASCII-only strings and would be unable to deal with multi-byte strings (such as UTF-8 strings). Thus using PyUnicode_AsUTF8 could result in weird errors and my PR instead enforces ascii strings.

If Caffe2 enforces ASCII characters everywhere it would be incorrect, because filenames for instance can contain non-ASCII characters. The place where it is used in [`pybind_state.cc`](https://github.com/caffe2/caffe2/blob/master/caffe2/python/pybind_state.cc#L216) for example is a filename (granted it is only used for logging there, but still). Seeing as python passes Unicode objects, and filenames etc. can contain Unicode characters, the only logical solution would seem to use UTF-8.

I'm not sure what you mean exactly with weird errors; where in the Caffe2 core code would non-ASCII characters cause an issue? I could look into those and try to fix them.

hgaiser:
Kindly nudging @tomdz for a reply, I'm still curious about this :)

tomdz:
For Python 2, Python and pybind bind between `std::string` and Python's `str` object, which is ASCII/bytes. There are a few places in Caffe2 where ASCII is assumed and stuff would fail for multi-byte strings (e.g. wherever sub strings are used, such as in the net printer).
In Python 3 this is more complicated as there is no magic conversion between `bytes` and `str` (which is `unicode` in Python 2), and thus a bunch of tests failed with Python 3 when going from Python to C++ or back (such as for PythnonOp).
Making Caffe2 fully unicode aware is on the roadmap I believe, but for now we've made the decision to stick to ASCII/bytes and handle unicode on the way from Python to C++.

PR #464, opened 2017-04-30
Enable c++11 support through CMake functions.

hgaiser:
This PR enables C++11 support using CMake variables (`CMAKE_CXX_STANDARD` for choosing the standard and `CMAKE_CXX_STANDARD_REQUIRED` for disabling CMake's default functionality of "decaying" to c++98 if c++11 is not available).

These variables are added since CMake 3.1.13 so a bump of the minimum required cmake version was necessary.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.12 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.1 FATAL_ERROR)
 
 project(Caffe2 CXX C)
 
@@ -100,9 +100,14 @@ endif()
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "binaries")
 
 # ---[ Build flags
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c11")
+set(CMAKE_C_STANDARD 11)
+set(CMAKE_C_STANDARD_REQUIRED on)
 if(NOT MSVC)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -fPIC")
+  # Enable C++11 support
+  set(CMAKE_CXX_STANDARD 11)
+  set(CMAKE_CXX_STANDARD_REQUIRED on)
+
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-narrowing")
 else()
   if (NOT ${BUILD_SHARED_LIBS})
--- a/cmake/MiscCheck.cmake
+++ b/cmake/MiscCheck.cmake
@@ -1,6 +1,7 @@
 INCLUDE(CheckCXXSourceCompiles)
 
-set(CMAKE_REQUIRED_FLAGS "-std=c++11")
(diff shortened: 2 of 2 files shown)


Yangqing:
Thanks so much for this!

A quick check: http://packages.ubuntu.com/trusty/cmake cmake on 14.04 is actually still 2.8, so do you think it is possible for us to provide a backward compatible path for folks who have an older cmake?

hgaiser:
> Thanks so much for this!
> 
> A quick check: http://packages.ubuntu.com/trusty/cmake cmake on 14.04 is actually still 2.8, so do you think it is possible for us to provide a backward compatible path for folks who have an older cmake?

Ah .. 14.04 is a target platform for caffe2? I suppose it makes sense since it's a LTS, but it can be annoying in cases such as this. I suppose there are two options then, merge this PR and provide people a [guide](https://askubuntu.com/questions/610291/how-to-install-cmake-3-2-on-ubuntu-14-04) to upgrade cmake in Ubuntu 14.04 or don't merge this PR and keep it the way it was.

ps. I just updated this PR to also set the `CMAKE_C_STANDARD` variable.

Yangqing:
@hgaiser I found that one can potentially use this to do cross version support. Would you like to give that a shot?

if(CMAKE_VERSION VERSION_LESS "3.1")
  include(CheckCXXCompilerFlag)
  CHECK_CXX_COMPILER_FLAG("-std=c++11" COMPILER_SUPPORTS_CXX11)
  if(NOT COMPILER_SUPPORTS_CXX11)
    MESSAGE(FATAL_ERROR "Your compiler does not support c++11")
  endif()
  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
else()
  set(CMAKE_CXX_STANDARD 11)
endif()

hgaiser:
> @hgaiser I found that one can potentially use this to do cross version support. Would you like to give that a shot?
> 
> if(CMAKE_VERSION VERSION_LESS "3.1")
> include(CheckCXXCompilerFlag)
> CHECK_CXX_COMPILER_FLAG("-std=c++11" COMPILER_SUPPORTS_CXX11)
> if(NOT COMPILER_SUPPORTS_CXX11)
> MESSAGE(FATAL_ERROR "Your compiler does not support c++11")
> endif()
> set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
> else()
> set(CMAKE_CXX_STANDARD 11)
> endif()

Similar to the discussion [here](https://github.com/BVLC/caffe/pull/5294#issuecomment-305108000), having cross version support in my opinion defeats the purpose of this PR. This PR is intended to use a cleaner method of choosing CXX standard, but if both are still present it only adds unnecessary complexity.

My advise would be to drop official support for Ubuntu 14.04 (it's over 3 years old ;) ) and provide a guide for installing cmake >3.1 if necessary. The alternative is to ignore this PR in favor of support on Ubuntu 14.04.

hgaiser:
@Yangqing what is the verdict?

Yangqing:
@hgaiser I asked around a little bit and seems that Cmake <3.0 is still something preferred... so maybe let's keep the old way? It's definitely not very ideal and I in fact recommend putting the backward-compatible, but newer approach for 3.1 :)

williford:
@Yangqing I implemented your suggestion in PR #1027. Would you take a look when you get the chance?

Yangqing:
Closing in favor of #1027, thanks so much guys :)

PR #469, opened 2017-05-01
Allow the spatial batch norm operator to be computed in place

milewis1:


salexspb:
should this also be reflected in the op tests? (for both order cases).

milewis1:
Yes it should. That was next in my plan, but there was so little interest I
never followed up. I can certainly update it though.

salexspb:
It just feels a little bit unsafe otherwise :) The code of the op has to be written in a curtain way for in place to work. Have you verified that it is correct?

milewis1:
It is. However, I did check the operator and have been using it successfully in my models. However, since I'm no longer using Caffe2 and there hasn't been any interest in the SpatialBN work. I'm going to close out this PR.

danielhauagge:
I'm very interested in this and could help. To get this merged, what would be the next steps?

milewis1:
When I first made the change I check the operator and made sure it could
work in place. I think the real thing needed would be getting tests
written.

PR #495, opened 2017-05-03
Mark in-place GPU dropout as broken, add test

lukeyeager:
I'll let y'all decide how you want to fix this (probably need a persistent curand buffer). Here's a test to verify the fix.

### DIFF
--- a/caffe2/operators/dropout_op.cu
+++ b/caffe2/operators/dropout_op.cu
@@ -33,6 +33,7 @@ bool DropoutOp<float, CUDAContext>::RunOnDevice() {
     // boolean numbers, we will generate into dY and write the result to
     // mask.
     float* Ydata = Y->mutable_data<float>();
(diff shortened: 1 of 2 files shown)


pietern:
LGTM... @Yangqing?

lukeyeager:
I'm happy to go ahead and implement the fix as well if you'd like. I can see several ways to handle it:

* Allocate/deallocate a new GPU buffer each time
* Allocate a GPU buffer once, and reallocating when the size changes
* Add a new output for the op which nobody would ever use
* Change the type of the mask output to float instead of bool (can still set to 1/0 after curand is done with it)

lukeyeager:
Ping. This is a pretty brutal silent failure right now.

akyrola:
Thanks for noticing this. Of course happy to get a fix for this :)

PR #502, opened 2017-05-04
Update CentOS installation guide.

kyu-sz:
The gflags provided in yum of the current CentOS does not match the dependency requirement of the latest glog version. It also need to be cloned from Github and manually installed from source. Also, the glog need the latest CMake3 to reconfigure.

### DIFF
--- a/_docs/getting-started/centos.md
+++ b/_docs/getting-started/centos.md
@@ -55,15 +55,15 @@ snappy-devel
 gflags and glog is not found in yum for this version of Linux, so install from source:
 
 ```
-git clone https://github.com/gflags/gflags.git
-cd gflags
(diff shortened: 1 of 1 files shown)


kyu-sz:
I have just signed the CLA.

aaronmarkham (commented):
Given this change we'd need to remove the yum install gflags-dev that happens next. It's just after the troubleshooting glog compliation section.

kyu-sz:
Sorry, I've ignored an option that's critical. The Caffe2 seems to find symbols from the shared library of gflags, so the gflags should be compiled with option "-DBUILD_SHARED_LIBS=ON". @aaronmarkham

PR #504, opened 2017-05-04
Check GCC version if compiling with CUDA support

pietern:
Otherwise compilation fails pretty far into the build, which is inconvenient.

The error reported when trying to compile with GCC 6:

    CUDA 8.0 is not compatible with GCC version >= 6. Use the following
    options to configure GCC version 5:

      -DCMAKE_CXX_COMPILER=/usr/bin/g++-5
      -DCMAKE_C_COMPILER=/usr/bin/gcc-5
      -DCUDA_HOST_COMPILER:FILEPATH=/usr/bin/gcc-5


CMakeLists.txt
```diff
@@ -57,8 +57,8 @@ option(USE_MPI "Use MPI" ON)
 option(USE_GLOO "Use Gloo" ON)
 option(BUILD_SHARED_LIBS "Build libcaffe2.so" ON)
 option(USE_OPENMP "Use OpenMP for parallel code" ON)
-option(BUILD_PYTHON "Build python binaries" ON)
-option(BUILD_BINARY "Build c++ binary executives" ON)
+option(BUILD_PYTHON "Build Python binaries" ON)
+option(BUILD_BINARY "Build C++ binary executives" ON)
```
lukeyeager: While you're at it, maybe "executables" instead of "executives"?  :wink:

pietern: Haha, didn't see that mistake! Good one!

PR #525, opened 2017-05-08
Fix CUDA check for gcc > 5.

hgaiser:
In response to https://github.com/caffe2/caffe2/pull/504 , this PR modifies the gcc compiler check for CUDA slightly. All ABI since [gcc-3](https://gcc.gnu.org/onlinedocs/libstdc++/manual/abi.html) are compatible with eachother. The check from https://github.com/caffe2/caffe2/pull/504 forced the 'regular' CXX / CC compiler to be set to gcc < 6 but this is not required.

According to the documentation for [FindCUDA](https://cmake.org/cmake/help/v3.0/module/FindCUDA.html), `CUDA_HOST_COMPILER` is set to `CMAKE_C_COMPILER` by default. This PR checks if `CMAKE_C_COMPILER` is too new for CUDA 8 and whether `CUDA_HOST_COMPILER` is set to `CMAKE_C_COMPILER`. It also modifies the message slightly.


hgaiser:
@pietern can you give feedback on this?

Yangqing:
Also cc'ing @slayton58 - Simon, what is the recommended way from NVidia about setting host compiler in cmake?

slayton58:
Either using `-DCUDA_HOST_COMPILER=<compiler>` or manually setting `-ccbin=<compiler>` in `CUDA_NVCC_FLAGS` is fine, no preference from us :)

hgaiser:
@Yangqing can this then be merged?

PR #526, opened 2017-05-08
Python3 compatibility (WIP)

hgaiser:
This PR is a first draft to get python2/3 compatibility in caffe2. Most noticeable changes come from the difference in `str` for python2 vs python3. python2 defines `bytes` and `str` as the same thing but in python3 they are two very different types.

Specifically the changes in `pybind_state` require some attention by caffe2 devs, as these are the areas where I am uncertain about the changes made.

Note that this PR is not finished, not all tests are running successfully in python3 yet and I have to test python2 again as well before this can be merged. I hope to continue work on this in a week or so, but meanwhile I was hoping to get some feedback on the changes made thus far.

ps. this PR depends on https://github.com/caffe2/caffe2/pull/463

### DIFF
--- a/caffe2/experiments/python/SparseTransformer.py
+++ b/caffe2/experiments/python/SparseTransformer.py
@@ -61,7 +61,7 @@ def transFCRelu(cur, id2node, name2id, ops, model):
     """
     # 1. add trans before the start of this chain
     # assuming that cur is an FC_Prune, and it has only one input
-    pre = cur.prev.itervalues().next()
+    pre = next(cur.prev.itervalues())
     # Create an node /op and insert it.
     # TODO(wyiming): check whether it is correct here
     current_blob = model.Transpose(cur.op.input[0], cur.op.input[0] + "_trans")
@@ -106,13 +106,13 @@ def transFCRelu(cur, id2node, name2id, ops, model):
         cur.visited = True
         pre = cur
         flag = False
-        for _, temp in cur.ops.iteritems():
+        for _, temp in cur.ops.items():
             if temp.optype == "Relu" or temp.optype == "FC_Prune":
                 flag = True
                 cur = temp
         if not flag:
             # assume that there is only 1 output that is not PrintOP
-            cur = cur.ops.itervalues().next()
+            cur = next(cur.ops.itervalues())
             cur.deleteInput(pre)
             print("No FC/RElu children")
             print(cur.op.type)
@@ -134,7 +134,7 @@ def Prune2Sparse(cur, id2node, name2id, ops, model):
         transFCRelu(cur, id2node, name2id, ops, model)
 
     cur.visited = True
-    for name, n in cur.ops.iteritems():
+    for name, n in cur.ops.items():
         Prune2Sparse(n, id2node, name2id, ops, model)
 
 
@@ -145,13 +145,13 @@ def net2list(net_root):
     bfs_queue = []
     op_list = []
     cur = net_root
-    for _, n in cur.ops.iteritems():
+    for _, n in cur.ops.items():
         bfs_queue.append(n)
     while bfs_queue:
         node = bfs_queue[0]
         bfs_queue = bfs_queue[1:]
         op_list.append(node.op)
-        for _, n in node.ops.iteritems():
+        for _, n in node.ops.items():
             bfs_queue.append(n)
 
     return op_list
--- a/caffe2/proto/caffe2.proto
+++ b/caffe2/proto/caffe2.proto
@@ -98,10 +98,12 @@ message Argument {
   optional string name = 1;
   optional float f = 2;
   optional int64 i = 3;
-  optional bytes s = 4;
-  repeated float floats = 5;
-  repeated int64 ints = 6;
-  repeated bytes strings = 7;
+  optional string s = 4;
+  optional bytes b = 5;
+  repeated float floats = 6;
+  repeated int64 ints = 7;
+  repeated string strings = 8;
+  repeated bytes bytes = 9;
 }
 
 // DeviceType that Caffe2 currently supports.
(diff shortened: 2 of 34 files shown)


hgaiser:
Did someone find the time to glance over this?

caffe2/experiments/python/SparseTransformer.py
```diff
@@ -106,13 +106,13 @@ def transFCRelu(cur, id2node, name2id, ops, model):
         cur.visited = True
         pre = cur
         flag = False
-        for _, temp in cur.ops.iteritems():
+        for _, temp in cur.ops.items():
```
lukeyeager, line 109: Using `six.iteritems()` would maintain the performance benefit in Py2. Not sure if the maintainers would want to add a new dependency, though.

caffe2/proto/caffe2.proto
```diff
@@ -98,10 +98,12 @@ message Argument {
...
-  optional bytes s = 4;
-  repeated float floats = 5;
-  repeated int64 ints = 6;
-  repeated bytes strings = 7;
+  optional string s = 4;
+  optional bytes b = 5;
+  repeated float floats = 6;
+  repeated int64 ints = 7;
+  repeated string strings = 8;
+  repeated bytes bytes = 9;
```
lukeyeager, line 106: I'm pretty sure it's never OK to change these fields - only add new ones.

caffe2/python/mint/app.py
```diff
@@ -74,14 +74,14 @@ def visualize_print_log(filename):
...
-        step = max(data.shape[0] / -args.sample, 1)
+        step = max(data.shape[0] // -args.sample, 1)
     else:
         step = args.sample
     xdata = np.arange(0, data.shape[0], step)
     # if there is only one curve, we also show the running min and max
     if data.shape[1] == 1:
         # We also print the running min and max for the steps.
-        trunc_size = data.shape[0] / step
+        trunc_size = data.shape[0] // step)
```
lukeyeager: Syntax error, no?

caffe2/python/data_workers.py
```diff
@@ -5,6 +5,8 @@
 from __future__ import print_function
 from __future__ import unicode_literals
 
+from future import standard_library
```
lukeyeager, line 8: NOTE: this is a new requirement

caffe2/python/operator_test/instance_norm_test.py
```diff
@@ -2,10 +2,12 @@
 from __future__ import division
 from __future__ import print_function
 
+from builtins import zip
```
lukeyeager: NOTE: this is a new requirement

lukeyeager (changes requested):
If you're picking a compatibility library, I suggest six because it's been around [longer](http://packages.ubuntu.com/search?keywords=python-six) than [future](http://packages.ubuntu.com/search?keywords=python-future) and because it does everything you need in one package instead of both `future` and `builtins`. But, I figure there are probably lots of compatibility libraries around, and you may know something I don't about some shortcoming of `six`.

caffe2/experiments/python/SparseTransformer.py
hgaiser: I considered this, but felt that since it is only an experiment script it was not necessary to add a compatibility module here. Do you disagree?

caffe2/proto/caffe2.proto
hgaiser: Can you elaborate on this? The bytes field is now called `s`, which in my opinion should be used for string. I can move `optional string s = 4;` to the end to preserve order, but I would really prefer to rename `bytes s` to `bytes b`.

caffe2/python/mint/app.py
hgaiser: Yes :) thanks.

caffe2/python/data_workers.py
hgaiser: True, I'll add it in the comment of the PR.

hgaiser:
Thank you @lukeyeager for the review. Regarding the choice of `six` vs `future`, it seems the recommended approach is to use `future` unless you must support older versions of python (`future` supports python 2.6+ and 3.3+). According to their [FAQ](http://python-future.org/faq.html#what-is-the-relationship-between-future-and-six), `future` is more feature rich than `six`. The official [python page for porting](https://docs.python.org/3/howto/pyporting.html) also suggests to choose `future`.

What is your reason for preferring `six` over `future` ? Is it that `six` is already used in caffe2? Would it, given the above information, be desirable to port the small bits of `six` currently used to `future`, or would the preference be to stick with `six` ?

caffe2/experiments/python/SparseTransformer.py
lukeyeager: Fair point. Not a big deal here.

caffe2/proto/caffe2.proto
lukeyeager: It's for backwards compatibility. If someone saves a trained model with one build of Caffe2, they should be able to load it in another build of Caffe2. See, for example, the hoops that the Caffe1 developers were constantly jumping through: https://github.com/BVLC/caffe/blob/rc5/src/caffe/proto/caffe.proto#L10-L22

Now, this only applies if we expect people to be saving to and loading from binary protocol buffers. If that's not an expected use-case, then it doesn't matter. Let's let the @caffe2 guys weigh in on this.

lukeyeager:
> What is your reason for preferring six over future ?

Good argument - you convinced me. I just hadn't heard about `future` before and was more familiar with `six`. Thanks for showing it to me!

hgaiser:
Rebased on master and fixed some issues as mentioned by @lukeyeager .

hgaiser:
This also seems to be largely obsolete after https://github.com/caffe2/caffe2/commit/6530c9a035682fbe94e882653ad624ca30c26c84 by @tomdz . I'll close this for now and open new PR(s) where necessary.

PR #539, opened 2017-05-09
fixing lmdb.cc when compiled on Windows (mkdir -> _mkdir)

### LINKED ISSUE #462: Windows 10: error in lmdb.cc
Hello together,

there seems to be an error in the file lmdb.cc in line 141, when you compile it under windows respectively VS2015:
```
if (mode == NEW) {
    CAFFE_ENFORCE_EQ(
        mkdir(source.c_str(), 0744), 0, "mkdir ", source, " failed");
  }
```
The function mkdir is deprecated in VS (see: https://msdn.microsoft.com/en-us/library/ms235326.aspx). Thus, I get an error if I compile caffe2 with lmdb support. A possible fix at least in VS could be:

```
  if (mode == NEW) {
    CAFFE_ENFORCE_EQ(
        _mkdir(source.c_str()), 0, "mkdir ", source, " failed");
  }
```

gfursin:
Should fix #462 .

### DIFF
--- a/caffe2/db/lmdb.cc
+++ b/caffe2/db/lmdb.cc
@@ -1,5 +1,9 @@
 #include "lmdb.h"  // NOLINT
 
+#if defined(_WIN32)
+#include <direct.h>
+#endif
+
 #include <sys/stat.h>
 
 #include <string>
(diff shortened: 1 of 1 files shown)


gfursin:
Ah, I should check not only _WIN32, but that the target is Windows - any suggestions how to do it?

gfursin:
Actually, seems like Travis and Appveyor fail not because of my changes...

caffe2/db/lmdb.cc
```diff
@@ -1,5 +1,9 @@
 #include "lmdb.h"  // NOLINT
 
+#if defined(_WIN32)
```
Yangqing, line 3: #if defined(_MSC_VER)

is what you may be looking for :)

Yangqing:
Thanks for the fix - the continuous build might be just having an out of time error.

PR #547, opened 2017-05-10
Fixing some tests

lukeyeager:
As @dzhulgakov said at https://github.com/caffe2/caffe2/pull/227#issuecomment-295084443, it would be nice to avoid this stream of CPU-only test fixes.

The second fix could have been avoided if tests were run on TravisCI. I think the TravisCI infra could be greatly improved if we used ccache like your colleagues at PyTorch: https://github.com/pytorch/pytorch/pull/614. Would you be interested in a PR which does this?


lukeyeager:
> I think the TravisCI infra could be greatly improved if we used ccache like your colleagues at PyTorch: pytorch/pytorch#614. Would you be interested in a PR which does this?

Example implementation at https://github.com/caffe2/caffe2/pull/550

PR #550, opened 2017-05-10
[TravisCI] Add target for quick build+test

lukeyeager:
Once the build is cached, QUICKTEST takes less than 3 minutes to install+build+test (first build is ~13 minutes).

Future TravisCI improvements:
* Refactor other build targets so they're fast enough to build in under 45 mins
* Run tests for other build targets
* Run Python tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,7 +63,18 @@ matrix:
           packages:
           - g++-5
       env: COMPILER=g++-5
+    # 10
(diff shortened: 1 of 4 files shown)


lukeyeager:
After https://github.com/caffe2/caffe2/pull/551 is merged, I'll rebase this so Travis actually passes.

PR #570, opened 2017-05-13
Optional force conv algorithms

slayton58:
Allow user to force cuDNN convolution algorithms from python - useful if you're using a standard network and don't want to pay the cost of exhaustive search.

Defined as an array in the order of [fwd, wgrad, dgrad].

Also refactors cudnn_conv_op slightly to split the code to do wgrad and dgrad a little more.

### DIFF
--- a/caffe2/operators/conv_op_cudnn.cc
+++ b/caffe2/operators/conv_op_cudnn.cc
@@ -31,6 +31,14 @@ inline void LogCuDNNPerfStats(
             << stat.memory;
   }
 }
+
+// Easier indexing into force_algo_ vector
+enum {
+  ALGO_FWD = 0,
+  ALGO_WGRAD = 1,
+  ALGO_DGRAD = 2
+} algoIndex_t;
+
 }  // namespace
 
 class CudnnConvOpBase : public ConvPoolOpBase<CUDAContext> {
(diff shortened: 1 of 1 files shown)


asaadaldien (commented):
LGTM!
One comment, can we explicitly pass the algorithm to force as individual arguments (ex: force_algo_fwd, force_algo_bwd_filter and force_algo_bwd_data) ?

slayton58:
I can add this, I'm not the biggest fan of the verbosity of those though. I did want to be able to specify a dictionary for this: `force_algo={'fwd' : 0, 'dgrad' : 1, 'wgrad' : 2}`, but it was awkward :(

asaadaldien:
My only concern is the future modifications of this code one might use wrong index -> algorithm.
which is not an easy bug to find :). So I think explicitly passing this might be a good idea.

slayton58:
Fair enough :) How about I expose both ways then?

asaadaldien:
I am fine with supporting two ways of parametrizing this as long as :
1- Both cases mutate the same state parameters which are explicit (ex : int force_algo_fwd, force_algo_bwd_filter and force_algo_bwd_data)
2- Users can use at most one.

slayton58:
I actually misinterpreted you. I thought you wanted an additional way of exposing this through the front-end, not a change in the backend.

I don't see the advantage of what you're asking. You can do the same by simply using an enum to access the existing `vector<int> force_algo_` variable.

edit: This also greatly simplifies having access to both `force_algo=[0,0,0]` and `force_algo_fwd=0, force_algo_bwd_filter=0, force_algo_bwd_data=0`

asaadaldien:
I want an additional way to expose the arguments in the frontend.
In the backend you can use vector<int> force_algo_ with an enum { force_algo_fwd, force_algo_bwd_filte, force_algo_bwd_data} or explicit ints as long as both cases mutates the same backend state parameters.

slayton58:
Updated. `force_algo` takes precedence over `force_algo_fwd, force_algo_wgrad, force_algo_dgrad`

PR #588, opened 2017-05-17
Execute submodule init command in srcdir

ChrisKitching:
Having this submodule git command in the cmake script is questionable. If you are having it, at least ensure it works correctly for out-of-tree builds (which are the norm when using cmake).


hgaiser:
I prefer to remove the command entirely and let the user perform these steps. The command is removed in https://github.com/caffe2/caffe2/pull/382

ChrisKitching:
I am a fan of that idea.

PR #596, opened 2017-05-17
Add stack traces on fatal signals

danzimm:
When a fatal signal is fired to a task that links against caffe2 this PR adds stacktraces from every thread that's currently running. Only linux is supported currently. The signals that are currently supported are SIGABRT, SIGINT, SIGILL, SIGFPE, SIGBUS and SIGSEGV (more signals can easily be added, but for now this seemed like the major signals that might be fired - see signal_handler.cc:138 for the table of signals).

I've added tests that verify that each of those signals indeed output the expected number of stacktraces.

We need to add linking against libdl since on linux apparently it's not implicitly always linked in (I'm coming from macOS where I believe it is).

Example output can be found [here](https://gist.github.com/danzimm/814faa1229d9c54f359d23ba038344a6) - note that the signal name changes depending on the signal that was sent (as well as the number in parenthesis that corresponds to the specified signal).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -113,6 +113,10 @@ else()
   endif()
 endif()
 
+if(NOT APPLE AND UNIX)
+  list(APPEND Caffe2_DEPENDENCY_LIBS dl)
+endif()
+
 if (CAFFE2_CPU_FLAGS)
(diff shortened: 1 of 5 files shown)


pietern:
Restarted Travis build. This should now come back green.

pietern:
Thanks @danzimm!

PR #606, opened 2017-05-18
[WIP] fp16 training

slayton58:
fp16 parameter initialization for convolution, FC
Map of low precision param -> high precision copy
Example of fp16 training in CIFAR10

### DIFF
--- a/caffe2/python/data_parallel_model.py
+++ b/caffe2/python/data_parallel_model.py
@@ -195,6 +195,19 @@ def Parallelize_GPU(
         devices, model_helper_obj, model_helper_obj.param_init_net, sync_names
     )
 
+    # if in a multi-precision environment, make sure that the broadcast parameters
+    # (low-precision) are still in sync with the high-precision copies.
+    # Otherwise training goes badly
+    # todo(slayton): Work with non-fp16 low-precision copies
+    for device in devices:
+        device_opt = core.DeviceOption(caffe2_pb2.CUDA, device)
+        with core.DeviceScope(device_opt):
+            with core.NameScope("gpu_{}".format(device)):
+                for param in model_helper_obj.GetParams():
+                    if param in model_helper_obj.param_to_float_copy:
+                        param_float = model_helper_obj.param_to_float_copy[param]
+                        model_helper_obj.HalfToFloat(param, param_float)
+
     if optimize_gradient_memory:
         _OptimizeGradientMemoryDEPRECATED(
             model_helper_obj, losses_by_gpu, devices
@@ -538,7 +551,6 @@ def _AllReduce(devices, model, net, param, use_nccl=False, control_input=None):
 def _SyncParams(devices, model, net, unique_param_names):
     for param in unique_param_names:
         _Broadcast(devices, model, net, param)
-
 
 def _AddDistributedParameterSync(
     devices,
--- a/caffe2/python/data_parallel_model.py
+++ b/caffe2/python/data_parallel_model.py
@@ -198,15 +198,7 @@ def Parallelize_GPU(
     # if in a multi-precision environment, make sure that the broadcast parameters
     # (low-precision) are still in sync with the high-precision copies.
     # Otherwise training goes badly
-    # todo(slayton): Work with non-fp16 low-precision copies
-    for device in devices:
-        device_opt = core.DeviceOption(caffe2_pb2.CUDA, device)
-        with core.DeviceScope(device_opt):
-            with core.NameScope("gpu_{}".format(device)):
-                for param in model_helper_obj.GetParams():
-                    if param in model_helper_obj.param_to_float_copy:
-                        param_float = model_helper_obj.param_to_float_copy[param]
-                        model_helper_obj.HalfToFloat(param, param_float)
+    _UpdateFloatCopies(devices, model_helper_obj)
 
     if optimize_gradient_memory:
         _OptimizeGradientMemoryDEPRECATED(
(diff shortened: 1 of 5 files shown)


caffe2/python/data_parallel_model.py
```diff
@@ -185,6 +185,19 @@ def Parallelize_GPU(
 
     _SyncParams(devices, model_helper_obj, model_helper_obj.param_init_net)
 
+    # if in a multi-precision environment, make sure that the broadcast parameters
```
asaadaldien: Move this to a method _UpdateFloatCopies for example?

caffe2/python/examples/cifar10_fp16.py
```diff
@@ -0,0 +1,210 @@
...
+
+def AddInput(model, batch_size, db, db_type, param_type="float", is_test=False):
+    """Adds the data input part."""
+    reader = model.CreateDB(
+        "reader_"+db,
+        db=db,
+        db_type=db_type)
+
+    # Note: Using C2 datum for this test
+    data, label = model.ImageInput(
```
asaadaldien: `brew.image_input(model,....` ?

caffe2/python/examples/cifar10_fp16.py
```diff
@@ -0,0 +1,210 @@
...
+    )
+    data = model.StopGradient(data, data)
+
+    return data, label
+
+# Move conv1 to fp16 only
+def AddCifar10Model(model, data, param_type):
+
+    with brew.arg_scope([brew.conv, brew.fc], dtype="float16"):
+        conv1 = brew.conv(model, data, 'conv1', 3, 32, 5,
```
asaadaldien: Please use max 80 character / line lint rule in this file.

caffe2/python/examples/cifar10_fp16.py
```diff
@@ -0,0 +1,210 @@
...
+    return accuracy
+
+def AddTrainingOperators(model, softmax, label, using_snapshot, use_fp16=False):
+    """Adds training operators to the model."""
+    xent = model.LabelCrossEntropy([softmax, label], 'xent')
+    loss = model.AveragedLoss(xent, "loss")
+
+    AddAccuracy(model, softmax, label)
+
+    print("adding gradient ops")
```
asaadaldien: Please use logging with debugging verbosity level (logger.info) instead of prints.

caffe2/python/examples/cifar10_fp16.py
```diff
@@ -0,0 +1,210 @@
...
+        data, label = AddInput(
+            train_model, batch_size=100,
+            db='cifar10_train_lmdb',
+            db_type='lmdb', param_type=param_type
+            )
+
+        softmax = AddCifar10Model(train_model, data, param_type)
+
+        AddTrainingOperators(train_model, softmax, label, param_type)
+        # AddBookkeepingOperators(train_model)
```
asaadaldien: Remove uncommented line.

caffe2/python/examples/cifar10_fp16.py
```diff
@@ -0,0 +1,210 @@
...
+
+    train_model.net.Proto().type = 'dag'
```
asaadaldien: Consider setting network type and num_workers as a command line input arguments.

caffe2/python/examples/cifar10_fp16.py
```diff
@@ -0,0 +1,210 @@
...
+    model.Print('loss', [], to_file=1)
+    for param in model.params:
+        model.Summarize(param, [], to_file=1)
+        model.Summarize(model.param_to_grad[param], [], to_file=1)
+
+def createTrainModel(param_type="float"):
+    order_scope = { 'order' : 'NCHW' }
+    train_model = ModelHelper(name="cifar10_quick_train", arg_scope=order_scope)
+
+    with core.DeviceScope(core.DeviceOption(caffe2_pb2.CUDA, 0)):
```
asaadaldien: This example will work only on GPU. Why not having device as a command line input argument so users  can run it on both CPU & GPU (HalfToFloat & FloaToHalf has CPU/CUDA implementations).

asaadaldien (changes requested):

slayton58:
I've updated data_parallel_model.py based on your comments, the cifar10 code was intended to be an example of how to use those additions to train in fp16. If you want it to actually be included I can fix it up based on your comments, but that wasn't the original intention.

caffe2/python/examples/cifar10_fp16.py
slayton58: What exactly is the line lint rule? I don't see any utilities supplied with the code for me to lint myself...

caffe2/python/examples/cifar10_fp16.py
slayton58: This is explicitly a GPU test. AFAIK CPU operators don't have any fp16 implementations.

caffe2/python/examples/cifar10_fp16.py
```diff
@@ -0,0 +1,244 @@
...
+def AddTrainingOperators(model, softmax, label, param_type="float"):
+    """Adds training operators to the model."""
+    xent = model.LabelCrossEntropy([softmax, label], 'xent')
+    loss = model.AveragedLoss(xent, "loss")
+
+    AddAccuracy(model, softmax, label)
+
+    model.AddGradientOperators([loss])
+
+    ITER = brew.iter(model, "ITER")
```
salexspb, line 77: why not use optimizers.py  instead of this hardcoded procedure?

caffe2/python/model_helper.py
```diff
@@ -143,6 +143,7 @@ def __init__(self, name=None, init_params=True, allow_not_known_ops=True,
             'use_cudnn': True,
             'cudnn_exhaustive_search': False,
         }
+        self.param_to_float_copy = {}
```
salexspb, line 146: this doesn't seem to be very generic to put into the main modeling class and then having all the optimizers and modeling helpers access it... You also have quite a bit of copy paste in each of the helpers (fc and conv in this case)
Seems like you just need to have some metadata on parameters. There is a ParamterInfo class, but we are rethinking it right now. Parameters can have a bunch of things associated with them. Including optimizer specific information which seems you need here.


Another approach I thought of is that if BlobReference has some info  of  anticipated type of the blob. Then you could even automatically insert conversions to and from fp16 when operations like MomentumSGD are called. 

I would like to chat with Andrey Malevich from our team who is working on new parameter abstractions for ModelHelper class and see if this kind of info fits into his new stuff.

salexspb (changes requested):

slayton58:
This work was done in a different way and merge, closing.

PR #611, opened 2017-05-19
bug fix in CMakeLists.txt (CAFFE2_CPU_FLAGS and CAFFE2_WHITELIST)

caiyang:
Fixed a bug in CMakeLists.txt: should not use option cmd for setting the initial value(empty string) of CAFFE2_CPU_FLAGS and CAFFE2_WHITELIST, because option can only be used for boolean(ON/OFF) variables. Use set cmd instead. The bug can cause compilation errors if CAFFE_CPU_FLAGS is set to ON, since an invalid 'ON' flag will be added to CXX_FLAGS. (2) Add build_* in .gitignore to allow multiple build directories in repo


herry13:
@caiyang your pull request has a similar build error comparing to mine (#609). Do you know what is the cause?

Yangqing:
Thanks guys - the build error seems to be out-of-time error because appveyor sets 1 hour as the maximum. We are in the process of dealing with such issues. Both PRs look great and I'll merge them once I am back in office; let me self assign first.

PR #617, opened 2017-05-19
Fix LRN schema for cuDNN op

slayton58:
Correct schema generation was previously broken leading to invalid gradient op creation.

Also exhibited in model_device_helper, where invalid schema were being created on the CPU when kwargs['engine'] == 'CUDNN'

### DIFF
--- a/caffe2/python/helpers/normalization.py
+++ b/caffe2/python/helpers/normalization.py
@@ -5,12 +5,15 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-from caffe2.python import core
+from caffe2.python import core, scope
 
 
 def lrn(model, blob_in, blob_out, order="NCHW", use_cudnn=False, **kwargs):
     """LRN"""
-    if use_cudnn:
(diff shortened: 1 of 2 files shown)


caffe2/python/helpers/normalization.py
```diff
@@ -5,12 +5,15 @@
...
-from caffe2.python import core
+from caffe2.python import core, scope
 
 
 def lrn(model, blob_in, blob_out, order="NCHW", use_cudnn=False, **kwargs):
     """LRN"""
-    if use_cudnn:
+    dev = kwargs['device_option'] if 'device_option' in kwargs \
+        else scope.CurrentDeviceScope()
+    is_cpu = dev is None or dev.device_type == caffe2_pb2.CPU
```
akyrola, line 16: this does not compile since caffe2_pb2 is not imported

slayton58: Interesting - is working fine on my end.. Will add import momentarily.

akyrola: actually our automatic linter noticed that problem. Weird that it works for you...!

PR #630, opened 2017-05-22
Option to use NCCL for broadcast

slayton58:
Fixes some performance issues when `broadcast_computed_params=True` is passed to Parallelize_GPU. Enabled via the same `use_nccl` flag as AllReduce

### DIFF
--- a/caffe2/python/data_parallel_model.py
+++ b/caffe2/python/data_parallel_model.py
@@ -143,7 +143,7 @@ def Parallelize_GPU(
 
     log.info("Add gradient all-reduces for SyncSGD")
     if broadcast_computed_params:
-        _BroadcastComputedParams(devices, model_helper_obj, rendezvous)
+        _BroadcastComputedParams(devices, model_helper_obj, rendezvous, use_nccl)
 
     if len(model_helper_obj._grad_names) > 0:
         _AllReduceGradients(devices, model_helper_obj, rendezvous, use_nccl)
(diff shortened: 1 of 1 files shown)


caffe2/python/data_parallel_model.py
```diff
@@ -484,10 +484,24 @@ def FinalizeAfterCheckpoint(model, blobs=None):
...
+def _Broadcast(devices, model, net, param, use_nccl=False):
     # TODO(akyrola): replace with NCCLBroadcast when it's working
     # Copy params from gpu_0 to other
     master_gpu = devices[0]
+
+    if use_nccl:
+        if _IsGPUBlob(model, param):
+            master_device_opt = core.DeviceOption(caffe2_pb2.CUDA, master_gpu)
+            with core.DeviceScope(master_device_opt):
+                log.info('Use NCCL for Broadcast')
```
akyrola: this is quite noisy, log.debug( is better.

akyrola:
So previously we had problem specifically with NCCLBroadcast. Forgot what the issue was. Does it work now?

caffe2/python/data_parallel_model.py
slayton58: I could have sworn there was an equivalent call for NCCL allreduce, which is why I put it in... Appears there isn't, so removed the output.

slayton58:
I'm having zero issues in my tests, I'll have to check again explicitly if non-zero roots work (if I recall correctly that was the issue previously)

akyrola:
@slayton58 , did you test with "non-zero roots"?

slayton58:
My basic tests (force `use_nccl=True` in data_parallel_model.py, set `root=devices[1]` in `_Broadcast` and run data_parallel_model_test.py) work fine with no issues.

PR #633, opened 2017-05-23
Bugfix for data_parallel_model - add Copy ops

lukeyeager:
Bug was exposed by this test:
```
caffe2/python/data_parallel_model_test.py::GPUDataParallelModelTest::test_equiv
```
Fixes this error:
```
========= Invalid __global__ read of size 4
=========     at 0x000000a0 in caffe2::math::_Kernel_float_Add(int, float const *, float const *, float*)
E0522 10:33:44.497241 18450 context_gpu.h:126] Encountered CUDA error: unspecified launch failure
F0522 10:33:44.497547 18450 operator.h:200] Computation on device returned error in operator
input: "gpu_0/fc_b_grad" input: "gpu_1/fc_b_grad" output: "gpu_0/fc_b_grad" name: "dpm" type: "Sum" device_option { device_type: 1 cuda_gpu_id: 0 }
```

### DIFF
--- a/caffe2/python/data_parallel_model.py
+++ b/caffe2/python/data_parallel_model.py
@@ -513,10 +513,10 @@ def _AllReduce(devices, model, net, param, use_nccl=False, control_input=None):
         d2 = model._devices[d2i]
         device_opt = core.DeviceOption(caffe2_pb2.CUDA, d1)
         with core.DeviceScope(device_opt):
-            net.Sum(
-                [blobs_group[d1], blobs_group[d2]], [blobs_group[d1]],
-                name="dpm",
-            )
+            # Copy from d2 to d1
+            d2_copy = 'gpu_{}/{}_copy'.format(d1, param)
+            model.Copy(blobs_group[d2], d2_copy)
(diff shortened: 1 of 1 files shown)


akyrola:
I have never seen this error, and we run this code a lot.

lukeyeager:
Do you ever run it without NCCL and without GLOO?

akyrola:
Yes. We disable NCCL by default. No-one uses NCCL currently (unless using Gloo). I wonder if your GPU setup incorrect somehow?

lukeyeager:
Looks like you might be right. I'll close this assuming it's a problem with my machine for now (I may reopen if I discover something). Thanks for the review!

lukeyeager:
@akyrola have you ever tested with non-heterogeneous GPUs? The error was because my GPUs don't have P2P access.

akyrola:
No we haven’t! Good catch. Is it possible to add a check whether there is uniform access or not and use different reduction then? I am concerned of using Copy always, since it will slow things down.

lukeyeager:
> I am concerned of using Copy always, since it will slow things down.

Agreed. I'll look into it now.

PR #636, opened 2017-05-23
Allow running predictor without specifying outputs.

kyu-sz:
Sometimes we need result from intermediate layers rather than the external output layers. This can be done by getting blobs from the workspace. By allowing running the predictor without specifying outputs, we do not need to declare the output tensors in advance. We can just simply leave the outputs parameter as null.

While this can also be achieved by removing the redundant layers in the network, in some cases we need both versions in a same program, and this is not flexible.

### DIFF
--- a/caffe2/core/predictor.cc
+++ b/caffe2/core/predictor.cc
@@ -48,9 +48,11 @@ void Predictor::run(const TensorVector& inputs, TensorVector* outputs) {
 
   CAFFE_ENFORCE(ws_.RunNet(run_net_.name()));
 
-  outputs->resize(run_net_.external_output_size());
-  for (auto i = 0; i < outputs->size(); ++i) {
-    (*outputs)[i] = extractOutputTensor(&ws_, run_net_.external_output(i));
+  if (outputs != nullptr) {
+    outputs->resize(run_net_.external_output_size());
+    for (auto i = 0; i < outputs->size(); ++i) {
+      (*outputs)[i] = extractOutputTensor(&ws_, run_net_.external_output(i));
+    }
   }
 }
 }
--- a/caffe2/core/predictor.h
+++ b/caffe2/core/predictor.h
@@ -24,7 +24,7 @@ class Predictor {
 
   // Postcondition:
   //   outputs->size() == run_net.external_inputs.size()
-  void run(const TensorVector& inputs, TensorVector* outputs);
+  void run(const TensorVector& inputs, TensorVector* outputs = nullptr);
 
   const NetDef& def() const {
     return run_net_;


Yangqing:
@salexpb Alex, would you like to take a look? This looks reasonable.

@kyu-sz you might want to explicitly set the output to the intermediate blobs you intend to access though - this allows us to further do optimizations such as graph rewriting with intermediate blobs being reused for memory optimization.

salexspb:
I actually wanted to ask people take on all these external outputs and external inputs. To me they look redundant as analyzing net allows to compute those. They usually are not automatically computed and relying on them being correctly specified in Net's proto can be dangerous IMO.

Specifically in the Predictor case, how do you know where is which prediction and how to deal with it, if you don't know output names?

caffe2/core/predictor.cc
```diff
@@ -48,9 +48,11 @@ void Predictor::run(const TensorVector& inputs, TensorVector* outputs) {
 
   CAFFE_ENFORCE(ws_.RunNet(run_net_.name()));
 
-  outputs->resize(run_net_.external_output_size());
-  for (auto i = 0; i < outputs->size(); ++i) {
-    (*outputs)[i] = extractOutputTensor(&ws_, run_net_.external_output(i));
+  if (outputs != nullptr) {
```
salexspb, line 51: did you mean outputs == nullptr?

salexspb (changes requested):

caffe2/core/predictor.cc
kyu-sz: No, here I mean if outputs are not specified, then it is by default nullptr, and the following operations should not be executed.

salexspb: Hm, sorry, I was confused. 

What is your usecase, do you run from c++ ? If so, what does this simple predictor give you? You could just run a net the same way. You will have to prepare inputs, but you need to do so in any way.

kyu-sz: Yes, I am running from C++. Actually I am not using this predictor in my project now, but directly forwarding my net from the workspace, with inputs and outputs transferred using the "CopyFrom" API. I am using a network transferred from Caffe, and if I initialize it on GPU, I find it cannot be used from a predictor because all layers, including the "data" layer and the output layers, contain TensorCUDA, but the predictor requires the IO blobs to contain TensorCPU.

Maybe I have some misunderstanding of the purpose of this predictor. If this modification conflicts with your design purpose, just omit it.

salexspb: This predictor is not very smart actually, as you can see it doesn't do much logic on top of normal Caffe2 Net interfaces. Which are perfectly fine to use in your c++ code directly. 

So if it doesn't give you too  much if would indeed have the Predictor code as simple as possible. Idea of predictor is that you don't need to access anything except inputs and outputs. If you do - you better to work with a Net interfaces IMO.

Its original purpose was to serve these two use-case:

1. Mobile(lightweight simple Predictor)
2. Python bindings

We plan to open-source another predictor which does smarter logic with threading and initialization nets and uses predictor_exporter single file format (see corresponding file in the repo)

kyu-sz: Oh I see. Thank you for such detailed explanation. I agree that to keep it simple, users shall not need to access the workspace, then outputs are necessary to be specified.

I will close this PR.

Issue #349, opened 2017-04-21
make_mnist_db doesn't generate db files


yiwsun:
When running tutorial MNIST.ipynb, the function GenerateDB() runs fine, no error reported but it did not generate files mnist-train-nchw-leveldb or mnist-test-nchw-leveldb.

While calling make_mnist_db directly from command line, it reported an error:
* *Caffe2 flag error: Cannot convert argument to bool: --db
Note that if you are passing in a bool flag, you need to explicitly specify it, like --arg=True or --arg True. Otherwise, the next argument may be inadvertently used as the argument, causing the above error.
Caffe2 flag: illegal argument: --channel_first*

Any thoughts how to fix it?

I am using Caffe2 on windows server 2016 with python2.7. Running tutorial Basics.ipynb works.

aaronmarkham:
There's some output from the function that shows the binary command and params. Can you copy that here? I'm thinking file path issues...

yiwsun:
Here is the output in the Jupyter notebook run:
Looks like you ran this before, so we need to cleanup those old workspace files...
DB name:  C:\Users\sqlpythonadmin\Source\caffe2\caffe2\python\tutorials\tutorial_data\mnist\mnist-train-nchw-leveldb
Creating database with:  C:\Users\sqlpythonadmin\Source\caffe2\build\caffe2\binaries\Release\make_mnist_db.exe --channel_first --db leveldb --image_file C:\Users\sqlpythonadmin\Source\caffe2\caffe2\python\tutorials\tutorial_data\mnist\train-images-idx3-ubyte --label_file C:\Users\sqlpythonadmin\Source\caffe2\caffe2\python\tutorials\tutorial_data\mnist\train-labels-idx1-ubyte --output_file C:\Users\sqlpythonadmin\Source\caffe2\caffe2\python\tutorials\tutorial_data\mnist\mnist-train-nchw-leveldb
DB name:  C:\Users\sqlpythonadmin\Source\caffe2\caffe2\python\tutorials\tutorial_data\mnist\mnist-test-nchw-leveldb
[...]

PR #640, opened 2017-05-24
Clean up binary build cmake script

Yangqing:
After the change we will be able to simply define targets and find dependencies.

Test: built locally.


caffe2/CMakeLists.txt
```diff
@@ -223,20 +222,41 @@ if (BUILD_PYTHON)
...
+  # Caffe2 proto files
+  install(DIRECTORY ${CMAKE_BINARY_DIR}/caffe2 DESTINATION ${CMAKE_INSTALL_PREFIX}
+          FILES_MATCHING PATTERN "*.py")
+endif()
+
+
+# ---[ Any builds that should happen after the main targets should be added here.
+# Binaries
+if (BUILD_BINARY)
+  add_subdirectory(binaries)
```
lukeyeager, line 260: While you're at it, is it an option to install things to `./bin/` like most other projects instead of to `./binaries/`?

Yangqing: Yup, sounds good, will do.

QinZiwen: so cool

aaronmarkham: Did this change maybe result in discrepancies in where we call certain binary files from tutorials? See this comment:
https://github.com/caffe2/caffe2/issues/349#issuecomment-309502165

Issue #654, opened 2017-05-24
Mac OS X build instructions are not right for protobufs


ezyang:
Steps to reproduce:

Follow the instructions at https://caffe2.ai/docs/getting-started.html?platform=mac&configuration=compile for Mac OS X build from source. The most important things:

1. `brew install protobufs`. At the time of writing the protobufs that brew installs by default is 3.3.0
2. Run `conda install -y --channel https://conda.anaconda.org/conda-forge  gflags glog  numpy protobuf=3.2.0` (as specified in the docs)
3. `cmake -DUSE_CUDA=OFF ..` and `make`

Expected result: Successful build

Actual result: You get the error:

```
In file included from /Users/ezyang/Dev/caffe2/build/caffe/proto/caffe.pb.cc:5:
/Users/ezyang/Dev/caffe2/build/caffe/proto/caffe.pb.h:17:2: error: This file was
      generated by an older version of protoc which is
#error This file was generated by an older version of protoc which is
 ^
/Users/ezyang/Dev/caffe2/build/caffe/proto/caffe.pb.h:18:2: error: incompatible
      with your Protocol Buffer headers. Please
#error incompatible with your Protocol Buffer headers.  Please
 ^
/Users/ezyang/Dev/caffe2/build/caffe/proto/caffe.pb.h:19:2: error: regenerate
      this file with a newer version of protoc.
#error regenerate this file with a newer version of protoc.
 ^
3 errors generated.
```

This is actually the same error as https://github.com/caffe2/caffe2/issues/474 but I want to reopen the issue because there is legitimately a problem with caffe2 and its build documentation. I can see several problems:

1. The documentation on the build page suggests installing protobufs with both brew and Anaconda. There are multiple levels of wrongness here. First, brew installs 3.3.0, and Anaconda installs 3.2.0. That can't be right. Second, if you run both instructions, you end up with TWO copies of protobufs; only one is necessary. Finally, these aren't even the right version of protobufs for HEAD: HEAD needs 3.1.0 as of https://github.com/caffe2/caffe2/commit/d9e90a968d29116d9a60e61f7f358de7aef84498

2. cmake should test the version of protobufs, and error if it's the wrong version. That will give a lot more interpretable error, because you will be able to see what the wrong version is.

[...]

PR #673, opened 2017-05-25
Improve protobuf OS X install instructions.

ezyang:
Partial fix for #654.

Signed-off-by: Edward Z. Yang <ezyang@fb.com>


aaronmarkham (changes requested):
The use of an environment seems like a great addition. 
Regarding protobuf, however I believe we still need the system protobuf AND the Python protobuf to be installed. These should be the same version to avoid conflicts.

aaronmarkham (approved):

PR #674, opened 2017-05-25
Fix data_parallel_model for inversed GPU indices

lukeyeager:
Adds a test for `devices=[1,0]` and fixes up DPM to make it pass.

### DIFF
--- a/caffe2/python/data_parallel_model.py
+++ b/caffe2/python/data_parallel_model.py
@@ -7,6 +7,7 @@ from __future__ import print_function
 from collections import OrderedDict
 import logging
 import copy
+import re
 
 from caffe2.python import model_helper, dyndep, scope, workspace, core, memonger
 from caffe2.proto import caffe2_pb2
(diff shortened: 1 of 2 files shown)


akyrola:
In my understanding DPM should not care about order of GPUs passed to it as it always loops over the devices. I believe probably just the test would be needed to be changed.

akyrola:
You are right, I broke this recently. Let me think a bit if we can have simpler solution.

akyrola:
I think this is fine, but complicates things a bit. I wonder if we could assert that we only accept devices in sorted order? Or is there a need for you to have them in different order?

lukeyeager:
I think I could always use `CUDA_VISIBLE_DEVICES` to switch the order for any necessary cases, so forcing them to be sorted seems like a perfectly acceptable solution.

lukeyeager:
I didn't really expect you to like the solution in this PR - I just find that a proof-of-concept fix is helpful for explaining a bug sometimes.

akyrola:
Yep, let’s force the sorted-ness, as otherwise these kind of errors will pop up later as well, as developer of data_parallel_model is unlikely to consider unsorted devices as happening

PR #690, opened 2017-05-27
Update centos.md

kyu-sz:
Add solution for trouble linking with "libCaffe2_GPU.so" due to "undefined reference to `ncclReduceScatter'".
Add "-DBUILD_SHARED_LIBS=ON" to glog building.


Yangqing:
Thanks for the update!

PR #697, opened 2017-05-30
Fp16 training initializers

slayton58:
Adds support for generating and training pfp16 models. Added SGD optimizer for multi-precision trainers and a new callback to data_parallel_model in order to help multi-precision models keep their different copies of parameters in sync during training.

### DIFF
--- a/caffe2/python/data_parallel_model.py
+++ b/caffe2/python/data_parallel_model.py
@@ -25,6 +25,7 @@ def Parallelize_GPU(
     forward_pass_builder_fun,
     param_update_builder_fun=None,
     optimizer_builder_fun=None,
+    post_sync_builder_fun=None,
     devices=range(0, workspace.NumCudaDevices()),
     rendezvous=None,
     net_type='dag',
@@ -57,7 +58,11 @@ def Parallelize_GPU(
                         Alternative to param_update_builder_fun, allows one
                         to add an optimizer for the whole model. Called only
                         once, without name or devicescope.
-
+      post_sync_builder_fun:
+                        Function applied after initial parameter sync has been
+                        completed, such as keeping multi-precision parameters
+                        in sync.
+                        Signature: post_sync_builder_fun(model)
       devices:          List of GPU ids, such as [0, 1, 2, 3],
       rendezvous:       used for rendezvous in distributed computation, if None
                         then only one node is used. To create rendezvous,
@@ -222,6 +227,15 @@ def Parallelize_GPU(
     _SyncParams(
         devices, model_helper_obj, model_helper_obj.param_init_net, sync_names
     )
+
+    # Handle any operations that need to be done after parameter sync
+    # i.e. making sure multi-precision copies of parameters are up-to-date
+    if post_sync_builder_fun is not None:
+        for device in devices:
+            device_opt = core.DeviceOption(caffe2_pb2.CUDA, device)
+            with core.DeviceScope(device_opt):
+                with core.NameScope("gpu_{}".format(device)):
+                    post_sync_builder_fun(model_helper_obj)
 
     if optimize_gradient_memory:
         _OptimizeGradientMemorySimple(model_helper_obj, losses_by_gpu, devices)
--- a/caffe2/python/helpers/conv.py
+++ b/caffe2/python/helpers/conv.py
@@ -63,12 +63,12 @@ def _ConvBase(
 
     if model.init_params:
         weight = model.create_param(
-            param_name=blob_out+'_w',
+            param_name=blob_out + '_w',
             shape=weight_shape,
             initializer=WeightInitializer)
         if use_bias:
             bias = model.create_param(
-                param_name=blob_out+'_b',
+                param_name=blob_out + '_b',
                 shape=[dim_out, ],
                 initializer=BiasInitializer)
     else:
--- a/caffe2/python/helpers/conv.py
+++ b/caffe2/python/helpers/conv.py
@@ -18,8 +18,8 @@ def _ConvBase(
     kernel,
     weight_init=None,
     bias_init=None,
-    weight_initializer=None,
-    bias_initializer=None,
+    WeightInitializer=None,
+    BiasInitializer=None,
     group=1,
     transform_inputs=None,
     use_cudnn=False,
@@ -54,23 +54,23 @@ def _ConvBase(
         weight_shape.extend(kernels)
         weight_shape.append(int(dim_in / group))
 
-    weight_initializer = initializers.update_initializer(
-        weight_initializer, weight_init, ("XavierFill", {})
+    WeightInitializer = initializers.update_initializer(
+        WeightInitializer, weight_init, ("XavierFill", {})
     )
-    bias_initializer = initializers.update_initializer(
-        bias_initializer, bias_init, ("ConstantFill", {})
+    BiasInitializer = initializers.update_initializer(
+        BiasInitializer, bias_init, ("ConstantFill", {})
     )
 
     if model.init_params:
         weight = model.create_param(
             param_name=blob_out+'_w',
             shape=weight_shape,
-            initializer=weight_initializer)
+            initializer=WeightInitializer)
         if use_bias:
             bias = model.create_param(
                 param_name=blob_out+'_b',
                 shape=[dim_out, ],
-                initializer=bias_initializer)
+                initializer=BiasInitializer)
     else:
         weight = core.ScopedBlobReference(
             blob_out + '_w', model.param_init_net)
@@ -149,8 +149,8 @@ def conv(
     kernel,
     weight_init=None,
     bias_init=None,
-    weight_initializer=None,
-    bias_initializer=None,
+    WeightInitializer=None,
+    BiasInitializer=None,
     group=1,
     transform_inputs=None,
     **kwargs
@@ -158,7 +158,7 @@ def conv(
     """2-dimensional convolution.
     """
     return _ConvBase(model, False, blob_in, blob_out, dim_in, dim_out, kernel,
-                     weight_init, bias_init, weight_initializer, bias_initializer, group, transform_inputs, **kwargs)
+                     weight_init, bias_init, WeightInitializer, BiasInitializer, group, transform_inputs, **kwargs)
 
 
 def conv_transpose(
--- a/caffe2/python/helpers/conv.py
+++ b/caffe2/python/helpers/conv.py
@@ -6,7 +6,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 from caffe2.python import core
-
+from caffe2.python.modeling import initializers
 
 def _ConvBase(
     model,
@@ -18,6 +18,8 @@ def _ConvBase(
     kernel,
     weight_init=None,
     bias_init=None,
+    weight_initializer=None,
+    bias_initializer=None,
     group=1,
     transform_inputs=None,
     use_cudnn=False,
@@ -43,8 +45,6 @@ def _ConvBase(
 
     use_bias =\
             False if ("no_bias" in kwargs and kwargs["no_bias"]) else True
-    weight_init = weight_init if weight_init else ('XavierFill', {})
-    bias_init = bias_init if bias_init else ('ConstantFill', {})
     blob_out = blob_out or model.net.NextName()
     weight_shape = [dim_out]
     if order == "NCHW":
@@ -54,20 +54,23 @@ def _ConvBase(
         weight_shape.extend(kernels)
         weight_shape.append(int(dim_in / group))
 
+    weight_initializer = initializers.update_initializer(
+        weight_initializer, weight_init, ("XavierFill", {})
+    )
+    bias_initializer = initializers.update_initializer(
+        bias_initializer, bias_init, ("ConstantFill", {})
+    )
+
     if model.init_params:
-        weight = model.param_init_net.__getattr__(weight_init[0])(
-            [],
-            blob_out + '_w',
+        weight = model.create_param(
+            param_name=blob_out+'_w',
             shape=weight_shape,
-            **weight_init[1]
-        )
+            initializer=weight_initializer)
         if use_bias:
-            bias = model.param_init_net.__getattr__(bias_init[0])(
-                [],
-                blob_out + '_b',
+            bias = model.create_param(
+                param_name=blob_out+'_b',
                 shape=[dim_out, ],
-                **bias_init[1]
-            )
+                initializer=bias_initializer)
     else:
         weight = core.ScopedBlobReference(
             blob_out + '_w', model.param_init_net)
@@ -146,6 +149,8 @@ def conv(
     kernel,
     weight_init=None,
     bias_init=None,
+    weight_initializer=None,
+    bias_initializer=None,
     group=1,
     transform_inputs=None,
     **kwargs
@@ -153,7 +158,7 @@ def conv(
     """2-dimensional convolution.
     """
     return _ConvBase(model, False, blob_in, blob_out, dim_in, dim_out, kernel,
-                     weight_init, bias_init, group, transform_inputs, **kwargs)
+                     weight_init, bias_init, weight_initializer, bias_initializer, group, transform_inputs, **kwargs)
 
 
 def conv_transpose(
--- a/caffe2/python/helpers/conv.py
+++ b/caffe2/python/helpers/conv.py
@@ -54,10 +54,10 @@ def _ConvBase(
         weight_shape.extend(kernels)
         weight_shape.append(int(dim_in / group))
 
-    WeightInitializer = initializers.update_initializer(
+    weight_initializer = initializers.update_initializer(
         WeightInitializer, weight_init, ("XavierFill", {})
     )
-    BiasInitializer = initializers.update_initializer(
+    bias_initializer = initializers.update_initializer(
         BiasInitializer, bias_init, ("ConstantFill", {})
     )
 
@@ -65,12 +65,12 @@ def _ConvBase(
         weight = model.create_param(
             param_name=blob_out + '_w',
             shape=weight_shape,
-            initializer=WeightInitializer)
+            initializer=weight_initializer)
         if use_bias:
             bias = model.create_param(
                 param_name=blob_out + '_b',
                 shape=[dim_out, ],
-                initializer=BiasInitializer)
+                initializer=bias_initializer)
     else:
         weight = core.ScopedBlobReference(
             blob_out + '_w', model.param_init_net)
--- a/caffe2/python/modeling/initializers.py
+++ b/caffe2/python/modeling/initializers.py
@@ -30,7 +30,14 @@ class Initializer(object):
         )
 
 class pFP16Initializer(object):
-    def __init__(self, operator_name, **kwargs):
+    def __init__(self, operator_name=None, **kwargs):
+        self.operator_name = operator_name
+        self.operator_kwargs = kwargs
+        print("constructed with {}".format(self.operator_name))
+
+    def update(self, operator_name, kwargs):
+        if self.operator_name is not None:
+            raise Exception("Operator name overwrites are not allowed")
         self.operator_name = operator_name
         self.operator_kwargs = kwargs
 
@@ -47,10 +54,10 @@ class pFP16Initializer(object):
             param_id=None,
             param=param,
             shape=shape,
-            param_copy={'float' : param_fp32}
+            blob_copy={'float' : param_fp32}
         )
 
-def update_initializer(initializer,
+def update_initializer(initializer_class,
                        operator_name_and_kwargs,
                        default_operator_name_and_kwargs):
     '''
--- a/caffe2/python/modeling/initializers.py
+++ b/caffe2/python/modeling/initializers.py
@@ -29,10 +29,9 @@ class Initializer(object):
             shape=shape,
         )
 
-class pFP16Initializer(object):
+class pFP16Initializer(Initializer):
     def __init__(self, operator_name=None, **kwargs):
-        self.operator_name = operator_name
-        self.operator_kwargs = kwargs
+        super(pFP16Initializer, self).__init__(operator_name, **kwargs)
 
     def update(self, operator_name, kwargs):
         if self.operator_name is not None:
@@ -43,11 +42,11 @@ class pFP16Initializer(object):
     def create_param(self, param_name, init_net, shape):
         # create master fp32 copy
         param_fp32 = init_net.__getattr__(self.operator_name)(
-            [], param_name+'_fp32', shape=shape,
+            [], param_name + "_fp32", shape=shape,
             **self.operator_kwargs)
         # cast to fp16 copy
         param = init_net.FloatToHalf(
-            param_name+"_fp32", param_name)
+            param_fp32, param_name)
 
         return ParameterInfo(
             param_id=None,
--- a/caffe2/python/modeling/initializers.py
+++ b/caffe2/python/modeling/initializers.py
@@ -33,7 +33,6 @@ class pFP16Initializer(object):
     def __init__(self, operator_name=None, **kwargs):
         self.operator_name = operator_name
         self.operator_kwargs = kwargs
-        print("constructed with {}".format(self.operator_name))
 
     def update(self, operator_name, kwargs):
         if self.operator_name is not None:
--- a/caffe2/python/modeling/initializers.py
+++ b/caffe2/python/modeling/initializers.py
@@ -30,9 +30,6 @@ class Initializer(object):
         )
 
 class pFP16Initializer(Initializer):
-    def __init__(self, operator_name=None, **kwargs):
-        super(pFP16Initializer, self).__init__(operator_name, **kwargs)
-
     def update(self, operator_name, kwargs):
         if self.operator_name is not None:
             raise Exception("Operator name overwrites are not allowed")
--- a/caffe2/python/modeling/initializers.py
+++ b/caffe2/python/modeling/initializers.py
@@ -3,6 +3,7 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 from caffe2.python.modeling.parameter_info import ParameterInfo
+from caffe2.python.core import DataType
 
 
 class Initializer(object):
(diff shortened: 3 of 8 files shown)


caffe2/python/modeling/initializers.py
```diff
@@ -29,6 +29,33 @@ def create_param(self, param_name, init_net, shape):
             shape=shape,
         )
 
+class pFP16Initializer(object):
+    def __init__(self, operator_name=None, **kwargs):
+        self.operator_name = operator_name
+        self.operator_kwargs = kwargs
+
+    def update(self, operator_name, kwargs):
```
salexspb: why not inherit and then reuse base class method?

caffe2/python/modeling/initializers.py
```diff
@@ -29,6 +29,33 @@ def create_param(self, param_name, init_net, shape):
...
+        self.operator_kwargs = kwargs
+
+    def create_param(self, param_name, init_net, shape):
+        # create master fp32 copy
+        param_fp32 = init_net.__getattr__(self.operator_name)(
+            [], param_name+'_fp32', shape=shape,
+            **self.operator_kwargs)
+        # cast to fp16 copy
+        param = init_net.FloatToHalf(
+            param_name+"_fp32", param_name)
```
salexspb: nit: param_name + "_fp32"

caffe2/python/modeling/initializers.py
```diff
@@ -29,6 +29,33 @@ def create_param(self, param_name, init_net, shape):
...
+        self.operator_name = operator_name
+        self.operator_kwargs = kwargs
+
+    def update(self, operator_name, kwargs):
+        if self.operator_name is not None:
+            raise Exception("Operator name overwrites are not allowed")
+        self.operator_name = operator_name
+        self.operator_kwargs = kwargs
+
+    def create_param(self, param_name, init_net, shape):
```
salexspb, line 40: you could call base class method and then add a conversion part + modify ParameterInfo object returned from the base class.

caffe2/python/modeling/parameter_info.py
```diff
@@ -15,7 +15,7 @@ class ParameterType(object):
 class ParameterInfo(object):
     def __init__(
             self, param_id, param, key=None, shape=None, length=None,
-            grad=None):
+            grad=None, blob_copy=None):
```
salexspb, line 18: could you add a comment about what blob_copy is for? Also how about calling it "float_copy" ?

caffe2/python/modeling/initializers.py
```diff
@@ -29,6 +29,33 @@ def create_param(self, param_name, init_net, shape):
...
+            **self.operator_kwargs)
+        # cast to fp16 copy
+        param = init_net.FloatToHalf(
+            param_name+"_fp32", param_name)
+
+        return ParameterInfo(
+            param_id=None,
+            param=param,
+            shape=shape,
+            blob_copy={'float' : param_fp32}
```
salexspb: does it have to be a dict? Is there a usecase now where we need it? If not, should we make it simpler for now?

caffe2/python/optimizer.py
```diff
@@ -140,7 +139,7 @@ def _run(self, net, param_init_net, param_info):
 
         if self.momentum > 0:
             momentum_data = param_init_net.ConstantFill(
-                param, str(param) + "_momentum", value=0.)
+                [param], param + '_momentum', value=0.)
```
salexspb: I think @akyrola 's diff is almost ready which should solve this.

caffe2/python/optimizer.py
```diff
@@ -168,6 +167,61 @@ def scale_learning_rate(self, scale):
...
+        net.MomentumSGDUpdate(
+            [grad+"_fp32", momentum_data, lr, param_fp32],
+            [grad, momentum_data, param_fp32],
+            momentum=self.momentum,
+            nesterov=0)
+
+        # Copy updated param back to fp16
+        net.FloatToHalf(param_fp32, param)
+
+    def scale_learning_rate(self, scale):
```
salexspb: do you use this?

caffe2/python/optimizer.py
```diff
@@ -168,6 +167,61 @@ def scale_learning_rate(self, scale):
...
+
+        assert not isinstance(grad, core.GradientSlice), \
+                "Doesn't support sparse gradients"
+
+        # Copy gradient to fp32
+        net.HalfToFloat(grad, grad+"_fp32")
+
+        # update (fused) in fp32
+        net.MomentumSGDUpdate(
+            [grad+"_fp32", momentum_data, lr, param_fp32],
```
salexspb: nit: spaces around +

caffe2/python/optimizer.py
```diff
@@ -168,6 +167,61 @@ def scale_learning_rate(self, scale):
...
+
+        momentum_data = param_init_net.ConstantFill(
+            [param], param + "_momentum", value=0.)
+        self._aux_params.local.append(momentum_data)
+
+        assert not isinstance(grad, core.GradientSlice), \
+                "Doesn't support sparse gradients"
+
+        # Copy gradient to fp32
+        net.HalfToFloat(grad, grad+"_fp32")
```
salexspb: spaces around +

salexspb: nit: spaces. Also you could assign this to a BlobReference and use it later instead of relying on string matching

caffe2/python/optimizer.py
```diff
@@ -168,6 +167,61 @@ def scale_learning_rate(self, scale):
...
+        )
+
+        ONE = param_init_net.ConstantFill([], "ONE", shape=[1], value=1.0)
+        self._aux_params.shared.append(ONE)
+
+        momentum_data = param_init_net.ConstantFill(
+            [param], param + "_momentum", value=0.)
+        self._aux_params.local.append(momentum_data)
+
+        assert not isinstance(grad, core.GradientSlice), \
```
salexspb: can we reuse more code from the SgdOptimizer? And also then get this one for free. At least calls till this line could be split to a function in SgdOptimizer and here you could call it (no need to change other optimizers).

caffe2/python/optimizer.py
salexspb: do you use this anywhere?

caffe2/python/data_parallel_model.py
```diff
@@ -205,6 +211,15 @@ def Parallelize_GPU(
         devices, model_helper_obj, model_helper_obj.param_init_net, sync_names
     )
 
+    # Handle any operations that need to be done after parameter sync
+    # i.e. making sure multi-precision copies of parameters are up-to-date
+    if post_sync_builder_fun is not None:
```
salexspb, line 233: I would like @akyrola  to have a look into this part as I am not a DPM expert

salexspb (changes requested):

salexspb:
Btw, @akyrola also has a diff out which makes SgdOptimizer use fused MomentumSGDUpdate which probably should allow you to reuse SgdOptimizer code (his diff is accepted)

caffe2/python/modeling/initializers.py
slayton58: Because I don't like python's inheritance syntax :) Fixed.

caffe2/python/modeling/initializers.py
slayton58: Fixed.

caffe2/python/modeling/initializers.py
slayton58: I felt this was cleaner (plus see previous comment about not liking python's inheritance) - it's very explicit about creating the different precisions.

caffe2/python/modeling/initializers.py
slayton58: Doesn't have to be, but it makes sense to me that it is a dict - you also need to know what the different multi-precision types are when updating parameters, so this solves that problem too.

caffe2/python/modeling/parameter_info.py
slayton58: See above - it's intended as a general place to store copies, not just float.

caffe2/python/optimizer.py
slayton58: `net.FloatToHalf`? Yes. You update the fp32 copy of the weights, then truncate them down to fp16 for the next fprop + bprop pass

caffe2/python/optimizer.py
slayton58: Fixed.

caffe2/python/optimizer.py
salexspb: no, the function scale_learning_rate. I didn't see its usage in the lines you changed.

salexspb: omg, I hate github commenting capabilities.. Can I select a single line or a set of lines I need? Here I click on the line and it shows up as 4.. And I can't select code blocks.

slayton58: No, we don't use it anywhere, I assumed it was necessary for the interface, and so had to be provided.

caffe2/python/optimizer.py
slayton58: I don't really see how apart from abstracting out some more of the details (which I consider out of the scope of this patch set) -- I already call out to `SgdOptimizer` in the case of fp32 parameters.

caffe2/python/optimizer.py
slayton58: Fixed.

slayton58: Done.

caffe2/python/modeling/initializers.py
```diff
@@ -29,6 +29,32 @@ def create_param(self, param_name, init_net, shape):
             shape=shape,
         )
 
+class pFP16Initializer(Initializer):
+    def __init__(self, operator_name=None, **kwargs):
+        super(pFP16Initializer, self).__init__(operator_name, **kwargs)
+
+    def update(self, operator_name, kwargs):
```
salexspb: why do you still have this method?

caffe2/python/modeling/initializers.py
salexspb: What do you mean? An initializer here is essentially a normal one + fp16 conversion, why not reuse more code?

caffe2/python/modeling/parameter_info.py
salexspb: Thats not clear why is copy needed. Could you add a comment block, please?

caffe2/python/optimizer.py
salexspb: so do you need this change in your diff ? isn't it equivalent to what it was before?

caffe2/python/optimizer.py
salexspb: why? if you inherit you need only to provide things you override

caffe2/python/optimizer.py
```diff
@@ -379,6 +442,13 @@ def build_sgd(model, base_learning_rate, **kwargs):
     sgd_optimizer = SgdOptimizer(base_learning_rate, **kwargs)
     return _build(model, sgd_optimizer)
 
+def build_multi_precision_sgd(model, base_learning_rate, **kwargs):
+    sgd_optimizer = MultiPrecisionSgdOptimizer(base_learning_rate, **kwargs)
+    return _build(model, sgd_optimizer)
+    #for param_info in model.GetOptimizationParamInfo():
```
salexspb: remove extra comments?

caffe2/python/optimizer.py
```diff
@@ -177,6 +176,70 @@ def scale_learning_rate(self, scale):
...
+            learning_rate_blob=str(param) + "_lr",
+            policy=self.policy,
+            **(self.init_kwargs)
+        )
+
+        dev = scope.CurrentDeviceScope()
+        if dev is None:
+            dev = core.DeviceOption(caffe2_pb2.CPU)
+
+        ONE = param_init_net.ConstantFill(
```
salexspb: why do you need ONE? I don't see it being used below

caffe2/python/optimizer.py
```diff
@@ -177,6 +176,70 @@ def scale_learning_rate(self, scale):
...
+            return SgdOptimizer._run(self, net, param_init_net, param_info)
+
+        grad = param_info.grad
+        if self.base_learning_rate <= 0:
+            return
+
+        lr, _ = self.build_lr(
+            net, param_init_net,
+            base_learning_rate=-self.base_learning_rate,
+            learning_rate_blob=str(param) + "_lr",
```
salexspb: why do you need str() on param here?  when you sum up an object and a string it would call __str__ on it automatically

salexspb (changes requested):

caffe2/python/modeling/initializers.py
slayton58: Because right now it'll look like:

```
param_fp32_info = super(Initializer, self).create_param(
        param_name + "fp32",
        init_net,
        shape)

param = init_net.FloatToHalf(param_fp32_info.blob,
                                                param_name)

return ParameterInfo(
        param_id=None,
        param=param,
        shape=shape,
        blob_copy={'float' : param_fp32_info.blob})
```

That's not really cleaner in my opinion

caffe2/python/modeling/parameter_info.py
slayton58: I already added that in the body of the method. Would you prefer it prefixing the function? Inline with the signature?

caffe2/python/optimizer.py
slayton58: I believe this is actually still needed, otherwise I think the namescope will still be applied.

caffe2/python/optimizer.py
slayton58: Done.

caffe2/python/optimizer.py
slayton58: Removed.

caffe2/python/data_parallel_model.py
akyrola: This is fine

caffe2/python/optimizer.py
akyrola: Actually not needed anymore, if you use optimizer_builder_fun with data_parallel_model, as that is called without any Namescopes.

caffe2/python/optimizer.py
```diff
@@ -185,6 +184,53 @@ def scale_learning_rate(self, scale):
...
+
+        # Copy gradient to fp32
+        grad_fp32 = net.HalfToFloat(grad, grad + "_fp32")
+
+        # update (fused) in fp32
+        net.MomentumSGDUpdate(
+            [grad_fp32, momentum_data, lr, param_fp32],
+            [grad, momentum_data, param_fp32],
+            momentum=self.momentum,
+            nesterov=0)
```
akyrola: nesterov=self.nesterov

caffe2/python/optimizer.py
```diff
@@ -185,6 +184,53 @@ def scale_learning_rate(self, scale):
         self.base_learning_rate *= scale
         return
 
+class MultiPrecisionSgdOptimizer(SgdOptimizer):
+    def __init__(self, base_learning_rate, momentum, policy, **kwargs):
+        super(SgdOptimizer, self).__init__() # self, base_learning_rate, **kwargs)
+        self.base_learning_rate = base_learning_rate
+        self.momentum = momentum
+        self.policy = policy
+        self.init_kwargs = kwargs
```
akyrola: self.nesterov = nesterov

and add nesterov as parameter, default=0

caffe2/python/optimizer.py
```diff
@@ -387,6 +433,10 @@ def build_sgd(model, base_learning_rate, **kwargs):
     sgd_optimizer = SgdOptimizer(base_learning_rate, **kwargs)
     return _build(model, sgd_optimizer)
 
+def build_multi_precision_sgd(model, base_learning_rate, **kwargs):
+    sgd_optimizer = MultiPrecisionSgdOptimizer(base_learning_rate, **kwargs)
```
akyrola: nit:
  multiprec_sgd_optimizer = ...

caffe2/python/optimizer.py
slayton58: Fixed.

caffe2/python/optimizer.py
slayton58: Fixed.

caffe2/python/data_parallel_model.py
akyrola: @slayton58  looks like you don't use the post_sync_builder_fun anymore? Perhaps because optimizer_builder_fun is now possible? 

In that case, let's remove it.

caffe2/python/helpers/conv.py
```diff
@@ -18,6 +18,8 @@ def _ConvBase(
     kernel,
     weight_init=None,
     bias_init=None,
+    WeightInitializer=None,
+    BiasInitializer=None,
```
akyrola, line 22: why different casing for WeightInitializer and BiasInitializer?

caffe2/python/helpers/conv.py
```diff
@@ -54,20 +54,23 @@ def _ConvBase(
...
+    BiasInitializer = initializers.update_initializer(
+        BiasInitializer, bias_init, ("ConstantFill", {})
+    )
+
     if model.init_params:
-        weight = model.param_init_net.__getattr__(weight_init[0])(
-            [],
-            blob_out + '_w',
+        weight = model.create_param(
+            param_name=blob_out+'_w',
```
akyrola: nit: spaces around +

caffe2/python/helpers/conv.py
```diff
@@ -54,20 +54,23 @@ def _ConvBase(
...
             shape=weight_shape,
-            **weight_init[1]
-        )
+            initializer=WeightInitializer)
         if use_bias:
-            bias = model.param_init_net.__getattr__(bias_init[0])(
-                [],
-                blob_out + '_b',
+            bias = model.create_param(
+                param_name=blob_out+'_b',
```
akyrola: ditto

caffe2/python/optimizer.py
```diff
@@ -157,7 +156,7 @@ def _run(self, net, param_init_net, param_info):
 
         if self.momentum > 0:
             momentum_data = param_init_net.ConstantFill(
-                param, str(param) + "_momentum", value=0.)
+                param, str(param) + '_momentum', value=0.)
```
akyrola: unnecessary change?

caffe2/python/optimizer.py
```diff
@@ -387,6 +434,12 @@ def build_sgd(model, base_learning_rate, **kwargs):
     sgd_optimizer = SgdOptimizer(base_learning_rate, **kwargs)
     return _build(model, sgd_optimizer)
 
+def build_multi_precision_sgd(model, base_learning_rate, **kwargs):
+    multi_prec_sgd_optimizer = MultiPrecisionSgdOptimizer(
+            base_learning_rate, **kwargs
+    )
+    return _build(model, sgd_optimizer)
```
akyrola: doesn't compile... sgd_optimizer not defined anymore

akyrola (commented):
A couple of more comments, and I don't think this currently works (see comment).

caffe2/python/data_parallel_model.py
slayton58: No, I still need it. `optimizer_builder_fun` happens here: https://github.com/caffe2/caffe2/blob/master/caffe2/python/data_parallel_model.py#L183-L191 while `SyncParams` happens later, here: https://github.com/caffe2/caffe2/blob/master/caffe2/python/data_parallel_model.py#L222-L224

My sync needs to happen after the `SyncParams` call

caffe2/python/helpers/conv.py
slayton58: Ask @salexspb :) He wanted the CamelCase for the *Initializer to indicate they're classes being passed in, not objects.

caffe2/python/helpers/conv.py
slayton58: Fixed.

caffe2/python/helpers/conv.py
slayton58: Fixed.

caffe2/python/optimizer.py
slayton58: Fixed.

caffe2/python/modeling/parameter_info.py
```diff
@@ -26,6 +26,9 @@ def __init__(
         self.length = max(1, length if length is not None else 1)
         self.grad = grad
         self._cloned_init_net = None
+        # Optionally store equivalent copies of the blob
+        # in different precisions (i.e. half and float copies)
+        self.blob_copy = blob_copy
```
salexspb: could you notice here that it is a dict?

caffe2/python/optimizer.py
```diff
@@ -185,6 +184,54 @@ def scale_learning_rate(self, scale):
...
+        super(SgdOptimizer, self).__init__()
+        self.base_learning_rate = base_learning_rate
+        self.momentum = momentum
+        self.policy = policy
+        self.nesterov = nesterov
+        self.init_kwargs = kwargs
+
+    def _run(self, net, param_init_net, param_info):
+        param = param_info.blob
+        param_fp32 = param_info.blob_copy['float'] if param_info.blob_copy is not None else None
```
salexspb: could you replace 'float' with DataType enum in core.py?

caffe2/python/optimizer.py
```diff
@@ -185,6 +184,54 @@ def scale_learning_rate(self, scale):
...
+
+        # If we have a straight fp32 parameter, run the base class
+        if param_fp32 == None:
+            return SgdOptimizer._run(self, net, param_init_net, param_info)
+
+        grad = param_info.grad
+        if self.base_learning_rate <= 0:
```
salexspb: what is this if statement for? IMO this is error prone. is this an input error check? Then should we through otherwise?

caffe2/python/optimizer.py
```diff
@@ -185,6 +184,54 @@ def scale_learning_rate(self, scale):
...
+        self._aux_params.local.append(momentum_data)
+
+        assert not isinstance(grad, core.GradientSlice), \
+                "Doesn't support sparse gradients"
+
+        # Copy gradient to fp32
+        grad_fp32 = net.HalfToFloat(grad, grad + "_fp32")
+
+        # update (fused) in fp32
+        net.MomentumSGDUpdate(
```
salexspb, line 230: Not suggesting to do it in this PR, but going forward it would be nice to have a way to convert an optimizer to fp16 format without replicating its  code.  What do you think?

salexspb (commented):
Just a few small things and should be good to go!

caffe2/python/helpers/conv.py
```diff
@@ -54,20 +54,23 @@ def _ConvBase(
         weight_shape.extend(kernels)
         weight_shape.append(int(dim_in / group))
 
+    WeightInitializer = initializers.update_initializer(
```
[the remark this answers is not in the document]
slayton58: Done.

caffe2/python/modeling/parameter_info.py
slayton58: Done.

caffe2/python/optimizer.py
slayton58: A net will generally have some parameters in fp16, some in float (i.e. Resnets where all filters are fp16, but batch norm parameters are all fp32) - this just drops back to the standard fp32 update path if an fp32 parameter is encountered.

caffe2/python/optimizer.py
slayton58: It would be - I'd like to discuss that with you in fact.

There's a couple of ways of doing it.

caffe2/python/optimizer.py
salexspb: I am talking here about line 205. It is the one I clicked on when leaving this comment. How do I make commenting here less confusing for you? :)

caffe2/python/optimizer.py
slayton58: Done.

caffe2/python/optimizer.py
slayton58: This is to match the behaviour of the SgdOptimizer (line 127)

salexspb: It has == 0, not <= 0. < 0 sounds like a user error.

slayton58: I see, typo. Fixed.

salexspb:
Looks good code wise! 

What about testing? We already have a test for initializers and optimizers, could you look into adding these new guys in there?

slayton58:
Added (fairly basic) tests

caffe2/python/optimizer_test.py
```diff
@@ -21,6 +23,19 @@ def check_optimizer(self, optimizer):
             np.testing.assert_allclose(np.array([1.0]), tensor, atol=1e-5)
 
 
+class TestMultiPrecisionSgd(OptimizerTestBase, TestCase):
+    def build_optimizer(self, model):
+        self._skip_gpu = False
+        return build_multi_precision_sgd(model, base_learning_rate=0.1)
```
salexspb, line 31: will this actually exercise the logic? I think if you don't use an fp16 initializer , then you will just fallback to the normal sgd...

caffe2/python/optimizer_test.py
```diff
@@ -21,6 +23,19 @@ def check_optimizer(self, optimizer):
...
+
+    def check_optimizer(self, optimizer):
```
salexspb, line 33: could you actually test that we at least have those params with a copy and conversion ops in place?

salexspb (changes requested):
Could you improve the optimizer test to cover the new logic, please?

caffe2/python/helpers/conv.py
```diff
@@ -54,20 +54,23 @@ def _ConvBase(
...
+    )
+    bias_initializer = initializers.update_initializer(
+        BiasInitializer, bias_init, ("ConstantFill", {})
+    )
+
     if model.init_params:
-        weight = model.param_init_net.__getattr__(weight_init[0])(
-            [],
-            blob_out + '_w',
+        weight = model.create_param(
```
akyrola, line 65: This unfortunately breaks things, as now weight & bias are added twice to params.

Test python/data_parallel_model_test.py -- test_checkpoint_params will fail.

I am uncommiting this for now.

akyrola: Sorry, this was mistake by me -- the offending diff came later. Trying to reimport.

Issue #706, opened 2017-05-31
Missing implementation for "MPIConstantFill"


lukeyeager:
```
$ test/mpi_test 
...
C++ exception with description "[enforce fail at operator.cc:110] op. Cannot create operator of type 'MPIConstantFill' on the device 'CPU'. Verify that implementation for the corresponding device exist. It might also happen if the binary is not linked with the operator implementation code. If Python frontend is used it might happen if dyndep.InitOpsLibrary call is missing. Operator def: output: "X" type: "MPIConstantFill" arg { name: "shape" ints: 10 } arg { name: "value" f: 0 } " thrown in the test body.
...
[==========] 5 tests from 1 test case ran. (4 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 5 tests, listed below:
[  FAILED  ] MPITest.TestMPIBroadcast
[  FAILED  ] MPITest.TestMPIReduce
[  FAILED  ] MPITest.TestMPIAllgather
[  FAILED  ] MPITest.TestMPIAllreduce
[  FAILED  ] MPITest.TestInPlaceMPIAllreduce

$ test/mpi_gpu_test
...
C++ exception with description "[enforce fail at operator.cc:110] op. Cannot create operator of type 'MPIConstantFill' on the device 'CUDA'. Verify that implementation for the corresponding device exist. It might also happen if the binary is not linked with the operator implementation code. If Python frontend is used it might happen if dyndep.InitOpsLibrary call is missing. Operator def: output: "X" type: "MPIConstantFill" arg { name: "shape" ints: 10 } arg { name: "value" f: 0 } device_option { device_type: 1 } " thrown in the test body.
...
[==========] 5 tests from 1 test case ran. (214 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 5 tests, listed below:
[  FAILED  ] MPITest.TestMPIBroadcast
[  FAILED  ] MPITest.TestMPIReduce
[  FAILED  ] MPITest.TestMPIAllgather
[  FAILED  ] MPITest.TestMPIAllreduce
[  FAILED  ] MPITest.TestInPlaceMPIAllreduce
```

Yangqing:
That is my fault. Will fix.

PR #709, opened 2017-05-31
MPIConstantFill -> ConstantFill

Yangqing:
(this is due to an earlier blind vim find-replace error)


Yangqing:
Fixes #706

lukeyeager:
I think you want the same change in `caffe2/mpi/mpi_test.cc`, too.

PR #719, opened 2017-06-01
Fix a few typos and grammars in comment

haracejacob:
Fix a few typos and grammars in comment

by using language-check, python library
spell_checker source code is here : https://github.com/17-1-SKKU-OSS/011A/blob/master/spell_checker/spell_checker.py 
here is the text file which indicates what things should be fixed :  https://github.com/17-1-SKKU-OSS/011A/tree/master/spell_checker/fix/caffe2


aaronmarkham (commented):
Thanks. These all look good.

PR #551, opened 2017-05-11
Fix KeepOnShrink tests

lukeyeager:
Fix https://github.com/caffe2/caffe2/issues/417

PR #723, opened 2017-06-01
Another fix for KeepOnShrink tests

### LINKED ISSUE #417: Flaky KeepOnShrink test
We occasionally hit failures like this:

```
[ RUN      ] TensorCPUTest/1.KeepOnShrink
/opt/caffe2/caffe2/core/blob_test.cc:364: Failure
      Expected: larger_ptr
      Which is: 0x224c400
To be equal to: smaller_ptr
      Which is: 0x2267ee0
/opt/caffe2/caffe2/core/blob_test.cc:371: Failure
      Expected: larger_ptr
      Which is: 0x224c400
To be equal to: new_ptr
      Which is: 0x224c460
[  FAILED  ] TensorCPUTest/1.KeepOnShrink, where TypeParam = int (0 ms)
```

lukeyeager:
*Fix #417 again (#551 was insufficient)*

Even after a reallocation, the data address can still be the same if malloc returns the same newly freed address.

* Be very explicit and careful about how we set these flags so they don't interfere with other tests
* Disable the failing check

This somewhat takes the teeth out of this test, since it no longer verifies that the reallocation actually occurs.

Test with:
```
blob_test --gtest_filter=TensorCPUTest*Shrink* \
    --gtest_shuffle --gtest_repeat=100 --gtest_throw_on_failure
```
/cc @sunwael

### DIFF
--- a/caffe2/core/blob_test.cc
+++ b/caffe2/core/blob_test.cc
@@ -347,8 +347,10 @@ TYPED_TEST(TensorCPUTest, NoLongerSharesAfterResize) {
 }
 
 TYPED_TEST(TensorCPUTest, KeepOnShrink) {
+  // Set flags (defaults)
   FLAGS_caffe2_keep_on_shrink = true;
   FLAGS_caffe2_max_keep_on_shrink_memory = LLONG_MAX;
+
   vector<int> dims{2, 3, 5};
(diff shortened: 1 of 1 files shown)


akyrola:
Ah, quite a tricky issue. Let's just land this. THanks!

PR #725, opened 2017-06-02
fix TensorProtosDBInput  function for parallel use

madongyu:
I think somebody forget  pass the **kwargs  to self.param_init_net.CreateDB function
it will lead to some name scope error when using multi gpus for parallel training


orionr:

PR #735, opened 2017-06-05
TravisCI Overhaul

lukeyeager:
Uncached build: https://travis-ci.org/lukeyeager/caffe2/builds/239677224
Cached build: https://travis-ci.org/lukeyeager/caffe2/builds/239686725

### Improvements:
* Parallel builds everywhere
* All builds use CCache for quick build times (help from https://github.com/pytorch/pytorch/pull/614, https://github.com/ccache/ccache/pull/145)
* Run ctests when available (continuation of https://github.com/caffe2/caffe2/pull/550)
* Upgraded from cuDNN v5 to v6
* Fixed MKL build (by updating pkg version)
* Fixed android builds (https://github.com/caffe2/caffe2/commit/b6f905a67b8cdc301203c08d5a598bb1ed6d1873#commitcomment-22404119)

### Things that are broken:
* ~~Building NNPACK fails with no discernible error message (currently disabled entirely)~~
* ~~Android builds continue to fail with existing error:~~
* ~~OSX builds time-out:~~

### Summary

| Before | After | Changes |
| --- | --- | --- |
| COMPILER=g++ | linux | without CUDA |
| COMPILER=g++-5 | linux-gcc5 | without CUDA |
| COMPILER=g++ | linux-cuda | updated to cuDNN v6 |
| BLAS=MKL | linux-mkl | updated pkg version |
| BUILD_TARGET=android | linux-android | |
| COMPILER=clang++ | osx | |
| BUILD_TARGET=ios | osx-ios | |
| BUILD_TARGET=android | osx-android | |
| QUICKTEST | **GONE** | |
| COMPILER=g++-4.8 | **GONE** | |
| COMPILER=g++-4.9 | **GONE** | |

PR #757, opened 2017-06-07
Fix observer_test

lukeyeager:
Fixes bug in tests introduced with https://github.com/caffe2/caffe2/commit/52dafaa7db9e974b25df665eb18d2dbb80b85edc.
```
[ RUN      ] ObserverTest.TestNotifyAfterDetach
/data/caffe2/caffe2/core/observer_test.cc:130: Failure
      Expected: 0
To be equal to: counter.load()
      Which is: 1212
```
Bug disovered with TravisCI builds at https://github.com/caffe2/caffe2/pull/735, but can also be seen in the existing builds at https://travis-ci.org/caffe2/caffe2/jobs/239913458.


lukeyeager:
Fixed by https://github.com/caffe2/caffe2/commit/13c99428c76f8f10985aef78a24e8692d496199b, but that solution depends on tests being executed in a particular order.

Issue #738, opened 2017-06-05
Installation issue: Unable to find Eigen


akshay-raj-dhamija:
Hello,

I get the following error while trying to make from the commit acee64a9f95612e69d0a615c09e0676ac076849d. I get the same error in various earlier commits whereas the commit 0a3c8413fc26cffa255d34ba7f4830e6c9724854 is working fine.
Eigen is already installed.
```
Building CXX object caffe2/CMakeFiles/Caffe2_CPU.dir/contrib/gloo/allreduce_ops.cc.o
In file included from /scratch/adhamija/caffe2/caffe2/core/context.h:12:0,
                 from /scratch/adhamija/caffe2/caffe2/core/tensor.h:14,
                 from /scratch/adhamija/caffe2/caffe2/core/net.h:18,
                 from /scratch/adhamija/caffe2/caffe2/core/operator.h:13,
                 from /scratch/adhamija/caffe2/caffe2/contrib/gloo/allreduce_ops.h:6,
                 from /scratch/adhamija/caffe2/caffe2/contrib/gloo/allreduce_ops.cc:1:
/scratch/adhamija/caffe2/caffe2/utils/math.h:20:22: fatal error: Eigen/Core: No such file or directory
 #include "Eigen/Core"
                      ^
compilation terminated.
```
Below is the complete make output
```
-- The BLAS backend of choice:Eigen
-- Could NOT find NNPACK (missing:  NNPACK_INCLUDE_DIR NNPACK_LIBRARY) 
-- Will try to build NNPACK from source. If anything fails, follow the NNPACK prerequisite installation steps.
CMake Warning at cmake/External/nnpack.cmake:157 (message):
  NNPACK is chosen to be installed, but confu and ninja that are needed by it
  are not installed.  As a result we won't build with NNPACK.
Call Stack (most recent call first):
  cmake/Dependencies.cmake:52 (include)
  CMakeLists.txt:72 (include)


CMake Warning at cmake/Dependencies.cmake:57 (message):
  Not compiling with NNPACK.  Suppress this warning with -DUSE_NNPACK=OFF
Call Stack (most recent call first):
  CMakeLists.txt:72 (include)


-- Found gflags  (include: /usr/include, library: /usr/lib/x86_64-linux-gnu/libgflags.so)
-- Found system gflags install.
-- Found glog    (include: /usr/local/include, library: /usr/local/lib/libglog.so)
-- Found system glog install.

[...]

PR #770, opened 2017-06-09
Fix for #738

akshay-raj-dhamija:
This fixes the variable name `EIGEN3_INCLUDE_DIRS` to `EIGEN3_INCLUDE_DIR` in the cmake dependencies as mentioned in issue https://github.com/caffe2/caffe2/issues/738
Ref: https://cmake-basis.github.io/apidoc/latest/FindEigen3_8cmake_source.html


akshay-raj-dhamija:
@hgaiser @lukeyeager Could you please review this small change for a merger.
Thanks!

hgaiser:
This error seems weird to me as the cmake that is provided by Eigen should set `EIGEN3_INCLUDE_DIRS` (https://bitbucket.org/eigen/eigen/src/8d1ccfd9c5a0bf1a662bd0fe35858f4e527d2bac/cmake/Eigen3Config.cmake.in?at=default&fileviewer=file-view-default). I couldn't find a cmake module in the cmake GitHub repository. What is that cmake-basis you are referencing? If it was using the Eigen's version of it's cmake module it would work fine. Can you try to find out why it is not using that cmake module? Is it even available?

akshay-raj-dhamija:
Thanks for your prompt response.
It appears that though the [current cmake](https://bitbucket.org/eigen/eigen/src/8d1ccfd9c5a0bf1a662bd0fe35858f4e527d2bac/cmake/Eigen3Config.cmake.in?at=default&fileviewer=file-view-default#Eigen3Config.cmake.in-15
) from eigen sets `EIGEN3_INCLUDE_DIRS`, it isn't consistent with a [Previous Version](https://bitbucket.org/eigen/eigen/src/1db6cf71f62897825a643f5faca8a54adc90b41d/cmake/Eigen3Config.cmake.in?fileviewer=file-view-default#Eigen3Config.cmake.in-22) of eigen.
But, it also appears that `EIGEN3_INCLUDE_DIR` hasn't changed between these versions. To support previous versions of eigen it might be better to use `EIGEN3_INCLUDE_DIR`.

hgaiser:
The only difference between the two cmake files you linked is the name of the internal variable they use to substitute when compiling the cmake file (is it called compiling? Or template substitution?). The variable `EIGEN3_INCLUDE_DIRS` appears to exist since the `EIGEN3_*` variables were added (https://bitbucket.org/eigen/eigen/diff/cmake/Eigen3Config.cmake.in?diff2=21fbeccc596f&at=vladimir_ch/bug-796-fix-eigen3config.cmake). I still believe the real issue is that your system is not finding the cmake provided by Eigen.

pietern:
Thanks @hgaiser for investigating. Also checked /usr/lib/cmake/eigen3 (on Ubuntu) and confirmed it sets both, so we can stick to EIGEN_INCLUDE_DIRS.

gemarcano:
[Upstream's FindEigen3.cmake](https://bitbucket.org/eigen/eigen/src/2784ec4bdf50a92acfd0dd2396a81ffc3e1b993a/cmake/FindEigen3.cmake?at=default&fileviewer=file-view-default) only defines `EIGEN_INCLUDE_DIR`... makes me wonder what version the Ubuntu repositories are using, but it isn't latest (unless Ubuntu is patching something, this means in the future a future release is going to break this).

EDIT: I see that the file you're looking at is Eigen3Config.cmake, which does set that secondary variable, even upstream. This means I'm running into another problem. Disregard this comment.

pietern:
This was also the case on CentOS. I updated this to use EIGEN3_INCLUDE_DIR (singular) in #1505.

PR #788, opened 2017-06-12
Fix stats_ops_test

lukeyeager:
The global StatRegistry doesn't get reset when the workspace is reset.
```
>       self.assertTrue(len(workspace.FetchBlob('k3')) == 2)
E       AssertionError: False is not true
```
https://travis-ci.org/lukeyeager/caffe2/jobs/240162665

/cc @azzolini 

NOTE: this error doesn't show up if you just run `stats_ops_test.py` directly. It shows up when you run other tests in the same session before this test:
```
pytest -v caffe2/python/
```

### DIFF
--- a/caffe2/python/operator_test/stats_ops_test.py
+++ b/caffe2/python/operator_test/stats_ops_test.py
@@ -18,7 +18,8 @@ class TestCounterOps(TestCase):
         previous_keys = workspace.FetchBlob('prev_k')
         existing = len(previous_keys)
 
-        keys = [__name__ + '/key1', __name__ + '/key2']
+        prefix = '/'.join([__name__, 'TestCounterOps', 'test_stats_ops'])
+        keys = [prefix + '/key1', prefix + '/key2']
         values = [34, 45]
         workspace.FeedBlob('k', np.array(keys, dtype=str))
         workspace.FeedBlob('v', np.array(values, dtype=np.int64))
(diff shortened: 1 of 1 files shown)


lukeyeager:
In particular, try running these two tests:
```
pytest caffe2/python/checkpoint_test.py caffe2/python/operator_test/stats_ops_test.py
```

caffe2/python/operator_test/stats_ops_test.py
```diff
@@ -11,20 +11,37 @@
...
-        workspace.FeedBlob('k', np.array(['k0', 'k1'], dtype=str))
-        workspace.FeedBlob('v', np.array([34, 45], dtype=np.int64))
+        # The global StatRegistry isn't reset when the workspace is reset,
+        #   so there may be existing stats from a previous test
+        workspace.RunOperatorOnce(core.CreateOperator(
+            'StatRegistryExport', [], ['prev_k', 'prev_v', 'prev_ts']))
+        previous_keys = workspace.FetchBlob('prev_k')
+        existing = len(previous_keys)
+
+        keys = [__name__ + '/key1', __name__ + '/key2']
```
salexspb: what if it is from the same file? Shouldn't this include the test name? You could also use line number.

salexspb (changes requested):

caffe2/python/operator_test/stats_ops_test.py
lukeyeager: Ok, updated.

PR #790, opened 2017-06-13
Initialize test_model before training

davinwang:
If the test model is created after the training, the test model will be initialized and the accuracy is less than 0.10. By creating the test model before training, the test model will get the trained parameter and accuracy is around 0.95.


caffe2/python/tutorials/MNIST.ipynb
```diff
@@ -553,6 +553,10 @@
     "workspace.RunNetOnce(train_model.param_init_net)\n",
     "# creating the network\n",
     "workspace.CreateNet(train_model.net, overwrite=True)\n",
+    "# Also initialize the test model before training\n",
+    "workspace.RunNetOnce(test_model.param_init_net)\n",
```
salexspb, line 557: does it actually have to be called? could you check if we can remove this initialization call all together?

salexspb: Also init_param=False on the model should take care of not overriding them. Could you check what is in param_init_net to see if it overrides something? 

See conv.py and fc.py, there we don't initialize parameters if model.init_params == False

salexspb (changes requested):

caffe2/python/tutorials/MNIST.ipynb
davinwang: It turns out error if I remove either of the 2 lines.
If this is not the expected usage, then please help fix the core and brew module......

salexspb: could you post here how test_model.param_init_net.Proto() looks like for you?

davinwang: Sure, here is the file generated by
`with open(os.path.join(root_folder, "test_init_net.pbtxt"), 'w') as fid:
    fid.write(str(test_model.param_init_net.Proto()))`
before `RunNetOnce` and `CreateNet`
[test_init_net.pb.txt](https://github.com/caffe2/caffe2/files/1079640/test_init_net.pb.txt)

LimingYin: Here is my solution:
`test_model = model_helper.ModelHelper(name="mnist_test", arg_scope=arg_scope, init_params=False)
data, label = AddInput(test_model, batch_size=100,db=os.path.join(data_folder, 'mnist-test-nchw-leveldb'),db_type='leveldb')

workspace.RunNetOnce(test_model.param_init_net)

softmax = AddLeNetModel(test_model, data)
AddAccuracy(test_model, softmax, label)

workspace.CreateNet(test_model.net)
workspace.RunNet(test_model.net.Proto().name)`


First, just define the test_model's input part, and immediately initalize it.
Then add model to it.
At last, create and run.

davinwang: @salexspb Thank you. I kinda understand why you ask for the test_model.param_init_net.Proto(). It actually contains 5 ops: `CreateDB`, `conv1_w`, `conv1_b`, `conv2_w`, `conv2_b`. You can find the attachment in above discussion. The expected result should be only `CreateDB`. So I started #824 to fix the root cause.

PR #797, opened 2017-06-14
Fix multi-precision SGD outputs

slayton58:
@salexspb This fixes a major perf issue (40% boost on alexnet end-to-end perf) in the multi-precision SGD optimizer - it was causing repeated cudaMalloc / cudaFree calls during training iterations due to the changing size of the `grad` blob as it moved from fp16 <-> fp32.


salexspb:
Oh man, you warned about implicit stuff being dangerous. Still we have hit it :( Is it possible to guard against such things?

slayton58:
Maybe with the type-inference stuff C2 has, as we'd expect `type(input_grad) == type(output_grad)`

salexspb:
should we just make that op enforce that?

PR #802, opened 2017-06-14
Fix potential GPU transform OOB access

slayton58:
Occurred when running with multiple GPUs, not all of which
are connected via P2P.

Essentially when mean_gpu_ and std_gpu_ are allocated and
populated in the constructor of ImageInputOp, it does not seem to
be guaranteed that the active context is the same as the final context
on which the Op will be run. This causes the image data and the
mean/std to be on different devices. With P2P we don't mind this, but
without this causes OOB memory accesses in the GPU transform
kernel.

### DIFF
--- a/caffe2/image/image_input_op.h
+++ b/caffe2/image/image_input_op.h
@@ -84,6 +84,7 @@ class ImageInputOp final
   bool is_test_;
   bool use_caffe_datum_;
   bool gpu_transform_;
+  bool mean_std_copied_ = false;
 
   // thread pool for parse + decode
   int num_decode_threads_;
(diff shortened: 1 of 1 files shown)


slayton58:
@Yangqing for review

akyrola:
Indeed, I don't think the GPU id is set when calling oeprator constructors.

akyrola:
Hmm, actually operator.h will set the device (context_.SwitchToDevice) at constructor. Let me say why ImageInputOp doesn't do it

akyrola:
it is because ImageINputOp does not inherit Operator, but OperatorBase via PrefetchOperator, and neither of those will call SwitchDevice()

slayton58:
Ahhh, good catch! Which is easier, changing ImageInputOp to inherit from Operator, or my fix?

akyrola:
Well I landed your fix already, I think it is fine. Actually I dislike doing data copy / computation in constructor anyway. But I am submitting diff to also fix the PrefetchOperator, as it can be used by other ops as well.

PR #804, opened 2017-06-14
Fix for protobuf with unicode_literals

lukeyeager:
Python 2.7, Protobuf 2.6

    >                   op.ClearField('uuid')
    E                   TypeError: field name must be a string

Fix: http://python-future.org/imports.html#should-i-import-unicode-literals

/cc @salexspb @tomdz

### DIFF
--- a/caffe2/python/core_gradients_test.py
+++ b/caffe2/python/core_gradients_test.py
@@ -3,6 +3,7 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+from future.utils import bytes_to_native_str
 from hypothesis import given
 import hypothesis.strategies as st
 import unittest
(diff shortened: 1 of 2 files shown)


lukeyeager:
Once https://github.com/caffe2/caffe2/pull/735 is merged, I'll work on adding targets which test both Python2 tests and Python3 tests.

salexspb:
IMO this is ugly to call that function for all the string literals. Are you sure this is the only one way to go?

lukeyeager:
I'm not sure, no. I just followed the instructions from the `python-future` guys. I was hoping you might know more about it than me. I haven't even tested this with Python3 or Protobuf3 - I'm not sure this solution works for all those configs.

salexspb:
I would actually ask @tomdz

PR #820, opened 2017-06-18
allow optional usage of use_caffe_datum in resnet50 trainer

neobarney:
set use_caffe_datum arg in AddImageInput to False as default value, and added use_caffe_datum as new optional argument in argparse


neobarney:
Ok I signed up to the CLA

caffe2/python/examples/resnet50_trainer.py
```diff
@@ -446,6 +448,8 @@ def main():
                         help="Save the trained model to a given name")
     parser.add_argument("--load_model_path", type=str, default=None,
                         help="Load previously saved model to continue training")
+    parser.add_argument("--use_caffe_datum", type=bool, default=False,
```
akyrola, line 455: default should be True?

neobarney: Default in image_input_op.cc is 0 (https://github.com/caffe2/caffe2/blob/master/caffe2/image/image_input_op.cc#L69)

So I think default in the trainer should be set to False too

orionr:

PR #833, opened 2017-06-20
Split cuDNN convolution gradient operators

slayton58:
Splits the monolithic `CuDNNConvGradientOp` into their components. `{CuDNNConvDataGradientOp, CuDNNConvFilterGradientOp, CuDNNConvBiasGradientOp}`.

This has a few advantages:
1. Potentially allows for better scaling via more fine-grained dependencies for communication (communication of bias / filter gradients no longer depend on the computation of all of `{dX, dW, db}`, just the relevant parameter gradient
2. Potentially overlap of wgrad, dgrad and bgrad calls
3. More fine-grained reporting of timings through convnet_benchmarks

\cc @asaadaldien as I know he's been working on something similar for other convolution ops.


lukeyeager:
TravisCI timed out on an unzip command? These osx build machines are killing me.

caffe2/operators/conv_dgrad_op_cudnn.cc
```diff
@@ -0,0 +1,291 @@
...
+        &ws_size));
+    cudnn_ws_nbytes_ = ws_size;
+
+    VLOG(1) << "CuDNN bwd data algorithm: " << algo_;
+    VLOG(1) << "CuDNN workspace size: " << cudnn_ws_nbytes_;
+  }
+
+  std::stringstream cudnnParamStr;
+  if (VLOG_IS_ON(2)) {
+    cudnnParamStr
```
asaadaldien: I think that need to be member function 
`string CudnnConvOpBase::GetcuDNNParamStr(DimensionParam)`
So you can reuse it below in the forward pass.

asaadaldien (commented):
LGTM! Just one single comment.

caffe2/operators/conv_dgrad_op_cudnn.cc
slayton58: This was debugging code that I meant to get rid of before push - removed

caffe2/operators/conv_op_cudnn_base.h
```diff
@@ -0,0 +1,278 @@
...
+  };
+
+  DimensionParam GetDimensionsFromTensorDims(
+      const int ndim,
+      const vector<TIndex>& dims,
+      StorageOrder order) {
+
+    DimensionParam param;
+
+    switch (order_) {
```
asaadaldien: A typo this should be:
`order`

asaadaldien (changes requested):

caffe2/operators/conv_op_cudnn_base.h
slayton58: Fixed.

caffe2/operators/conv_op_cudnn_base.h
```diff
@@ -0,0 +1,278 @@
...
+        param.D = ndim > 4 ? dims.at(3) : 1;
+        param.C = dims.at(ndim - 1);
+        break;
+      case StorageOrder::NCHW:
+        param.N = dims.at(0);
+        param.C = dims.at(1);
+        param.H = dims.at(2);
+        param.W = ndim > 3 ? dims.at(3) : 1;
+        param.D = ndim > 4 ? dims.at(4) : 1;
+    }
```
asaadaldien: Add a default case:
```
+    default:
 +      LOG(FATAL) << "Unknown storage order: " << order_;
```

asaadaldien (changes requested):

caffe2/operators/conv_op_cudnn_base.h
slayton58: Fixed (with `order` instead of `order_` ;) )

caffe2/operators/conv_dgrad_op_cudnn.cc
```diff
@@ -0,0 +1,270 @@
...
+
+template <typename T_X, typename T_W, typename T_DY,
+          typename MATH,
+          typename T_DX>
+bool CudnnConvDataGradientOp::DoRunWithType() {
+  auto& X = Input(INPUT);
+  auto& filter = Input(FILTER);
+  auto& dY = Input(OUTPUT_GRAD);
+  auto* dX = Output(INPUT_GRAD);
+
```
asaadaldien, line 36: Can we resize dX here once get a pointer access then use it everywhere else ?

```
dX->ResizeLike(X);
T_DX* dXdata = dX->template mutable_data<T_DX>();
```

asaadaldien (changes requested):

caffe2/operators/conv_dgrad_op_cudnn.cc
slayton58: I don't really understand the question. I can certainly get all the data pointers here and use them throughout if that's what you're asking, so add:

```
dX->ResizeLike(X);
filter_data = filter.template data<T_W>();
dY_data = dY.template data<T_DY>();
dX_data = dX->template data<T_DX>();
```

and use those values through the rest of the method

asaadaldien: Yes, lets do that please.
The reason I asked for this is to remove the redefinition of dX in lines 160 and 215 because it doesn't compile with -Werror=shadow-compatible-local and to remove resizing dX for every group in 216.

slayton58: ok, on it :)

slayton58: Ok, I have that fix locally.

One thing I've noticed is that when running the gradient checker, the host-side checker is trying to run the GPU-side net, which fails because the splitting of the ops hasn't happened anywhere apart from cuDNN. Is there a way around this?

asaadaldien: looks like #839 is going to fix the test issue ?

slayton58: This fixes _some_ of the issues, but a fix for https://github.com/caffe2/caffe2/pull/861 needs to be found before the rest can be fixed.

slayton58: Managed to work around #861 - simply enforcing `use_cudnn=False` in the single test that uses `ModelHelper` worked around it.

slayton58:
@asaadaldien Apologies for the delay on updating this one - all tests are passing locally now

caffe2/operators/conv_dgrad_op_cudnn.cc
```diff
@@ -0,0 +1,272 @@
...
+    cudnn_ws_nbytes_ = ws_size;
+
+    VLOG(1) << "CuDNN bwd data algorithm: " << algo_;
+    VLOG(1) << "CuDNN workspace size: " << cudnn_ws_nbytes_;
+  }
+
+  for (int i = 0; i < group_; ++i) {
+    cudnn_wrapper_.with_cudnn_state(cudnn_state_, [&](CuDNNState* state) {
+      // Compute the gradient w.r.t. the input.
+      auto* dX = Output(INPUT_GRAD);
```
asaadaldien: Please remove the redefinition of dX and the Resize (lines 217, 218).

caffe2/operators/conv_gradient_op.cc
```diff
@@ -16,35 +16,71 @@ class GetConvGradient : public GradientMakerBase {
 
     auto compute_dX = !argsHelper.GetSingleArgument<bool>("no_gradient_to_input", 0);
 
-    if (def_.input_size() == 3) {
+    auto engine = def_.engine();
+
+    // Use seperate gradient ops
+    if (engine == "CUDNN") {
```
asaadaldien: There is still some conv tests failing because of running host with engine='CUDA'.
We can fix this by checking that device is CUDA:: 
`if (engine == "CUDNN" && def._ device_option().device_type == 1)`

slayton58: I actually already have that locally, thought it'd got added. I'll update in the morning

caffe2/python/operator_test/conv_test.py
```diff
@@ -289,9 +346,10 @@ def test_3d_convolution_nchw(self, input_channels, output_channels,
            dilation=st.integers(1, 2),
            pad=st.integers(0, 2),
            use_bias=st.booleans(),
-           **hu.gcs)
-    def test_3d_convolution_cudnn_nchw(self, batch_size, stride, size, kernel,
-                                       dilation, pad, use_bias, gc, dc):
+           **hu.gcs_gpu_only)
```
asaadaldien, line 350: I think now this test should work for both GPU & CPU can you change:
`**hu.gcs_gpu_only` -> `**hu.gcs`

caffe2/python/operator_test/conv_test.py
```diff
@@ -315,17 +373,59 @@ def test_3d_convolution_cudnn_nchw(self, batch_size, stride, size, kernel,
...
         self.assertDeviceChecks(dc, op, inputs, [0])
+
+    @given(batch_size=st.integers(1, 2),
+           stride=st.integers(1, 2),
+           size=st.integers(3, 5),
+           kernel=st.integers(1, 2),
+           dilation=st.integers(1, 2),
+           pad=st.integers(0, 2),
+           use_bias=st.booleans(),
+           **hu.gcs_gpu_only)
```
asaadaldien, line 397: `**hu.gcs_gpu_only`-> `**hu.gcs`

caffe2/python/operator_test/conv_test.py
slayton58: These are explicitly cudnn tests -- there's no point in running non-GPU.

caffe2/python/operator_test/conv_test.py
slayton58: See above.

asaadaldien: Ok! can you please annotate the test to run only when GPU device is available ?
`@unittest.skipIf(not workspace.has_gpu_support, "No gpu support.")`
Otherwise it will crash on CPU only machines.

slayton58: Sorry for the delay, added the `skipIf` calls and rebased from latest master

orionr:

PR #813, opened 2017-06-16
changed download links to download.caffe2.ai

aaronmarkham:

PR #836, opened 2017-06-20
tutorial resources and binaries link updates

### LINKED ISSUE #349: make_mnist_db doesn't generate db files

aaronmarkham:
changed aws links to download.caffe2.ai; changed binary links to bin; removed executable permission on MNIST tutorial.
Fixes #349 
Replaces #813


JoelMarcey (approved):
lgtm

PR #843, opened 2017-06-21
Fixes for test_recurrent in hypothesis_test.py

lukeyeager:
/cc @akyrola is it possible this test has been broken ever since https://github.com/caffe2/caffe2/commit/5614816fce8e0d01f6f34fd9272b56a5c658f649?

More generally, why do we still have `hypothesis_test.py` at all? In the case of this test, surely one of these files does more than this one old test:

* `operator_test/cudnn_recurrent_test.py`
* `operator_test/recurrent_network_test.py`
* `operator_test/rnn_cell_test.py`


akyrola:
Yes, this has been broken unfortunately, and I have forgotten to fix it. It could be removed, frankly, as well.

akyrola:
Or actually not, since this test  is for the CUDNN version. Thanks for fixing.

PR #817, opened 2017-06-16
TravisCI: run Python tests

lukeyeager:
Run Python tests (with pytest) on TravisCI.

PR #858, opened 2017-06-23
TravisCI: fix OSX builds

lukeyeager:
Apparently, `brew install` fails if the package is already installed?
```
Error: automake 1.15 is already installed
```
https://travis-ci.org/caffe2/caffe2/jobs/245226634

Maybe TravisCI made some unannounced updates to their OSX images at around the same time [they updated their trusty images](https://blog.travis-ci.com/2017-06-21-trusty-updates-2017-Q2-launch). Something changed on their side two days ago, and the OSX builds have been failing ever since.

### DIFF
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -137,7 +137,6 @@ elif [ "$TRAVIS_OS_NAME" = 'osx' ]; then
     pip uninstall -y numpy  # use brew version (opencv dependency)
     brew tap homebrew/science  # for OpenCV
     brew install \
-        automake \
         ccache \
         glog \
         leveldb \
@@ -169,7 +168,7 @@ if [ "$BUILD_ANDROID" = 'true' ]; then
         $APT_INSTALL_CMD autotools-dev autoconf
         wget -O "$_ndk_zip" https://dl.google.com/android/repository/android-ndk-r13b-linux-x86_64.zip
     elif [ "$TRAVIS_OS_NAME" = 'osx' ]; then
-        brew install automake libtool
+        brew install libtool
         wget -O "$_ndk_zip" https://dl.google.com/android/repository/android-ndk-r13b-darwin-x86_64.zip
     else
         echo "OS \"$TRAVIS_OS_NAME\" is unknown"


lukeyeager:
This will get the OSX builds up-and-running again, but there will still be a 20-30 minute delay starting those targets for the forseeable future.

> For travis-ci.org we ... have had to make some hard decisions on how and when to increase this capacity as the cost for helping the open source community is fairly high.
https://github.com/travis-ci/travis-ci/issues/7304#issuecomment-287591291

aaronmarkham:
So the Travis log indicates that brew is installing protobuf 3.3... I thought there were issues with this version versus what's deployed within third party... if not, that's great! I'd want to update some installation instructions around this because protobuf version mismatches have been a problem for Caffe2 on OSX... 
Also, the removal of automake... it is specific to Travis, correct? 
How about the addition of libtool? Is that something we should have as a dependency in addition to automake for regular OSX build instructions?

lukeyeager:
@aaronmarkham I don't actually know much about the OSX builds - I just copied whatever was there from @bwasti's old code.

> So the Travis log indicates that brew is installing protobuf 3.3... I thought there were issues with this version versus what's deployed within third party... if not, that's great! I'd want to update some installation instructions around this because protobuf version mismatches have been a problem for Caffe2 on OSX...

I haven't checked. It's possible the build is actually pulling in code from third_party and this brew installation is ignored.

> Also, the removal of automake... it is specific to Travis, correct?

Yeah, as I mentioned in the OP it seems that Travis updated their base image or something. Automake is already installed, and brew throws an error when we try to install it again. Another possibility is that Travis has a newer version of brew now which handles re-installation differently. I didn't look into it very deeply.

> How about the addition of libtool? Is that something we should have as a dependency in addition to automake for regular OSX build instructions?

I'd talk to @bwasti - I'm not sure why it was there in the first place.

lukeyeager:
But I don't think updating the OSX build instructions should block this getting merged. Builds have been failing for several days because of this issue:
https://travis-ci.org/caffe2/caffe2/builds

Plus it's kind of holding me back on this PR: https://github.com/caffe2/caffe2/pull/817

aaronmarkham:
> But I don't think updating the OSX build instructions should block this getting merged. 

I don't think so either! Just wanted to know if there were changes to make.

Issue #860, opened 2017-06-23
Setting engine in brew.conv doesn't override use_cudnn


lukeyeager:
This non-intuitive behavior is giving @slayton58 trouble while working on https://github.com/caffe2/caffe2/pull/833.
```python
>>> from caffe2.python import brew, model_helper
>>> model = model_helper.ModelHelper()
>>> brew.conv(model, 'conv_in', 'conv_out', 10, 10, 5, engine='')
BlobReference("conv_out")
>>> print('Proto: "%s"' % model.Proto())
Proto: "name: "model"
op {
  input: "conv_in"
  input: "conv_out_w"
  input: "conv_out_b"
  output: "conv_out"
  name: ""
  type: "Conv"
  arg {
    name: "kernel"
    i: 5
  }
  arg {
    name: "exhaustive_search"
    i: 0
  }
  arg {
    name: "order"
    s: "NCHW"
  }
  engine: "CUDNN"
  uuid: 14532784363511038896
}
external_input: "conv_in"
external_input: "conv_out_w"
external_input: "conv_out_b"
"
```

slayton58:
\cc @Yangqing 

This is tricky, because defaulting to `use_cudnn=True` in `ModelHelper` is useful for normal runs, but a mess during testing (which as Luke mentioned is tripping me up).

Another way of working around this would be to explicitly setup ModelHelper without cuDNN for all testing runs (presumably abstracting out the creation of ModelHelper so that it can be modified centrally

PR #839, opened 2017-06-20
Refactor conv_test - no cuDNN+dilation+NHWC

lukeyeager:
Place all the cuDNN version checks in a helper function. Easier to use
in future tests and update for newer versions of cuDNN in one place.

Fixes this error in `test_convolution_gradients`:
```
RuntimeError: [enforce fail at conv_op_cudnn.cc:519] status == CUDNN_STATUS_SUCCESS. 9 vs 0. , Error at: /data/caffe2/caffe2/operators/conv_op_cudnn.cc:519: CUDNN_STATUS_NOT_SUPPORTED Error from operator: 
input: "X" input: "w" output: "Y" name: "" type: "Conv" arg { name: "stride" i: 1 } arg { name: "pad" i: 0 } arg { name: "order" s: "NHWC" } arg { name: "dilation" i: 2 } arg { name: "kernel" i: 1 } device_option { device_type: 1 } engine: "CUDNN"
```

PR #861, opened 2017-06-23
Conv frontend: checking engine and use_cudnn

lukeyeager:
*Fixes https://github.com/caffe2/caffe2/issues/860*

Raise an exception when the user specifies conflicting values for `engine` and `use_cudnn` in the conv frontend.

### DIFF
--- a/caffe2/python/helpers/conv.py
+++ b/caffe2/python/helpers/conv.py
@@ -38,6 +38,17 @@ def _ConvBase(
     else:
         kernels = [kernel] * 2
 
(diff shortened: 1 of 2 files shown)


slayton58 (approved):
This guards against some very counter-intuitive behaviour

lukeyeager:
To be clear, I expect the @caffe2 team may want to fix this differently (i.e. by letting one flag override the other more explicitly). I just filed a PR as an example fix for #860.

lukeyeager:
Rebased after merge conflict from #839.

PR #871, opened 2017-06-26
Tests: handle missing python-lmdb gracefully

lukeyeager:
Fix issue mentioned here: https://github.com/caffe2/caffe2/commit/875a9850c15f7bbc4466aec6054f25f327793bcd#commitcomment-22773221

Unblocks https://github.com/caffe2/caffe2/pull/817

/cc @tomdz

### DIFF
--- a/caffe2/python/operator_test/image_input_op_test.py
+++ b/caffe2/python/operator_test/image_input_op_test.py
@@ -6,12 +6,12 @@ from __future__ import unicode_literals
 import unittest
 try:
     import cv2
+    import lmdb
 except ImportError:
     pass  # Handled below
 
 from PIL import Image
 import numpy as np
-import lmdb
 import shutil
 try:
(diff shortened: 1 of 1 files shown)


lukeyeager:
@tomdz I'm assuming your existing test infrastructure can't gracefully handle tests which raise `unittest.SkipTest` exceptions? If you could share how you run Python tests internally, I can adjust the TravisCI scripts and our own internal CI scripts to match more closely so we have fewer conflicts like this.

lukeyeager:
Alternatively, we can make `lmdb` a first-class requirement and I can add it to the list of packages I install on Travis. I'm trying to keep the list as small as possible since I have nothing to go off of, but if you create a `requirements.txt` file, I'll happily use that instead.

akyrola:
Ugly, but fine.

PR #872, opened 2017-06-27
data_parallel_model: NCCLBroadcast root fix

lukeyeager:
The root is the root _rank_ and not the root _device_. Thus we always
use root=0, regardless of the devices used.

https://github.com/NVIDIA/nccl/blob/v1.3.0-1/src/broadcast.cu#L75

/cc @slayton58


akyrola:
Thanks

PR #878, opened 2017-06-27
Miscellaneous fixes for compatibility with MSVC

willyd:
I successfully built and used caffe2 with CUDA on Windows after applying the following modifications to the source code.


willyd:
Closed in favor of #1042

PR #886, opened 2017-06-28
Tests: fixing observer_test

lukeyeager:
Fixes bug reported at https://github.com/caffe2/caffe2/commit/3559ff93f97a99cd5052953a7d5c65476304f3f1#commitcomment-22813961. Builds have been broken since that commit was merged.

/cc @salexspb

### DIFF
--- a/caffe2/core/observer_test.cc
+++ b/caffe2/core/observer_test.cc
@@ -21,16 +21,13 @@ class DummyObserver final : public ObserverBase<T> {
   bool Stop() override;
 
   ~DummyObserver() {}
-
- private:
-  vector<unique_ptr<DummyObserver<OperatorBase>>> ops_obs;
 };
 
 template <>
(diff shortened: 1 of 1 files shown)


salexspb:
Oops,  I wonder why I  was able to land this at all. Thanks for the fix!

I  can't import it though: 

This pull request hasn't been processed yet, so can't be imported or landed; please try again in a few minutes.

lukeyeager:
^^ Does that mean your importer system waits for Travis to pass? If so, we should really think about turning off the OSX build targets. Because they're so slow - https://github.com/caffe2/caffe2/pull/858#issuecomment-310720922.

lukeyeager:
@salexspb can you import it now that Travis passed?

salexspb:
I actually included a fix  in an internal diff just in case this shipit thing takes more time (it was before travis passed).

PR #893, opened 2017-06-29
Optimizer: one LR op per (device, optimizer)

lukeyeager:
Try running this script through `nvprof`:
```py
import numpy as np
from caffe2.proto import caffe2_pb2
from caffe2.python import brew, core, optimizer, workspace
from caffe2.python.model_helper import ModelHelper

do = core.DeviceOption(caffe2_pb2.CUDA, 0)
with core.DeviceScope(do):
    model = ModelHelper(arg_scope={'order': 'NCHW'})
    conv1 = brew.conv(model, 'data', 'conv1', 1, 20, 5)
    pool1 = brew.max_pool(model, conv1, 'pool1', kernel=2, stride=2)
    conv2 = brew.conv(model, pool1, 'conv2', 20, 50, 5)
    pool2 = brew.max_pool(model, conv2, 'pool2', kernel=2, stride=2)
    fc3 = brew.fc(model, pool2, 'fc3', 50 * 4 * 4, 500)
    fc3 = brew.relu(model, fc3, fc3)
    pred = brew.fc(model, fc3, 'pred', 500, 10)
    softmax, loss = model.SoftmaxWithLoss([pred, 'label'], ['softmax', 'loss'])
    model.AddGradientOperators([loss])
    optimizer.build_sgd(model, 0.01,
                        policy='step', stepsize=1, gamma=0.999,
                        momentum=0.9, nesterov=False)
    workspace.FeedBlob('data', np.zeros((1, 1, 28, 28), dtype=np.float32))
    workspace.FeedBlob('label', np.zeros((1, 1), dtype=np.int32))

workspace.RunNetOnce(model.param_init_net)
workspace.CreateNet(model.net)

for _ in range(100):
    workspace.RunNet(model.net)
```
Before this change:
```
                    1.55%  1.4185ms       837  1.6940us  1.6630us  2.4000us  [CUDA memcpy HtoD]
                    0.72%  656.03us       200  3.2800us  3.1350us  3.5840us  [CUDA memcpy DtoD]
                    0.39%  7.1574ms      1034  6.9220us  3.8300us  18.677us  cudaMemcpyAsync
                    0.00%  34.180us         3  11.393us  9.0960us  12.910us  cudaMemcpy
```
And after it (look at the third column):
```
                    0.73%  657.15us       200  3.2850us  3.1040us  3.6160us  [CUDA memcpy DtoD]
                    0.26%  235.07us       137  1.7150us  1.6640us  2.3680us  [CUDA memcpy HtoD]
                    0.20%  3.4493ms       334  10.327us  6.4220us  16.958us  cudaMemcpyAsync
                    0.00%  37.376us         3  12.458us  9.4120us  15.412us  cudaMemcpy
```
That makes a pretty big difference in performance. Is there any particular reason you decided to have a separate `LearningRate` op for every parameter in https://github.com/caffe2/caffe2/commit/1317e3498cf6ac2e58ab93b5239ad09ca7f46bc6?

### DIFF
--- a/caffe2/python/optimizer.py
+++ b/caffe2/python/optimizer.py
@@ -44,10 +44,23 @@ class Optimizer(object):
     def _run(self, net, param_init_net, param_info):
         raise Exception("Not Impelemented")
 
-    @staticmethod
-    def build_lr(net, param_init_net, base_learning_rate,
-                 learning_rate_blob="lr", policy="fixed",
+    def get_lr_blob_name(self):
+        """Returns an LR blob name.
+        The name will be unique to the current device and optimizer type.
+        """
+        classname = self.__class__.__name__
+        s = scope.CurrentDeviceScope()
+        if s.device_type == caffe2_pb2.CUDA:
+            return '%s_lr_gpu%d' % (classname, s.cuda_gpu_id)
+        else:
+            return '%s_lr_cpu' % classname
+
+
+    def build_lr(self, net, param_init_net, base_learning_rate,
+                 learning_rate_blob=None, policy="fixed",
                  iter_val=0, **kwargs):
+        if learning_rate_blob is None:
+            learning_rate_blob = self.get_lr_blob_name()
         if not param_init_net.BlobIsDefined(_OPTIMIZER_ITERATION_NAME):
             # Add training operators.
             with core.DeviceScope(core.DeviceOption(caffe2_pb2.CPU)):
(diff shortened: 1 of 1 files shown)


lukeyeager:
> Is there any particular reason you decided to have a separate LearningRate op for every parameter in 1317e34?

Ping.

caffe2/python/optimizer.py
```diff
@@ -17,6 +17,15 @@
 
 AuxOptimizerParams = namedtuple("AuxOptimizerParams", ["local", "shared"])
 
+def _get_lr_blob_name():
```
asaadaldien: @lukeyeager There is some use cases where we have different optimizers with different learning rates. So we need to do this per device per optimizer class for example we can have:
`def _get_lr_blob_name(optimizer_name)` and add the optimizer_name as prefix to the lr blob name.

lukeyeager: That would work for your use-case, and produce correct code. But it also makes this change a no-op, and doesn't fix the performance problem. Would it be reasonable to add a flag like `unique_lr_per_param=False`?

asaadaldien: Why there will be perf problem? It should create only one learning rate blob per optimization class, so if we did : 
```
def _get_ler_blob_name(optimizer_prefix):
     s = scope.CurrentDeviceScope()
     if s.device_type == caffe2_pb2.CUDA:
         return "%s_lr_gpu_%d" % (optimizer_prefix, s.cuda_gpu_id)
``` 
and called `_get_ler_blob_name(self.__class__.name__)` for every optimizer class.
In your example there will be one lr blob per gpu SgdOptimizer_lr_gpu_0, SgdOptimizer_lr_gpu_0, ..etc

lukeyeager: Oh, I misread your original comment. That could work! Let me try it ...

lukeyeager:
Updated with @asaadaldien's suggestions - thanks!

asaadaldien:
Thanks!

PR #894, opened 2017-06-29
Add NCCLBroadcast to correct net

lukeyeager:
Otherwise was always added to main net instead of param_init_net when
desired (i.e. initial param sync)


akyrola:
Uh, this was pretty bad bug!

PR #911, opened 2017-07-08
Docker: add Dockerfile.cuda

lukeyeager:
In case anyone is interested, here's what I'm currently using for development on `master`. 

For comparison, see Dockerfiles [from](https://github.com/BVLC/caffe/blob/rc5/docker/gpu/Dockerfile) [other](https://github.com/pytorch/pytorch/blob/v0.1.12/Dockerfile) [frameworks](https://github.com/tensorflow/tensorflow/blob/v1.2.1/tensorflow/tools/docker/Dockerfile.gpu).

### DIFF
--- a/docker/Dockerfile.cuda
+++ b/docker/Dockerfile.cuda
@@ -1,0 +1,49 @@
+FROM nvidia/cuda:8.0-cudnn6-devel-ubuntu16.04
+
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+        build-essential \
+        cmake \
+        git \
+        graphviz \
+        libatlas-base-dev  \
+        libeigen3-dev \
+        libgoogle-glog-dev \
+        libleveldb-dev \
+        liblmdb-dev \
+        libopencv-dev \
+        libprotobuf-dev \
+        libsnappy-dev \
+        protobuf-compiler \
+        python-dev \
+        python-lmdb \
+        python-matplotlib \
+        python-networkx \
+        python-numpy \
+        python-pil \
+        python-pip \
+        python-protobuf \
+        python-pydot \
+        python-scipy \
+        python-skimage \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN pip install --no-cache-dir --upgrade pip setuptools wheel
+RUN pip install --no-cache-dir future hypothesis
+
+COPY . /opt/caffe2
+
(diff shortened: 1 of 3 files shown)


aaronmarkham:
Hi @lukeyeager, I think this is missing six and future and some tutorial deps like requests and jupyter and matplotlib (unless those are getting pulled in somehow through the other installs). You probably don't need the tutorial deps but I think six and future are things that came along with the brew update.

Also, there are some existing docker images in the [gh-pages branch](https://github.com/caffe2/caffe2/tree/gh-pages/docker). They're a few months old though, so thanks for putting up a new one! I think it makes sense to move new ones into master.

docker/Dockerfile.cuda
```diff
@@ -0,0 +1,49 @@
...
+        python-pil \
+        python-pip \
+        python-protobuf \
+        python-pydot \
+        python-scipy \
+        python-skimage \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN pip install --no-cache-dir --upgrade pip setuptools wheel
+RUN pip install --no-cache-dir future hypothesis
```
lukeyeager: I'm installing `future` here

lukeyeager:
@aaronmarkham thanks for the review! I am already installing future with pip (since no deb package is available), but you're right about six.

I didn't realize there were existing Dockerfiles on the `gh-pages` branch. Would you like for me to close this PR and let you move all those Dockerfiles over to `master`? I'm willing to add a few packages to this Dockerfile, but I'd rather not spend time refactoring all 10 of those old Dockerfiles.

aaronmarkham:
@lukeyeager, I'll move those files over. Regarding some of the differences:
+        libatlas-base-dev  \ 
+        libeigen3-dev \
+        -DCUDA_ARCH_NAME=Manual -DCUDA_ARCH_BIN="35 52 60 61" -DCUDA_ARCH_PTX="61" \
And using /opt for installation whereas I believe the old ones just went into /
Any comments on these? I'm happy to add them, but would like some reference on what the difference or benefit is going to be, so I better understand any new issues that pop up or issues that are going to be resolved.

Thanks!

aaronmarkham:
Now that I have tested the latest docker using much of your configuration let's close this one. Please let me know if I missed something in this [Dockerfile](https://github.com/caffe2/caffe2/pull/949/files#diff-786a81689df3c463e9e1f14cd71db160).

PR #924, opened 2017-07-12
fix strip prefix bug in SaveOp

uni19:
if strip_prefix_ not found in blob name, strip_prefix_.size() characters of blob name will be stripped.


akyrola:
Thanks for the fix!

Issue #468, opened 2017-05-01
Caffe Translator Error: Convolution Layer 


milewis1:
I have a Caffe model that I'm trying to translate into Caffe2. However, I'm running across the following error on the first operator:

RuntimeError: [enforce fail at conv_op_impl.h:25] X.ndim() == filter.ndim(). 4 vs 1 Error from operator: 
input: "data" input: "conv1_w" input: "conv1_b" output: "conv1" type: "Conv" arg { name: "stride" i: 2 } arg { name: "pad" i: 3 } arg { name: "kernel" i: 7 }

The original Caffe model start like this:

`layer {
  name: "data"
  type: "Input"
  top: "data"
  input_param {
    shape {
      dim: 1
      dim: 3
      dim: 224
      dim: 224
    }
  }
}
layer {
  name: "conv1"
  type: "Convolution"
  bottom: "data"
  top: "conv1"
  param {
    lr_mult: 1
    decay_mult: 1
  }
  param {
    lr_mult: 2
    decay_mult: 0
  }
  convolution_param {
    num_output: 64
    bias_term: true
    pad: 3
    kernel_size: 7
    stride: 2
  }
}`

My input is a single color image whose shape going in is: [1, 3, 224, 224]. Has anyone tried to do something similar?

teaglin:
I've had conversion issues too and haven't been able to resolve them yet. I tried a GoogleNet model, which doesn't seem to work in Caffe2, but works great in Caffe. Input is a single color image [1, 3, 224, 224].

`caffe2::EnforceNotMet: [enforce fail at fully_connected_op.h:61] K == W.size() / W.dim32(0). Dimension mismatch: X: 1 1024 2 2, W: 2 1024, b: 2, axis: 1, M: 1, N: 2, K: 4096 Error from operator: 
input: "pool5/7x7_s1" input: "loss3/classifier_w" input: "loss3/classifier_b" output: "loss3/classifier" type: "FC"`

KleinYuan:
Same [here](https://github.com/KleinYuan/Caffe2-iOS/issues/7)
But it means what it says that the [dimension](https://github.com/caffe2/caffe2/blob/0fccf7a4f7bb930c884aa70894f087d85e870b0a/caffe2/operators/fully_connected_op.h#L61) does not match.

littleowl:
[...]

PR #925, opened 2017-07-12
Change basename for proto tensor naming to avoid name collision

danielhauagge:
In `caffe_translator.py`, changed the base string used to name parameter tensors to avoid name clashing. This would happen with in-place operators. The script originally used as the base name the layer's output, this strategy caused all layers that had the same output name (which happens for in-place operators) to have the same base parameter name, which in some cases caused name collisions.

More discussion on this issue [here](https://github.com/caffe2/caffe2/issues/468).

### DIFF
--- a/caffe2/python/caffe_translator.py
+++ b/caffe2/python/caffe_translator.py
@@ -234,7 +234,7 @@ def _TranslateStridePadKernelHelper(param, caffe_op):
 def TranslateConvNd(layer, pretrained_blobs, is_test):
     param = layer.convolution3d_param
     caffe_op = BaseTranslate(layer, "Conv")
-    output = caffe_op.output[0]
+    output = layer.name
     caffe_op.input.append(output + '_w')
 
     AddArgument(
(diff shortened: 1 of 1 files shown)


danielhauagge:
I just signed the CLA, not sure if the message from the facebook-github-bot will update automatically with that info or not.

caffe2/python/caffe_translator.py
```diff
@@ -234,7 +234,7 @@ def _TranslateStridePadKernelHelper(param, caffe_op):
 def TranslateConvNd(layer, pretrained_blobs, is_test):
     param = layer.convolution3d_param
     caffe_op = BaseTranslate(layer, "Conv")
-    output = caffe_op.output[0]
+    output = layer.name
```
harouwu, line 237: Do you think we will also need to change the name of these blobs when they are used as inputs to other ops?

harouwu (changes requested):
Looks like we will have to change the name of ops that are using these outputs as its input.

caffe2/python/caffe_translator.py
danielhauagge: I tested this a bit and seemed to work fine. That said, I'm just starting with caffe2.

orionr:

Issue #514, opened 2017-05-05
Facing issue in building caffe2 with BLAS=MKL


aksgnit:
I am facing issue while building caffe2 with BLAS=MKL

i am using .....
cmake . -DCMAKE_INSTALL_PREFIX=/home/ashish/dp/caffe2/caffe2_install_mklblas -DBLAS=MKL -DMKL_INCLUDE_DIR=/cm/shared/intel_2017/compilers_and_libraries_2017.1.132/linux/mkl/include -DMKL_ROOT=/cm/shared/intel_2017/compilers_and_libraries_2017.1.132/linux/mkl

[ 34%] [ 35%] Building CXX object caffe2/CMakeFiles/Caffe2_CPU.dir/image/image_input_op.cc.o
Building CXX object caffe2/CMakeFiles/Caffe2_CPU.dir/mkl/mkl_operator.cc.o
[ 35%] Building CXX object caffe2/CMakeFiles/Caffe2_CPU.dir/mkl/mklmemory_serialization.cc.o
[ 35%] Building CXX object caffe2/CMakeFiles/Caffe2_CPU.dir/mkl/operators/conv_op.cc.o
[ 36%] Building CXX object caffe2/CMakeFiles/Caffe2_CPU.dir/mkl/operators/conv_op_mkldnn.cc.o
[ 36%] [ 36%] Building CXX object caffe2/CMakeFiles/Caffe2_CPU.dir/mkl/operators/fully_connected_op.cc.o
Building CXX object caffe2/CMakeFiles/Caffe2_CPU.dir/mkl/operators/local_response_normalization_op.cc.o
[ 37%] Building CXX object caffe2/CMakeFiles/Caffe2_CPU.dir/mkl/operators/operator_fallback_mkl.cc.o
In file included from /home/ashish/dp/caffe2/caffe2/utils/mkl_utils.h:63:0,
                 from /home/ashish/dp/caffe2/caffe2/mkl/mklmemory_serialization.cc:3:
[...]

PR #929, opened 2017-07-13
Added sensible default root location for MKL on Windows

### LINKED ISSUE #544: How to build caffe2 in Windows with MKL for fast inference?
I've been trying few days working on build dnn frameworks with MKL as BLAS backend. 
I think caffe2 + mkl would be a very promising solution, but it's hard to link MKL in windows. 
Has anybody work on it?
Thanks~

willyd:
MKL on windows works with this change. Tested with MKL 2017 Update 3 (https://software.intel.com/en-us/articles/intel-math-kernel-library-intel-mkl-2017-release-notes).

Should fix #544 

With MKL 2017 Update 3 #514 should not happen too.

Note: I used Anaconda which ships with its own MKL, so I had to make sure that the MKL 2017 Update 3 version was loaded by replacing the .dll in the `%AnacondaPrefix%\Library\bin` folder. Otherwise, numpy would load it's own version and I would have all sorts of missing procedures errors. Now that the same version is available through `conda` this is easily fixed with `conda install mkl==2017.0.3`


Yangqing:
Thanks!

PR #943, opened 2017-07-20
fix multi label bug in image_input_op

uni19:
As title.


akyrola:
I got following feedback: @akyrola , we intend to save storage by only saving sparse multi-label in label_proto. For example, if we have 10k labels in total, and a particular training example have a multi-label vector [0, 1, 1, 0, .....], we will only store indices [1, 2] in label_proto.

So this diff is not efficient in terms of storage.

pietern:
No comment on feedback. Closing.

PR #949, opened 2017-07-21
Docker move

### LINKED ISSUE #876: Official Docker Image:Failure
Hi
following the [Install](https://caffe2.ai/docs/getting-started.html)  on win 8.1 platform with Docker image:
1-docker pull caffe2ai/caffe2:c2.tutorial.0.7.1         : OK
2-docker run -it caffe2ai/caffe2:c2.tutorial.0.7.1    : OK
3-python -m caffe2.python.operator_test.relu_op_test     :

WARNING:root:This caffe2 python run does not have GPU support. Will run in CPU only mode.
WARNING:root:Debug message: libcuda.so.1: cannot open shared object file: No such file or directory
Segmentation fault

what can i do?im going to use CPU mode only,i have to set it manually?
-thanks

### LINKED ISSUE #923: error while loading shared libraries: libcuda.so.1 when running blob_gpu_test  in caffe2 docker
OS:  Ubuntu 16.04.1 LTS
tag: caffe2ai/caffe2:c2.cuda8.cudnn5.ubuntu16.04
cmd: docker run -ti --device /dev/nvidia0:/dev/nvidia0 --device /dev/nvidiactl:/dev/nvidiactl --device /dev/nvidia-uvm:/dev/nvidia-uvm caffe2ai/caffe2 /bin/bash

aaronmarkham:
Bringing over selected dockerfiles from documentation branch and updated the GPU Dockerfiles to use some of @lukeyeager provided docker configurations. Latest docker with CUDA 8.0 and cuDNN 6 can be pulled via `docker pull caffe2ai/caffe2` or built with `ubuntu-16.04-cuda8-cudnn6-all-options/Dockerfile`. 
**You must use nvidia-docker instead of docker to run the GPU-enabled dockers.** Tutorial files can be overlaid by building `ubuntu-16.04-gpu-tutorial/Dockerfile`. Supersedes #911. Closes #876. Closes #923.


docker/ubuntu-14.04-cpu-all-options/Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+########## REQUIRED DEPENDENCIES ################
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    python-dev \
+    python-pip \
+    git \
+    build-essential \
+    cmake \
+    libprotobuf-dev \
+    protobuf-compiler \
+    libgoogle-glog-dev
```
lukeyeager: > clean up the apt cache by removing /var/lib/apt/lists to reduce the image size, since the apt cache is not stored in a layer
https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#run

docker/ubuntu-14.04-cpu-all-options/Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+
+RUN git clone --recursive https://github.com/caffe2/caffe2.git && \
+    pip install --upgrade pip setuptools wheel && \
+    pip install future hypothesis numpy protobuf six
+
+########## OPTIONAL DEPENDENCIES ################
+
+RUN apt-get install -y --no-install-recommends \
```
lukeyeager: Why not merge this with the lines above for a single (huge) apt-get RUN command?

docker/ubuntu-14.04-cpu-all-options/Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+    python-dev \
+    python-pip \
+    git \
+    build-essential \
+    cmake \
+    libprotobuf-dev \
+    protobuf-compiler \
+    libgoogle-glog-dev
+
+RUN git clone --recursive https://github.com/caffe2/caffe2.git && \
```
lukeyeager: Consider moving the code clone to the end of the script. The code is likely to change faster than the package dependencies, and you'll lose your docker build cache.

docker/ubuntu-14.04-cpu-all-options/Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+    pip install --upgrade pip setuptools wheel && \
+    pip install future hypothesis numpy protobuf six
```
lukeyeager: Consider merging these pip commands into the pip command below.

docker/ubuntu-14.04-cpu-all-options/Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+    libsnappy-dev \
+    openmpi-bin \
+    openmpi-doc \
+    python-numpy \
+    python-pydot \
+    python-setuptools \
+    python-scipy \
+    wget
+
+RUN pip install \
```
lukeyeager: Consider using `--no-cache-dir` to avoid bloating your image with cached wheels.

docker/ubuntu-14.04-cpu-all-options/Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+    jupyter \
+    matplotlib \
+    pydot \
+    python-nvd3 \
+    pyyaml \
+    requests \
+    scikit-image \
+    scipy \
+    setuptools \
+    tornado
```
lukeyeager, line 37: Consider moving this list into a standard, shared `requirements.txt` file.

docker/ubuntu-14.04-cpu-all-options/Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+
+########## INSTALLATION STEPS ###################
+RUN cd caffe2 && make && cd build && make install
+
+ENV PYTHONPATH /usr/local
+ENV LD_LIBRARY_PATH /usr/local/lib:$LD_LIBRARY_PATH
+RUN python -c 'from caffe2.python import core' 2>/dev/null && echo "Success" || echo "Failure"
```
lukeyeager: This shouldn't be necessary.

lukeyeager (changes requested):
Here's a review of the first Dockerfile. I think most of the suggestions will apply to the other Dockerfiles as well.

docker/ubuntu-16.04-cuda8-cudnn6-all-options/Dockerfile
```diff
@@ -0,0 +1,93 @@
+FROM nvidia/cuda:8.0-cudnn6-devel-ubuntu16.04
+MAINTAINER Aaron Markham <aaronmarkham@fb.com>
+
+# caffe2 install with gpu support
+
+########## REQUIRED DEPENDENCIES ################
+RUN apt-get update
```
flx42: This should be combined with the `RUN` below, see the Docker recommendations:
https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#run

docker/ubuntu-16.04-cuda8-cudnn6-all-options/Dockerfile
```diff
@@ -0,0 +1,93 @@
...
+    python-matplotlib \
+    python-networkx \
+    python-numpy \
+    python-pydot \
+    python-setuptools \
+    python-scipy \
+    python-skimage \
+    wget
+
+RUN rm -rf /var/lib/apt/lists/*
```
flx42: This won't save any space, see @lukeyeager's comment

docker/ubuntu-16.04-cuda8-cudnn6-all-options/Dockerfile
```diff
@@ -0,0 +1,93 @@
...
+    pyyaml \
+    requests \
+    scikit-image \
+    scipy \
+    setuptools \
+    tornado
+
+########## INSTALLATION STEPS ###################
+RUN cd caffe2 && mkdir build && cd build
+WORKDIR /caffe2/build/
```
flx42: This `WORKDIR` seems unnecessary, and you should combine all those lines into a single `RUN`:
```
RUN cd caffe2 && mkdir build && cd build
RUN cmake .. \
    -DCUDA_ARCH_NAME=Manual \
    -DCUDA_ARCH_BIN="35 52 60 61" \
    -DCUDA_ARCH_PTX="61" \
    -DUSE_NNPACK=OFF \
    -DUSE_ROCKSDB=OFF
RUN make install
```

flx42 (changes requested):

docker/ubuntu-16.04-cuda8-cudnn6-all-options/Dockerfile
```diff
@@ -0,0 +1,93 @@
...
+RUN cmake .. \
+    -DCUDA_ARCH_NAME=Manual \
+    -DCUDA_ARCH_BIN="35 52 60 61" \
+    -DCUDA_ARCH_PTX="61" \
+    -DUSE_NNPACK=OFF \
+    -DUSE_ROCKSDB=OFF
+RUN make install
+
+ENV PYTHONPATH /usr/local
+ENV LD_LIBRARY_PATH /usr/local/lib:$LD_LIBRARY_PATH
```
flx42: Is the `LD_LIBRARY_PATH` needed here? You can probably just run `ldconfig` after your `make install`

docker/ubuntu-14.04-cpu-all-options/Dockerfile
aaronmarkham: Somewhat of a relic from having core dependencies and optional ones versus optimizing the docker build... definitely should change this!

docker/ubuntu-14.04-cpu-all-options/Dockerfile
aaronmarkham: Yes!

docker/ubuntu-16.04-cuda8-cudnn6-all-options/Dockerfile
aaronmarkham: Yes, this and the other recommendations would be a good idea. I'll move them.

docker/ubuntu-14.04-cpu-all-options/Dockerfile
aaronmarkham: I'm going to leave this for the next round since I've already made so many changes. Plus I really think it would be good to explore pinning versions of the dependencies when doing this, and that will take more research and testing. I could pin versions based on what I have as a working build, but it wouldn't really reflect minimum possible versions that would work. Happy to hear of any good approaches to this.

aaronmarkham:
I think I have updated the files along the lines of the suggestions from you guys. Thanks @lukeyeager and @flx42! The docker images for the [cpu builds are now up on dockerhub](https://hub.docker.com/r/caffe2ai/caffe2/tags/). I'll push the updated gpu image once I've had a chance to test it.

docker/ubuntu-16.04-cuda8-cudnn6-all-options/Dockerfile
aaronmarkham: I just realized I missed this one. Are you suggesting removing `ENV LD_LIBRARY_PATH /usr/local/lib:$LD_LIBRARY_PATH` and adding `ldconfig`? Would it be better to add it to an existing RUN line?

docker/ubuntu-14.04-cpu-all-options/Dockerfile
```diff
@@ -0,0 +1,52 @@
+FROM caffe2ai/caffe2:c2v0.8.0.cpu.min.ubuntu14.04
+MAINTAINER Aaron Markham <aaronmarkham@fb.com>
+
+# caffe2 install with cpu support
+# 14.04 uses an older pip so using -U instead of --no-cache-dir
```
lukeyeager: You're right. Here's what I typically use to get around it:
```Dockerfile
# Don't use deb package because trusty's pip is too old for --no-cache-dir
RUN curl -O https://bootstrap.pypa.io/get-pip.py \
    && python get-pip.py \
    && rm get-pip.py
RUN pip install --no-cache-dir --upgrade pip setuptools wheel
RUN pip install --no-cache-dir future hypothesis pytest
```
If you like having the deb package installed for some reason, you could use this formula instead (though I wouldn't recommend it):
```Dockerfile
RUN apt-get update \
    && apt-get install -y --no-install-recommends \
        python-pip \
    && rm -rf /var/lib/apt/lists/*
# Can't use the --no-cache-dir flag until after pip is upgraded
RUN pip install --upgrade pip
RUN pip install --no-cache-dir --upgrade pip setuptools wheel
RUN pip install --no-cache-dir future hypothesis pytest
```

docker/ubuntu-14.04-cpu-minimal/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+    git \
+    libgoogle-glog-dev \
+    libprotobuf-dev \
+    python-pip \
+    protobuf-compiler \
+    python-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN pip install -U --upgrade pip setuptools wheel && \
+    pip install -U future hypothesis numpy protobuf six
```
lukeyeager: FYI, the second `pip` command here won't actually use the upgraded pip.
https://lukeyeager.github.io/2016/11/29/upgrading-pip.html

lukeyeager (changes requested):

docker/ubuntu-16.04-cuda8-cudnn6-all-options/Dockerfile
flx42: Yes, you probably don't need the `LD_LIBRARY_PATH`. Under `ubuntu:16.04` it should already be in your `ld.so.conf` search path:
```
$ docker run -ti ubuntu:16.04 cat /etc/ld.so.conf.d/libc.conf
# libc default configuration
/usr/local/lib
```
You should call `ldconfig` just after the `make install`, in the same `RUN`.

docker/ubuntu-14.04-cpu-all-options/Dockerfile
```diff
@@ -0,0 +1,72 @@
+FROM ubuntu:14.04
+MAINTAINER Aaron Markham <aaronmarkham@fb.com>
```
flx42: `MAINTAINER` is deprecated, you should use a label instead:
https://docs.docker.com/engine/reference/builder/#maintainer-deprecated

flx42:
By the way, we now have cuDNN v7 images available on DockerHub!

aaronmarkham:
@flx42 Building another image with cuDNN7 now! The dockerfile for that is now part of this PR. I also put in the change for adding ldconfig and removing the LD_LIBRARY_PATH var.
@lukeyeager I updated the pip upgrade section according to your suggestions.

flx42:
Two more remarks: 
- You are not building in parallel, which will be slow, you could do `make -j"$(nproc)" install`
- Your cuDNN v7 Dockerfile doesn't seem to work, cuDNN often breaks the API for major versions, hence the code needs to be modified (and often some `ifdef`s are needed to preserve compatibility with the previous version). You might need to wait for a newer release of caffe2 before enabling cuDNN v7.

lukeyeager:
^^ I think @slayton58 is working on that.

slayton58:
This should have been fixed as of 707aed36e89ab9e2041de25166a4930fc4e24ee7

flx42:
Thanks @slayton58. Looks like the cuDNN v7 Dockerfile uses tag `v0.8.0` like the other Dockerfiles. Therefore this patch isn't included. The Dockerfile could be released later, or made to checkout this particular commit while waiting for a new tagged release.

aaronmarkham:
I'll see if the build runs with cudnn7 and the latest from github and push that as latest if it works and change the dockerfile to not use v0.8.0... maybe we can get an incremental tag 0.8.1 too, so docker script is fixed and always working.
Regarding nproc I took it out because it isn't cross platform. If you have other ideas on how to handle that I'm all ears because it's nice having it in there and not manually changing it.

flx42:
I'm not sure I understand the issue with `nproc`. A Dockerfile doesn't need to be cross-platform.

aaronmarkham:
I ran into an issue with it when running the Dockerfile on my Mac instead of building the image on my Ubuntu desktop. I can try again if it is supposed to work... I just thought it might be something weird with how Docker can or cannot interrogate the hardware of the host, given its OS (similar to windows/docker/gpu access issues), and having it is nice vs being a showstopper.

flx42:
I thought it would work, did it crash during the `docker build`?

aaronmarkham:
Yes, it did break during `docker build`. This is the error (transient and not specific to Mac... it can appear on Ubuntu hosts as well):
`make[3]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.`

Maybe I was running simultaneous builds... It seems to be fine now, so I can put it back in there! Or is there something we can do to guard against this popping up?

Using latest with cuDNN7 went well, so `docker pull caffe2ai/caffe2:latest` gets you Ubuntu 16.04, CUDA 8, cuDNN 7, and today's source. I'll sync that up with a tag here in a bit.

flx42:
This message is usually a symptom of a problem with the Makefile itself. And it's not a fatal error, it will just build serially instead.

PR #986, opened 2017-07-28
fix #983 by remove unsupported archs

### LINKED ISSUE #983: Compile error at resize_op.cu
Win10 + VS2015 sp3 + CUDA 8.0 + Anaconda

Could someone give me a workaround first to let me continue? And please let me know if more info is needed.

`"C:\Users\davin\Documents\Repos\caffe2\build\ALL_BUILD.vcxproj" (default target) (1) ->
"C:\Users\davin\Documents\Repos\caffe2\build\caffe2\Caffe2_GPU.vcxproj" (default target) (6) ->
(CustomBuild target) ->
  C:/Users/davin/Documents/Repos/caffe2/caffe2/operators/resize_op.cu(63): error : identifier "__ldg" is undefined [C:\
Users\davin\Documents\Repos\caffe2\build\caffe2\Caffe2_GPU.vcxproj]`

davinwang:
`resize_op.cu` line63 leverages `__ldg` feature of CUDA, which implies `compute_35` as minimum requirement.

### DIFF
--- a/caffe2/operators/resize_op.cu
+++ b/caffe2/operators/resize_op.cu
@@ -63,7 +63,7 @@ __global__ void NearestNeighborGradientKernel(
 #if __CUDA_ARCH__ >= 350
     atomicAdd(dX + out_index, __ldg(dY + index));
 #else
-	atomicAdd(dX + out_index, *(dY + index));
+    atomicAdd(dX + out_index, *(dY + index));
 #endif
   }
 }
--- a/caffe2/operators/resize_op.cu
+++ b/caffe2/operators/resize_op.cu
@@ -60,7 +60,11 @@ __global__ void NearestNeighborGradientKernel(
     const int out_x = fminf(x / width_scale, output_width - 1);
     const int out_index =
         ((n * num_channels + c) * output_height + out_y) * output_width + out_x;
+#if __CUDA_ARCH__ >= 350
     atomicAdd(dX + out_index, __ldg(dY + index));
+#else
+	atomicAdd(dX + out_index, *(dY + index));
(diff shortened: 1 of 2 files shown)


Yangqing:
@slayton58 FYI.

Yangqing:
Just checking... does it make sense to wrap this in an ifdef statement and not using __ldg if we are under 3.5?

https://stackoverflow.com/questions/16072740/the-behavior-of-cuda-arch-macro

slayton58:
It depends, I would personally say don't support < 3.5, and I would suspect that the vast majority of users will be on >= 3.5 (GK110 or later).

If you were to wrap then `#if __CUDA_ARCH__ >= 350` would be the appropriate way to do it.

davinwang:
CUDA 9.0 will deprecate compute_20 and compute_21, but still support compute_30. Maybe it's better to align with CUDA and wrap it at this moment. 
https://en.m.wikipedia.org/wiki/CUDA
So shall I withdraw my pull request?

Yangqing:
@davinwang thanks for the comment - would you mind leaving 30 to the cmake script and then add a wrap-around? Basically

#if __CUDA_ARCH__ >= 350
        atomicAdd(dX + out_index, __ldg(dY + index));
#else
        atomicAdd(dX + out_index, dY + index);
#endif

davinwang:
@Yangqing @slayton58 Thanks for the info. I made some change to `resize_op.cu` and `cuda.cmake` and only make sure it compiles correctly on Win10+CUDA8. Hope it will pass the unit test if any. :)

Issue #988, opened 2017-07-28
class comment of Context seems out-of-update with the code


wangkuiyi:
The [comment](https://github.com/caffe2/caffe2/blob/master/caffe2/core/context.h#L100) explains that the following two functions are to be implemented in a sub-class of `Context`:

```cpp
 * - static void* New(size_t nbytes): allocates memory.
 * - static void Delete(void* data): deletes memory.
```

However, it seems from the code that what is required is the following function:

```cpp
static std::pair<void*, MemoryDeleter> New(size_t nbytes);
```

Yangqing:
Thanks - will merge.

PR #989, opened 2017-07-28
Update class comment of Context

wangkuiyi:
Fixes https://github.com/caffe2/caffe2/issues/988


Yangqing:
Thanks for catching this!

PR #992, opened 2017-07-29
fix #985 compilation error due to type mismatch

### LINKED ISSUE #985: Compile error at top_k_radix_selection.cuh
env: Win10 + CUDA8.0 + VS2015 sp3 + Anaconda
code branch: v0.8.0
error `caffe2/operators/top_k_radix_selection.cuh(359): error : a value of type "caffe2::TIndex *" cannot be used to initialize an entity of type "long *"` reported in following line of code
`long* indicesSliceStart = &indicesPtr[slice * outputSliceSize];`

davinwang:
Fix the error during compilation on Win10+CUDA, not sure if it affects Linux and MacOS.
caffe2/operators/top_k_radix_selection.cuh(359): error : a value of type "caffe2::TIndex *" cannot be used to initialize an entity of type "long *"

### DIFF
--- a/caffe2/operators/top_k_radix_selection.cuh
+++ b/caffe2/operators/top_k_radix_selection.cuh
@@ -356,7 +356,7 @@ __global__ void gatherTopK(const T* inputPtr,
   // Find the start offset for our slice
   const T* inputSliceStart = &inputPtr[slice * inputSliceSize];
   T* topKSliceStart = &topKPtr[slice * outputSliceSize];
-  long* indicesSliceStart = &indicesPtr[slice * outputSliceSize];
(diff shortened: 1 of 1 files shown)


phg1024 (commented):
Maybe use IndicesType* instead of caffe2::TIndex* so its type is always consistent with the input?

Yangqing:
Yep, thanks - caffe2::TIndex is actually the one to use, IndicesType* is actually used in some operators (such as SparseLengthsSum) where we want to use either int32 or int64 for indices input (the former for performance considerations). In most non-critical cases TIndex is the one to use.

(also cc @dzhulgakov for a double check)

Yangqing:
(travelling and will merge when back)

PR #998, opened 2017-08-01
Fix #997

### LINKED ISSUE #997: Compile Error on Windows
Error during compilation on Win10+CUDA+Visual Studio 2015:

```
  C:/SDKs/caffe2\caffe2/utils/GpuBitonicSort.cuh(42): error : function call must have a constant value in a constant ex
pression [C:\SDKs\caffe2\build\caffe2\Caffe2_GPU.vcxproj]
  C:/SDKs/caffe2\caffe2/utils/GpuBitonicSort.cuh(44): error : function call must have a constant value in a constant ex
pression [C:\SDKs\caffe2\build\caffe2\Caffe2_GPU.vcxproj]
  C:/SDKs/caffe2\caffe2/utils/GpuBitonicSort.cuh(110): error : function call must have a constant value in a constant e
xpression [C:\SDKs\caffe2\build\caffe2\Caffe2_GPU.vcxproj]
```

Seems to be a compiler issue about the template function in `utils/math.h:408`:

```
// Returns true if the given integer type is a power-of-2 (positive only)
template <typename T>
constexpr bool integerIsPowerOf2(T v) {
  return (v && !(v & (v - 1)));
}
```

Manually replace this function with a macro seems to fix the problem
```
#define INTEGER_IS_POWER_OF_2(v) (v && !(v & (v - 1)))
```
Any better solution?

Yangqing:
cc @phg1024

### DIFF
--- a/caffe2/utils/math.h
+++ b/caffe2/utils/math.h
@@ -406,10 +406,17 @@ constexpr T roundUp(T a, T b) {
 }
 
 // Returns true if the given integer type is a power-of-2 (positive only)
+// Note(jiayq): windows reported an error per
+//     https://github.com/caffe2/caffe2/issues/997
+// and as a result will make it a macro.
+#ifdef _MSC_VER
+#define integerIsPowerOf2(v) ((v) && !((v) & ((v) - 1)))
+#else // _MSC_VER
 template <typename T>
 constexpr bool integerIsPowerOf2(T v) {
   return (v && !(v & (v - 1)));
 }
+#endif // _MSC_VER
 
 // Returns log2(n) for a positive integer type
 template <typename T>


willyd:
In my experience this does not fix #997 since the macro will be prefixed with `math::` in https://github.com/caffe2/caffe2/blob/master/caffe2/utils/GpuBitonicSort.cuh#L42.

I fixed this in #878 by using `std::integral_constant` see https://github.com/willyd/caffe2/blob/f556dd4cc2c8f1b0f2508dde5fb2dd94bac898d5/caffe2/utils/GpuBitonicSort.cuh#L42

willyd:
cc @Yangqing

Yangqing:
Thanks! Will work on those PRs.

PR #1000, opened 2017-08-01
Added O2 flag to default compilation path

bwasti:


Yangqing:
Sounds good to me.

taion:
Why the change to `-O2` from `-O3` per https://github.com/bwasti/AICamera/issues/31#issuecomment-319433393?

PR #1027, opened 2017-08-06
Require C++11 support with CMake functions.

williford:
This PR replaces PR #464. It requires C+11 support using the
new CMake variables (`CMAKE_CXX_STANDARD`, `CMAKE_CXX_STANDARD_REQUIRED`,
etc.) when CMake is version 3.1 or above. Otherwise, if CMake is older
(e.g. Ubuntu 14.04) it falls back to using the -std=c++11 flag and
issues a warning.

This PR is based on the comment from @Yangqing:
https://github.com/caffe2/caffe2/pull/464#issuecomment-305376923

The corresponding line in cmake/MiscCheck.cmake is removed in order to
reduce redundancy. Another option would be to move the C++11 logic to MiscCheck.cmake.


hgaiser:
I still feel this kinda defeats the purpose of using these cmake functions in the first place.. what's the benefit now?

Yangqing:
In fact, this breaks contbuild so I'll revert it...

williford:
Sorry for the issues with contbuild. I have just signed up for Travis and will check with it next time.

The main benefit from this code is that it throws an error if c++11 is not supported, rather than throwing a more obscure error on c++11 specific code. This could be accomplished by:

```
  include(CheckCXXCompilerFlag)
  CHECK_CXX_COMPILER_FLAG("-std=c++11" COMPILER_SUPPORTS_CXX11)
  if(NOT COMPILER_SUPPORTS_CXX11)
    MESSAGE(FATAL_ERROR "Your compiler does not support c++11")
  endif()
  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c11")
  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
```

although this would not work for compilers that do not support -std=c++11 (MSVC?).

I'm guessing requiring this is pretty low priority?

PR #1029, opened 2017-08-06
Fix build

Yangqing:
(1) BlobsQueue is causing a gcc error (google search suggeste it was a
bug, but we'll put the implementation in a separate cc file).
(2) Preparing for cuda 9: update cub.
(3) Prepare for cudnn 7: update cudnn rnn op.

### DIFF
--- a/caffe2/core/common.h
+++ b/caffe2/core/common.h
@@ -63,10 +63,12 @@ using std::vector;
 #define CAFFE_NOT_IMPLEMENTED CAFFE_THROW("Not Implemented.")
 
 // suppress an unused variable.
-#ifndef _MSC_VER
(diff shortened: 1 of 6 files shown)


Yangqing:
cc @slayton58 - Simon, seems that this should suffice for cuda 9 / cudnn 7? At least build wise.

slayton58:
Build-wise that looks ok. 

In terms of features, now that CUDA 9 has been merged I will have a bunch of commits for you for cuDNN v7 and TensorCore in the next day or so.

Yangqing:
Committed in 707aed36e89ab9e2041de25166a4930fc4e24ee7 - I had a github import issue so I committed from fb codebase. Same code.

PR #1035, opened 2017-08-07
Allow specifying device to prepare_prediction_net()

davinwang:
This is a supplementary to commit ce8267d425444f60ae650389fb41838847a44a5e. It allows specifying device to prepare_prediction_net() so prediction extractor can work with GPU.


davinwang:
This is a pre-requisite to enable GPU mode in MNIST.ipynb by default. Without this fix, MNIST prediction net will still work, but save/load (via Preditor Extractor) won't work. This is because the program cannot pass device_option to load_from_db() hence the blob will be loaded in CPU instead of GPU.

salexspb:
could you rebase, please and update the diff? The diff is  old something internally in our shipping system have changed, so I can't land it without it being on a more fresh revision.

davinwang:
@salexspb I just update the branch to latest, could you doublecheck please?

PR #1040, opened 2017-08-07
  Implement softmax with NNPACK

Knight-X:
Summary:
  Implement softmax calculation with nnpack package. This would
  accelerate computation in resource limited device.
  Add test case for softmax with NNPACK

### DIFF
--- a/caffe2/contrib/nnpack/nnpack_ops.cc
+++ b/caffe2/contrib/nnpack/nnpack_ops.cc
@@ -335,9 +335,36 @@ class NNPACKLeakyReluOp final : public LeakyReluOp<float, CPUContext> {
  private:
 };
 
+class NNPACKSoftmaxOp final : public Operator<CPUContext> {
+ public:
+  NNPACKSoftmaxOp(const OperatorDef& operator_def, Workspace* ws)
+      : Operator<CPUContext>(operator_def, ws) {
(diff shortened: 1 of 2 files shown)


Yangqing:
@Maratyszcza would you mind taking a look?

caffe2/contrib/nnpack/nnpack_ops.cc
```diff
@@ -335,9 +335,38 @@ class NNPACKLeakyReluOp final : public LeakyReluOp<float, CPUContext> {
...
+    // check if the machine has avx2 support.
+    OPERATOR_NEEDS_FEATURE(GetCpuId().avx2(), "NNPack requires AVX2");
+#endif
+  }
+
+  bool RunOnDevice() override {
+    auto& X = Input(0);
+    const auto N = X.dim32(0), C = X.dim32(1), H = X.dim32(2), W = X.dim32(3);
+    auto* Y = Output(0);
+    auto channels = C * H * W;
```
Maratyszcza: This is actually not right - see https://github.com/caffe2/caffe2/blob/master/caffe2/operators/softmax_op.cc#L9-L13 for the right index calculation.

Knight-X: Right, I will fix the error calculation.

orionr:

PR #1049, opened 2017-08-08
Add cmakelist files for ios build

Yangqing:
After our open source of the ios files, this should allow oss build for the ios related files, especially metal and mpscnn.

Not tested with older xcode that do not have metal, though.


Yangqing:
This is covered by 812c19e15f9d924b3fff0e57109d0bd3de287462 . closing.

PR #1055, opened 2017-08-08
Add NCCL_VERSION_MIN, use v2 API if installed

slayton58:
Basic NCCL 2 API support - the same as applied to gloo [here](https://github.com/facebookincubator/gloo/commit/49586d95569f9c093ed053ac2aa71e9dd3857ae4)

/cc @Yangqing @pietern

PR #1056, opened 2017-08-08
Add TensorCore support

slayton58:
Add support for TensorCore convolution and gemm on Volta hardware.

Currently built on top of #1055

### DIFF
--- a/caffe2/core/common_gpu.cc
+++ b/caffe2/core/common_gpu.cc
@@ -193,6 +193,18 @@ bool GetCudaPeerAccessPattern(vector<vector<bool> >* pattern) {
   return true;
 }
 
+bool TensorCoreAvailable() {
+  // requires CUDA 9.0 and above
+#if CUDA_VERSION < 9000
+  return false;
+#else
+  int device = GetCurrentGPUID();
+  auto& prop = GetDeviceProperty(device);
+
+  return prop.major >= 7;
+#endif
+}
+
 const char* cublasGetErrorString(cublasStatus_t error) {
   switch (error) {
   case CUBLAS_STATUS_SUCCESS:
--- a/caffe2/core/context_gpu.cu
+++ b/caffe2/core/context_gpu.cu
@@ -351,6 +351,14 @@ std::mutex& CUDAContext::mutex() {
   return m;
 }
 
+bool CUDAContext::TensorCoreAvailable() {
+  bool available = caffe2::TensorCoreAvailable();
+  if (!available) {
+    LOG(INFO) << "Requesting Tensor Cores when not available";
+  }
+  return available;
+}
+
 std::vector<long> CUDAContext::TotalMemoryByGpu() {
   std::lock_guard<std::mutex> lock(CUDAContext::mutex());
   CAFFE_ENFORCE(
--- a/caffe2/core/context_gpu.cu
+++ b/caffe2/core/context_gpu.cu
@@ -351,14 +351,6 @@ std::mutex& CUDAContext::mutex() {
   return m;
 }
 
-bool CUDAContext::TensorCoreAvailable() {
-  bool available = caffe2::TensorCoreAvailable();
-  if (!available) {
-    LOG(INFO) << "Requesting Tensor Cores when not available";
-  }
-  return available;
-}
-
 std::vector<long> CUDAContext::TotalMemoryByGpu() {
   std::lock_guard<std::mutex> lock(CUDAContext::mutex());
   CAFFE_ENFORCE(
--- a/caffe2/core/context_gpu.h
+++ b/caffe2/core/context_gpu.h
@@ -163,6 +163,10 @@ class CUDAContext final {
   // deadlocks
   static std::mutex& mutex();
 
+  // Determine whether the device using this context can use
+  // tensor cores
+  static bool TensorCoreAvailable();
+
   // Functions to query memory stats. Only available if flag
   // --caffe2_gpu_memory_tracking is enabled.
   static std::vector<long> TotalMemoryByGpu();
--- a/caffe2/core/context_gpu.h
+++ b/caffe2/core/context_gpu.h
@@ -163,10 +163,6 @@ class CUDAContext final {
   // deadlocks
   static std::mutex& mutex();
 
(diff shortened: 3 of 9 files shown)


slayton58:
\cc @Yangqing

caffe2/core/common_gpu.cc
```diff
@@ -193,6 +193,18 @@ bool GetCudaPeerAccessPattern(vector<vector<bool> >* pattern) {
...
 
+bool TensorCoreAvailable() {
+  // requires CUDA 9.0 and above
+#if CUDA_VERSION < 9000
+  return false;
+#else
+  int device = GetCurrentGPUID();
+  auto& prop = GetDeviceProperty(device);
+
+  return prop.major >= 7;
```
Yangqing, line 204: Do we want to actually cache it? This sounds quite expensive right?

caffe2/core/context_gpu.cu
```diff
@@ -351,6 +351,14 @@ std::mutex& CUDAContext::mutex() {
   return m;
 }
 
+bool CUDAContext::TensorCoreAvailable() {
+  bool available = caffe2::TensorCoreAvailable();
+  if (!available) {
+    LOG(INFO) << "Requesting Tensor Cores when not available";
```
Yangqing: VLOG(1) maybe?

caffe2/core/context_gpu.h
```diff
@@ -163,6 +163,10 @@ class CUDAContext final {
   // deadlocks
   static std::mutex& mutex();
 
+  // Determine whether the device using this context can use
+  // tensor cores
+  static bool TensorCoreAvailable();
```
Yangqing: Why both common_gpu and context_gpu have this file? :)

caffe2/operators/conv_op_cudnn.cc
```diff
@@ -15,9 +15,21 @@ static constexpr size_t kCONV_CUDNN_WORKSPACE_LIMIT_BYTES = 64 * 1024 * 1024;
 // This does not have any performance implications, as we will always find the
 // fastest algorithm; setting them to the right number of algorithms will enable
 // us to best report the statistics when doing an exhaustive search, though.
+#if CUDNN_VERSION_MIN(7,0,0)
```
Yangqing, line 18: A LOT OF VERSIONS! 

(just ranting, this looks great)

caffe2/operators/conv_op_cudnn.cc
```diff
@@ -54,7 +66,8 @@ class CudnnConvOpBase : public ConvPoolOpBase<CUDAContext> {
         deterministic_(
             OperatorBase::GetSingleArgument<int>("deterministic", 0)),
         cudnn_state_(OperatorBase::GetSingleArgument<int>("cudnn_state", 0)),
-        force_algo_(OperatorBase::GetRepeatedArgument<int>("force_algo", vector<int>{-1,-1,-1})) {
+        force_algo_(OperatorBase::GetRepeatedArgument<int>("force_algo", vector<int>{-1,-1,-1})),
+        enable_tensor_core_(OperatorBase::GetSingleArgument<bool>("enable_tensor_core", 0)) {
```
Yangqing: So, do we want to enable tensor core by default? Is it always going to be faster?

If so, I think we should enable it by default, and then do

enable_tensor_core_ &&= TensorCoreAvailable(device_id);

or something.

Yangqing:
Looks great - a few comments.

caffe2/core/common_gpu.cc
slayton58: From memory, `GetDeviceProperty` is already cached by C2 :)

Yangqing:
cc @pietern to take a look at nccl.

caffe2/operators/conv_op_cudnn.cc
slayton58: My decision has always been that it should be opt-in (you're not the first person I've had this discussion with!), because it changes the numerics of the operation.

I've considered that behaviour across architectures should be as close as possible, so fp16 storage on Maxwell, Pascal and Volta should behave as close as they can (they're not identical already, but not too far from it). Enabling TensorCore by default changes that, admittedly very slightly, but I prefer all changes of that sort to be explicitly opt-in.

caffe2/core/context_gpu.h
slayton58: That's a fair point. Originally there was some additional logic in the `CUDAContext::TensorCoreAvailable()` code, but not anymore. Where would you prefer the code to be? common_gpu or context_gpu?

caffe2/core/common_gpu.cc
slayton58: Yeah, it is.

caffe2/core/context_gpu.cu
slayton58: I'll make this change when you've decided where you want the single `TensorCoreAvailable()` method to end up.

caffe2/operators/conv_op_cudnn.cc
slayton58: Upon side discussion with @Yangqing, TensorCore will be enabled by default.

caffe2/utils/math_gpu.cu
```diff
@@ -204,6 +204,98 @@ void Gemm<float16, CUDAContext>(
...
+    const int N,
+    const int K,
+    const float alpha,
+    const float* A,
+    const float* B,
+    const float beta,
+    float* C,
+    CUDAContext* context,
+    TensorProto::DataType math_type) {
+  return Gemm<float,CUDAContext>(TransA,
```
Yangqing, line 224: Is this intended to be the same as Gemm call without TensorCore?

slayton58: Yes. Unless I'm misunderstanding this will use the default engine, which should be cuBLAS for CUDAContext.

Yangqing: Ah yep, realized that it is Gemm<float> and tensorcore is not applicable.

Yangqing:
#1055 landed. Mind rebase?

PR #1069, opened 2017-08-10
Add support for fp16 in convnet_benchmarks.py

sh1r0:


caffe2/python/convnet_benchmarks.py
```diff
@@ -80,25 +82,32 @@ def MLP(order, cudnn_ws):
                 dim_out=d,
                 weight_init=('XavierFill', {}),
                 bias_init=('XavierFill', {}),
+                WeightInitializer=initializer,
```
harouwu: Nice use of Initializer. But I think you can use `arg_scope` in `brew` module to make the whole code cleaner. PR #1093 is a good example of it.

caffe2/python/convnet_benchmarks.py
```diff
@@ -80,25 +82,32 @@ def MLP(order, cudnn_ws):
...
         "last",
         dim_in=d,
         dim_out=1000,
         weight_init=('XavierFill', {}),
         bias_init=('XavierFill', {}),
+        WeightInitializer=initializer,
+        BiasInitializer=initializer
     )
-    xent = model.net.LabelCrossEntropy(["last", "label"], "xent")
+    if use_fp16:
```
harouwu: Do you think it is possible to refactor all of these Loss part so that our model building function can be dtype unaware?

harouwu (changes requested):
Nice code! fp16 is super useful as an example of Caffe2. one major change request is to use`brew.arg_scope` to make code cleaner. Please check out caffe2.ai tutorial section: brewing your model. And PR #1093 is also a good example. 👍

sh1r0:
@harouwu I've updated the commit for `arg_scope`. For the 2nd comment, I think it would be nice if `LabelCrossEntropy`, `Softmax`, etc. operations could be dtype aware (probably implicit dtype conversion).

orionr:

PR #1079, opened 2017-08-11
cudnn conv group support

slayton58:
Enable the new convolution group functionality in cuDNN v7

### DIFF
--- a/caffe2/operators/conv_op_cudnn.cc
+++ b/caffe2/operators/conv_op_cudnn.cc
@@ -462,6 +462,9 @@ bool CudnnConvOp::DoRunWithType() {
       CUDNN_ENFORCE(cudnnSetConvolutionMathType(
             conv_desc_, CUDNN_TENSOR_OP_MATH));
     }
+
+    // enable cuDNN conv groups
+    CUDNN_CHECK(cudnnSetConvolutionGroupCount(conv_desc_, group_));
 #endif
 
     if (force_algo_[ALGO_FWD] >= 0) {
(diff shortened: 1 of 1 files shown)


caffe2/operators/conv_op_cudnn.cc
```diff
@@ -462,6 +462,9 @@ bool CudnnConvOp::DoRunWithType() {
       CUDNN_ENFORCE(cudnnSetConvolutionMathType(
             conv_desc_, CUDNN_TENSOR_OP_MATH));
     }
+
+    // enable cuDNN conv groups
+    CUDNN_CHECK(cudnnSetConvolutionGroupCount(conv_desc_, group_));
```
Yangqing, line 467: Do we need to guard this with ifdef as well?

lukeyeager: It is guarded already (see above)

7oud:
@slayton58 Would the grouped convolution with cudnn 7 be faster than mannually convolution through groups? Could you give some test result?

slayton58:
@7oud In best case special kernels that handle all groups in one shot will be used, worst case all groups in the convolution will be run in streams to try and acheive some overlap. I have no benchmarks handy, but I would strongly expect perf of the new routine to be `>=` the old loop over groups.

sh1r0:
@slayton58 I found that I could not make the group_conv_test (caffe2/python/operator_test/group_conv_test.py) pass with cuda 8 and cudnn 7. Could you have a quick check about that?

sh1r0:
The problem is caused by wrong dimension settings in cudnn tensor descriptors.

slayton58:
@sh1r0 this is fixed in #1117

sh1r0:
@slayton58 Good to know, thanks.

PR #1093, opened 2017-08-15
Add fp16 and tensorcore support to resnet50_trainer

lukeyeager:
Use like `--dtype=float16 --enable-tensor-core`

### DIFF
--- a/caffe2/python/examples/resnet50_trainer.py
+++ b/caffe2/python/examples/resnet50_trainer.py
@@ -263,25 +263,19 @@ def Train(args):
     def create_resnet50_model_ops(model, loss_scale):
         initializer = (pFP16Initializer if args.dtype == 'float16'
                        else Initializer)
-        conv_tensor_core = (True if args.enable_tensor_core else False)
-        fc_engine = ('TENSORCORE' if args.enable_tensor_core else '')
 
-        with brew.arg_scope([brew.conv],
+        with brew.arg_scope([brew.conv, brew.fc],
                             WeightInitializer=initializer,
                             BiasInitializer=initializer,
-                            enable_tensor_core=conv_tensor_core):
-            with brew.arg_scope([brew.fc],
-                                WeightInitializer=initializer,
-                                BiasInitializer=initializer,
-                                engine=fc_engine):
-                pred = resnet.create_resnet50(
-                    model,
-                    "data",
-                    num_input_channels=args.num_channels,
-                    num_labels=args.num_labels,
-                    no_bias=True,
-                    no_loss=True,
-                )
+                            enable_tensor_core=args.enable_tensor_core):
+            pred = resnet.create_resnet50(
+                model,
+                "data",
+                num_input_channels=args.num_channels,
+                num_labels=args.num_labels,
+                no_bias=True,
+                no_loss=True,
+            )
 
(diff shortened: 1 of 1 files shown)


caffe2/python/examples/resnet50_trainer.py
```diff
@@ -258,22 +261,40 @@ def Train(args):
...
-            model,
-            "data",
-            num_input_channels=args.num_channels,
-            num_labels=args.num_labels,
-            label="label",
-            no_bias=True,
-        )
+        initializer = (pFP16Initializer if args.fp16 else Initializer)
+        conv_tensor_core = (True if args.fp16 else False)
+        fc_engine = ('TENSORCORE' if args.fp16 else '')
```
lukeyeager: We could add a little sugar to the fc helper to clean this up. Happy to bundle that change with this PR if desired.

caffe2/python/examples/resnet50_trainer.py
```diff
@@ -338,12 +365,14 @@ def test_input_fn(model):
...
                 batch_size=batch_per_device,
                 img_size=args.image_size,
+                dtype=args.dtype,
             )
 
         data_parallel_model.Parallelize(
             test_model,
             input_builder_fun=test_input_fn,
             forward_pass_builder_fun=create_resnet50_model_ops,
+            post_sync_builder_fun=add_post_sync_ops,
```
harouwu, line 375: Is it also applied to fp32? I am not sure if this will raise error?

caffe2/python/examples/resnet50_trainer.py
```diff
@@ -258,22 +261,35 @@ def Train(args):
...
-            "data",
-            num_input_channels=args.num_channels,
-            num_labels=args.num_labels,
-            label="label",
-            no_bias=True,
-        )
+        initializer = (pFP16Initializer if args.dtype == 'float16'
+                       else Initializer)
+
+        with brew.arg_scope([brew.conv, brew.fc],
```
harouwu, line 267: Nice usage of brew!

harouwu (commented):
Looks good in general! Let's discuss about the inline comment and Travis CI failure, then we are good to go sir.

caffe2/python/examples/resnet50_trainer.py
lukeyeager: It doesn't cause an error. I confess I'm not entirely sure what this step accomplishes - @slayton58 wrote this part.

lukeyeager:
> Travis CI failure

As far as I can tell, this file is not tested at all - so these changes can't possibly cause any additional TravisCI errors.

caffe2/python/examples/resnet50_trainer.py
slayton58: It won't cause an error because it only adds nodes in the graph if there are fp16 and fp32 copies of weights.

When we generate the fp16 and fp32 weights, we then broadcast the _fp16_ copy to all GPUs, but then the fp16 and fp32 copies on a given GPU are out-of-sync. This adds some cast nodes to ensure that all fp16 and fp32 copies on all GPUs are the same.

harouwu (approved):
LGTM when post sync does check blob copy. Let's land it for now. Will have to test later when cuda9 test is supported.

PR #1100, opened 2017-08-16
Fixes compile errors

slayton58:
Seems to be required for CUDA 9 compilation


harouwu (approved):
According to our discussion, let's land this to unblock upcoming PR testing

PR #1103, opened 2017-08-17
cmake: generate macros.h with configure_file()

lukeyeager:
Using file(WRITE) caused the file to be rewritten for every CMake
reconfigure, which was causing unnecessary full rebuilds of the project
even when no source files changed.

The new strategy has the added benefit of enforcing that the macros.h file
is always generated correctly. When the main project relies on this
header for macro definitions (instead of relying on add_definitions()),
we can be more confident that the project will build correctly when used
as a library (which is the whole point of the macros.h file).

Upsides:
* No more unnecessary rebuilds
* Higher confidence that the project will compile properly as a third-party library

Downsides:
* Developers need to add an entry to `macros.h.in` whenever they would have added a new definition with `add_definitions()`


lukeyeager:
Rebased after merge conflict from https://github.com/caffe2/caffe2/pull/1104.

Yangqing (approved):

Yangqing:
Looks great - @harouwu mind merging?

harouwu:
I am working on a fix to disable Travis CI complaining test right now. After that if all passes, lets land it then

PR #1106, opened 2017-08-17
Add random input scaling

slayton58:
Add ability to specify a range for randomly scaling to a new shortest side. For example, for Resnet50 training, one would set `random_scale=[256,480]` in the `ImageInput` operator to resize to a random shortest side in the range [256, 480]

### DIFF
--- a/caffe2/image/image_input_op.h
+++ b/caffe2/image/image_input_op.h
@@ -60,7 +60,7 @@ class ImageInputOp final
   };
 
   bool GetImageAndLabelAndInfoFromDBValue(
-      const string& value, cv::Mat* img, PerImageArg& info, int item_id);
+      const string& value, cv::Mat* img, PerImageArg& info, std::mt19937 *randgen, int item_id);
   void DecodeAndTransform(
       const std::string& value, float *image_data, int item_id,
       const int channels, std::size_t thread_index);
@@ -99,6 +99,10 @@ class ImageInputOp final
   bool use_caffe_datum_;
   bool gpu_transform_;
   bool mean_std_copied_ = false;
+
+  // random minsize
+  vector<int> random_scale_;
+  bool random_scaling_;
 
   // thread pool for parse + decode
   int num_decode_threads_;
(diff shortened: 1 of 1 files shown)


caffe2/image/image_input_op.h
```diff
@@ -515,6 +541,12 @@ bool ImageInputOp<Context>::GetImageAndLabelAndInfoFromDBValue(
   cv::Mat scaled_img;
 
   int scale_to_use = scale_ > 0 ? scale_ : minsize_;
+
+  // set the random minsize
+  if (random_scaling_) {
+    scale_to_use = std::uniform_int_distribution<>(256, 480)(*randgen);
```
harouwu: Do you think we need to give people option to tune the scaled size? Or can we assume that all the image input will be scaled between 256 to 480?

caffe2/image/image_input_op.h
```diff
@@ -142,7 +146,16 @@ ImageInputOp<Context>::ImageInputOp(
       thread_pool_(std::make_shared<TaskThreadPool>(num_decode_threads_)),
       // output type only supported with CUDA and use_gpu_transform for now
       output_type_(
-          cast::GetCastDataType(ArgumentHelper(operator_def), "output_type")) {
+          cast::GetCastDataType(ArgumentHelper(operator_def), "output_type")),
+      random_scale_(
+          OperatorBase::template GetRepeatedArgument<int>("random_scale", {-1,-1})) {
```
harouwu, line 151: Perhaps adding instructions in the op doc will help people understand better why passing -1 as minimized scale will not be using random scaling.

harouwu (commented):
Only 2 incline comments. Travis CI test is a known issue. I would try to at least disable it today.

caffe2/image/image_input_op.h
slayton58: Good point, this is incorrect - it should be taking the bounds from the `random_scale_` vector

harouwu (approved):
Landing this diff. Although I think we still need to add op schema for ImageInputOp

harouwu (changes requested):
op level all looks good. But please add doc for the new arguments here: https://github.com/caffe2/caffe2/blob/master/caffe2/image/image_input_op.cc#L80

PR #1121, opened 2017-08-24
CMake helper to deprioritize Anaconda include path

pietern:
I ran into an issue where a subset of packages were found in the
Anaconda path. This path also contained includes for other packages
and the Anaconda path inadvertently took precendence over the intended
include path. The new `caffe2_include_directories` helper is a hacky
attempt to "fix" this by deprioritizing Anaconda paths in the hope
that intended include paths are searched before Anaconda.


Yangqing:
@ezyang Edward, since you are the master of conda builds - do you think this looks reasonably good?

Yangqing (approved):

ezyang:
I think it's OK. You might also consider deprioritizing `/usr/local/include` similarly; this one causes a lot of troubles for the OS X users!

On the other hand, it doesn't really strike at the root cause, which is that the library searching logic can insert include directories that, in the end, don't end up being the location that gcc/clang ends up using at the end. But maybe it is "good enough"!

pietern:
@ezyang Yes, it's addressing the symptom, not the cause. The "right" way to solve this, or at least signal this, is effectively reimplement a path searcher that checks that a given list of include paths will resolve headers to their expected paths. Fundamentally any include path can include any header, so to fully solve this we would have to copy headers to some staging path where we control where they come from.

ezyang:
@houseroad tracked down a nasty compilation problem (with this patch, conda builds of caffe2 don't work, but "equivalent" hand done builds do): if your source directory has the word "anaconda" in its path (as is the case by default with a conda-build), source include paths will get ordered incorrectly. (This will show up as the error  `"Expected GLOO_USE_CUDA to be defined"`). This was hard to track down because the build system was literally behaving differently depending on the path of the source directory.

Given this, I am not sure we should keep this hack. It doesn't seem like the last time it will trigger in a weird way like this.

PR #1124, opened 2017-08-25
Fix more MKL build issues

Yangqing:
Turns out that due to the cmake improvement by @lukeyeager , we now no longer rely on compiler flags but on the macros.h file to obtain CAFFE2_USE_MKL. This requires some minor changes in the MKL implementation to properly capture the macro before testing it.


lukeyeager:
Sorry for the annoyance from https://github.com/caffe2/caffe2/pull/1103. Hopefully this means C2 will be easier to use as a 3rd party library, though!

Yangqing:
no problem, thanks @lukeyeager - that was a great work to reduce unnecessary rebuilds!

Issue #1135, opened 2017-08-28
rasbian build error in workerspool.h


Knight-X:
In file included from /home/pi/workspace/caffe2/caffe2/utils/threadpool/ThreadPool.cc:2:0:
/home/pi/workspace/caffe2/caffe2/utils/threadpool/WorkersPool.h: In member function ‘void caffe2::WorkersPool::Execute(const std::vector<std::shared_ptr<caffe2::Task> >&)’:
/home/pi/workspace/caffe2/caffe2/utils/threadpool/WorkersPool.h:364:60: error: use of ‘auto’ in lambda parameter declaration only available with -std=c++14 or -std=gnu++14
     std::for_each(++tasks.begin(), tasks.end(), [this, &n](auto& task) {
                                                            ^~~~
/home/pi/workspace/caffe2/caffe2/utils/threadpool/WorkersPool.h: In lambda function:
/home/pi/workspace/caffe2/caffe2/utils/threadpool/WorkersPool.h:365:37: error: request for member ‘get’ in ‘task’, which is of non-class type ‘int’
       workers_[n++]->StartWork(task.get());
                                     ^~~
In file included from /usr/include/c++/6/algorithm:62:0,
                 from /home/pi/workspace/caffe2/caffe2/core/common.h:4,
                 from /home/pi/workspace/caffe2/caffe2/utils/threadpool/WorkersPool.h:1,
                 from /home/pi/workspace/caffe2/caffe2/utils/threadpool/ThreadPool.cc:2:
/usr/include/c++/6/bits/stl_algo.h: In instantiation of ‘_Funct std::for_each(_IIter, _IIter, _Funct) [with _IIter = __gnu_cxx::__normal_iterator<const std::shared_ptr<caffe2::Task>*, std::vector<std::shared_ptr<caffe2::Task> > >; _Funct = caffe2::WorkersPool::Execute(const std::vector<std::shared_ptr<caffe2::Task> >&)::<lambda(int&)>]’:
/home/pi/workspace/caffe2/caffe2/utils/threadpool/WorkersPool.h:366:6:   required from here
/usr/include/c++/6/bits/stl_algo.h:3769:5: error: no match for call to ‘(caffe2::WorkersPool::Execute(const std::vector<std::shared_ptr<caffe2::Task> >&)::<lambda(int&)>) (const std::shared_ptr<caffe2::Task>&)’
  __f(*__first);
  ~~~^~~~~~~~~~
[...]

PR #1136, opened 2017-08-28
fix build error in workersthread.h

Knight-X:
According #1135, fix error in C++11, because using auto as a parameter of lambda is supported by c++14. In order to build in c++11, it should declare with type.


pietern:
Thanks for the PR. This was fixed in 076a4a99f6c4026c7fe47f2062b256063059cafa.

PR #1140, opened 2017-08-29
Include Caffe2 headers before anything else

pietern:
This was a tricky one to debug. After pulling from master, my build
was complaining that certain identifiers in updated source files were
undefined. After building with VERBOSE=1, extracting the compilation
commands, and adding -M, I saw that CMake had included the Caffe2
installation directory as include path. Worse yet, this path had
precedence over the path to the actual source code. The compiler
included older headers when compiling newer source files.

This change forces the path to the Caffe2 source code to take
precedence over all other include paths. The only path that takes
precedence over *that* path is PROJECT_BINARY_DIR, which holds the
headers that are generated at compile time.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,12 +116,17 @@ if (CAFFE2_CPU_FLAGS)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CAFFE2_CPU_FLAGS}")
 endif()
 
-# ---[ Include path needed for proto
-include_directories(BEFORE ${PROJECT_BINARY_DIR})
+# Prefix path to Caffe2 headers.
+# If a directory containing installed Caffe2 headers was inadvertently
+# added to the list of include directories, prefixing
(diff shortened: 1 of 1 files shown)


pietern:
To clarify a bit more:

In my setup I had LD_LIBRARY_PATH contain the path to the Caffe2 installation directory. In a previous build/install, CMake installed libbenchmark.so there as well. On the subsequent build, the CMake module

lukeyeager:
This partially addresses one of the main concerns in https://github.com/caffe2/caffe2/pull/1112, but still won't avoid useless rebuilds. We need to convince CMake to stop searching the install directory for 3rd party libraries so we will continue to use the files in `third_party/` for rebuilds.

pietern:
@lukeyeager Yep, re: rebuilds. Great analysis in #1112.

lukeyeager:
Verified that the problematic `-isystem /caffe2/install/include` from https://github.com/caffe2/caffe2/pull/1112 appears later in the list, but still present.

PR #1144, opened 2017-08-29
tests: don't test SparseLengthsSum with fp16+CUDA

lukeyeager:
Fixes issue reported here: https://github.com/caffe2/caffe2/commit/8cbea8d3cae011467978d18e0a667cf53eb6319c#commitcomment-23958756

/cc @dzhulgakov


lukeyeager:
Fixed in https://github.com/caffe2/caffe2/commit/98e1b71960585f627561116df16edc4f4ad2833b.

PR #1151, opened 2017-08-30
Add fp16 support for various ops used by RNNs

lukeyeager:
This is a joint effort from @pooyadavoodi, @slayton58 and me.


slayton58:
@Yangqing It'd be good to get this looked at quickly if possible - the size of the PR means rebasing is going to be painful..

caffe2/operators/batch_matmul_op.h
```diff
@@ -65,29 +74,32 @@ class BatchMatMulOp final : public Operator<Context> {
...
+        use_scratch_ ? scratch_ : nullptr);
     return true;
   }
 
  protected:
   bool trans_a_;
   bool trans_b_;
+
+  bool use_scratch_;
+  Tensor<Context>* scratch_;
```
akyrola, line 102: better to use a non-pointer here, or shared/unique_ptr, as this is not cleared otherwise in destructor

caffe2/operators/lstm_unit_op.h
```diff
@@ -28,7 +29,7 @@ void LSTMUnit(
     bool drop_states,
     T* C,
     T* H,
-    const T forget_bias,
+    const float forget_bias,
```
akyrola, line 32: Sorry I don't get this, why remove the T template from lstm_unit?

caffe2/python/helpers/algebra.py
```diff
@@ -16,3 +16,11 @@ def transpose(model, blob_in, blob_out, use_cudnn=False, **kwargs):
 def sum(model, blob_in, blob_out, **kwargs):
     """Sum"""
     return model.net.Sum(blob_in, blob_out, **kwargs)
+
+def batch_mat_mul(model, blob_in, blob_out,
```
akyrola, line 20: nice

akyrola (commented):
This is great work!

Two comments, and after that I will import this and run some tests on our side as well. 

Can you run at least the following tests:

  operator_test/rnn_cell_test
  operator_test/recurrent_net_executor_test
  operator_test/cudnn_recurrent_net_test

akyrola:
+ needs rebase

lukeyeager:
Rebased (but now I need to go convert the `RNNApplyLinkOp` to fp16 as well).

> Can you run at least the following tests

Will do. Just so you know, this code (minus rebasing) is already integrated in our internal branches, which has pretty comprehensive automated GPU testing.

@akyrola I'm thinking about trying to break this down into smaller PRs so it's easier to review and so simple things can get merged quickly without getting bogged down by more complicated parts (e.g. tests for RecurrentNetworkOps). Any concern with that?

akyrola:
Splitting to smaller PRs is always a good idea :)

lukeyeager:
Great. I'll start doing that. I'll note when I've addressed feedback in this PR inside new PRs.

caffe2/operators/batch_matmul_op.h
lukeyeager: Done in https://github.com/caffe2/caffe2/pull/1175

lukeyeager:
Ok, I finally broke this all out into separate PRs. Sorry - I didn't realize how many there would be!

PR #1177, opened 2017-09-07
Add transport/interface arguments to CreateCommonWorld operator

pietern:
These arguments control which Gloo transport (TCP or IB) and which
network interface is used for the common world. If not specified, it
defaults to using TCP and the network interface for the IP that the
machine's hostname resolves to.

The valid values for the transport argument are "tcp" and "ibverbs".
For ibverbs to work, Gloo must have been compiled with ibverbs
support. If Gloo is built as part of Caffe2 (sourced from the
third_party directory), then you can pass -DUSE_IBVERBS=ON to CMake to
enable ibverbs support in Gloo.

### DIFF
--- a/caffe2/contrib/gloo/common.cc
+++ b/caffe2/contrib/gloo/common.cc
@@ -1,12 +1,49 @@
 #include "caffe2/contrib/gloo/common.h"
 
+#include "caffe2/core/logging.h"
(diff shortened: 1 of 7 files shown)


pietern:
Reimport of #1176. Thought GitHub was broken but it was just a delay on the webhook delivery.

PR #1182, opened 2017-09-07
fp16: add test for FC

lukeyeager:
fp16 and TensorCore support was already added to the op in https://github.com/caffe2/caffe2/pull/1056. This adds a test.


lukeyeager:
Tested with `max_examples=1000` to make sure the threshold and stepsize are set appropriately for fp16. Pulled examples back down to 50 so the testsuite doesn't actually take that long.

PR #1185, opened 2017-09-07
Use brew in rnn_cell.py

lukeyeager:
Was https://github.com/caffe2/caffe2/pull/1151.

### DIFF
--- a/caffe2/python/rnn_cell.py
+++ b/caffe2/python/rnn_cell.py
@@ -257,12 +257,10 @@ class LSTMCell(RNNCell):
 
         if extra_inputs is not None:
             extra_input_blobs, extra_input_sizes = zip(*extra_inputs)
-            fc_input, _ = model.net.Concat(
+            fc_input = brew.concat(
+                model,
                 [hidden_t_prev] + list(extra_input_blobs),
-                [
-                    self.scope('gates_concatenated_input_t'),
-                    self.scope('_gates_concatenated_input_t_concat_dims'),
-                ],
+                self.scope('gates_concatenated_input_t'),
                 axis=2,
             )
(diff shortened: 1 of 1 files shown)


lukeyeager:
Should probably merge https://github.com/caffe2/caffe2/pull/1184 before testing this.

caffe2/python/rnn_cell.py
```diff
@@ -622,7 +617,8 @@ def _apply(
                     layer_next_states,
                 )
                 if i > 0 and i in self.residual_output_layers:
-                    layer_input = model.net.Sum(
```
akyrola, line 625: This is failing with error:
  "indentation contains mixed spaces and tabs"

akyrola: Otherwise diff looks good

lukeyeager: Oops! Sorry about that.

caffe2/python/rnn_cell.py
```diff
@@ -481,12 +478,10 @@ def _prepare_output_sequence(self, model, state_outputs):
...
             with core.NameScope(self.name or ''):
-                output, _ = model.net.Dropout(
+                output = brew.dropout(
+                    model,
                     output,
-                    [
-                        str(output) + '_with_dropout_mask{}'.format(self.mask),
-                        str(output) + '_dropout_mask{}'.format(self.mask),
-                    ],
+                    str(output),
```
akyrola, line 484: Here was unfortunately a bug gone undetected. The output should not be in-place, that causes problems with gradient generation in some cases.

akyrola: Fix is submitted, landing soon

PR #1209, opened 2017-09-12
tests: use assertRaises, not expectedFail

lukeyeager:
I would expect that tests marked "expected failure" mean that there is a known issue in the code which will be fixed later. Both of these tests are simply verifying proper error-checking - nothing needs fixing.

Before (looks like something is wrong):
```
======================================= 2 xfailed in 0.27 seconds =======================================
```
After:
```
======================================= 2 passed in 0.28 seconds ========================================
```
/cc @akyrola @gsethi523


akyrola:
makes sense, thanks!

PR #1214, opened 2017-09-15
Fixes for NCCLReduce with non-zero root

slayton58:
All other NCCL ops expect paired src, dst pointers for each
GPU. Reduce doesn't, and the old logic would always set dst for
rank = 0 regardless of whether that was the root or not.
This change takes into account that Reduce only has one output, and it
should assign dst only for the root rank. Also changes the schema to
allow inplace for any input and Output(0).

### DIFF
--- a/caffe2/contrib/nccl/cuda_nccl_op_gpu.cc
+++ b/caffe2/contrib/nccl/cuda_nccl_op_gpu.cc
@@ -199,7 +199,7 @@ OPERATOR_SCHEMA(NCCLReduce)
     .InputsCanCrossDevices()
     .AllowInplace([](int in, int out) -> bool {
       return (out == 0);
-    });
+    })
     .DeviceInferenceFunction(ncclOpDevInfer);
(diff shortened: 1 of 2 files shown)


pietern:
Nice, thanks @slayton58!

pietern:
@slayton58 This conflicted with c96d28557e8efcf9e401383fa50ef051318ffeb8, could you please rebase?

slayton58:
@pietern Done

caffe2/contrib/nccl/cuda_nccl_op_gpu.cc
```diff
@@ -193,7 +197,9 @@ OPERATOR_SCHEMA(NCCLReduce)
     .NumOutputs(1)
     .IdenticalTypeAndShapeOfInput(0)
     .InputsCanCrossDevices()
-    .AllowInplace({{0, 0}})
+    .AllowInplace([](int in, int out) -> bool {
+      return (out == 0);
+    });
```
pietern: The ; must go to make DeviceInferenceFunction work.

PR #1238, opened 2017-09-20
Fix memonger test

Yangqing:
(waiting for travis)


Yangqing:
nvm, seems to be a python 3 + networkx issue and I don't have a proper environment to test it right now... will address later.

PR #1242, opened 2017-09-21
Partially fix memonger with networkx 2.0

Yangqing:
This fixes the apparent discrepancy (list vs iterator). After this, there are still 3 failures regarding topological sort but that seems a bit involved. Someone shall look deeper.


Yangqing:
Remaining failures:

https://travis-ci.org/Yangqing/caffe2/jobs/278072558

Issue #1249, opened 2017-09-23
caffe2 build success with -DUSE_CUDA=ON, but install failure on MacOS v10.12.6


Setogit:
1. Followed the instruction on:
```
https://caffe2.ai/docs/getting-started.html?platform=mac&configuration=compile
```
2. `cmake -DUSE_CUDA=ON ..` succeeded as follows:
```
  -- ******** Summary ********
  -- General:
  --   Git version           : 
  --   System                : Darwin
...
  --   USE_CUDA              : ON
  --     CUDA version        : 8.0
  --     CuDNN version       : 6.0.21
...
  -- Configuring done
  -- Generating done
  -- Build files have been written to: /Users/seto/Caffe2/caffe2/build
```
3. `sudo make install` failed as follows.
Note: I've successfully built and installed PyTorch from source with CUDA enabled on the same Mac OSX.
```
...
[ 58%] Building NVCC (Device) object caffe2/CMakeFiles/Caffe2_GPU.dir/operators/Caffe2_GPU_generated_layer_norm_op.cu.o
/Users/seto/Caffe2/caffe2/caffe2/operators/layer_norm_op.cu:264:54: error: non-constant-expression cannot be narrowed from type 'int' to 'unsigned long' in initializer list [-Wc++11-narrowing]
(gscratch_).Resize(std::__1::vector< unsigned long> {left, right});
                                                     ^~~~
...
CMake Error at Caffe2_GPU_generated_layer_norm_op.cu.o.Release.cmake:278 (message):
  Error generating file
  /Users/seto/Caffe2/caffe2/build/caffe2/CMakeFiles/Caffe2_GPU.dir/operators/./Caffe2_GPU_generated_layer_norm_op.cu.o


make[2]: *** [caffe2/CMakeFiles/Caffe2_GPU.dir/operators/Caffe2_GPU_generated_layer_norm_op.cu.o] Error 1
make[1]: *** [caffe2/CMakeFiles/Caffe2_GPU.dir/all] Error 2
make: *** [all] Error 2
```

Yangqing:
Fix landing in #1251 .

PR #1251, opened 2017-09-23
Fix OSX build w/CUDA=ON

Setogit:
connect to https://github.com/caffe2/caffe2/issues/1249

With this change, build, install, and smoke test pass on OSX with CUDA=ON.
```
$ cmake -DUSE_CUDA=ON ..
$ sudo make install
$ python -c 'from caffe2.python import core' 2>/dev/null && echo "Success" || echo "Failure"
Success
```


Yangqing:
Thanks for the fix!

PR #1256, opened 2017-09-24
Proper versioning and misc CMake improvements

Yangqing:
This brings proper versioning in Caffe2: instead of manual version macros, this puts the version information in CMake (replacing the TODO @bwasti line) and uses macros.h.in to then generate the version in the C++ header.

A few misc updates:
- Removed the mac os rpath, verified on local macbook that it is no longer needed.
- Misc updates for caffe2 ready:
  - Mapped cmake/Cuda.cmake with gloo's setting.
  - upstreamed third_party/nccl so it builds with cuda 9.
- Separated the Caffe2 cpu dependencies and cuda dependencies
  - now libCaffe2_CPU.so do not depend on any cuda libs.
  - caffe2 python extensions now depend on cpu and gpu separately too.
- Reduced the number of unused functions in Utils.cmake


lukeyeager:
@Yangqing when syncing `Cuda.cmake` with gloo, you removed two useful features: `CUDA_ARCH_NAME=Auto` and `CUDA_ARCH_NAME=Manual`. Is there a reason you don't like those features? Would you be open to PRs which bring this behavior back into gloo and C2?

PR #1258, opened 2017-09-24
Fix a caffe2-gloo dependency problem

houseroad:
The problem: 
Building caffe2 fails because the installed directory contains "anaconda".

The cause:
Compiling Gloo will generate a new config.h file in the binary folder.
If we put the original config.h in front, the compiler will complain "Expected GLOO_USE_CUDA to be defined".  

~~~Switch the positions of the include folders can solve the problem.~~~

Function caffe2_include_directories in cmake/Utils.cmake is a little bit hacky. If the directory contains "anaconda", it will append the new include directory after existing include path. Otherwise it will insert the directory before the path. So in the first case, the directories are inserted in order, and in the latter one, they are inserted reversely.

The solution:
See the commit.

@pietern #1121


pietern:
Nice catch. Agree it is hacky, but the only way we can fix this...

PR #1263, opened 2017-09-26
Caffe2 Benchmarking Framework

sf-wind:
Implement a framework to benchmark the Caffe2 inferencing time. It only contains the observer collecting the delay information for running the net and the operator. The driver of the benchmark is in a separate repository.

It does not interfere with the rest of the Caffe2.


caffe2/CMakeLists.txt
```diff
@@ -11,6 +11,9 @@ set(Caffe2_GPU_SRCS)
 set(Caffe2_CPU_TEST_SRCS)
 set(Caffe2_GPU_TEST_SRCS)
 
+# Caffe2_CPU_OBSERVER_SRCS is the list of files used for benchmarking purpose
+set(Caffe2_CPU_OBSERVER_SRCS)
```
bwasti: Is this ever needed at global scope?  I think the variable can exist exclusively in the `share` directory or even the `share/contrib/observers` directory

caffe2/share/contrib/benchmark_harness/git_driver.py
```diff
@@ -0,0 +1,109 @@
...
+        self.git.pull(getArgs().git_repository, getArgs().git_branch)
+        self.git.checkout(getArgs().git_commit)
+        new_commit_hash = self.git.run('rev-parse', 'HEAD').rstrip()
+        if new_commit_hash == self.commit_hash:
+            getLogger().info("Commit %s is already processed.", new_commit_hash)
+            return False
+        self.commit_hash = new_commit_hash
+        self.commit_hash_time = self.git.run('show', '-s', '--format=%ct', new_commit_hash)
+        if getArgs().android:
+            # shutil.rmtree(getArgs().git_dir + "/build_android")
```
bwasti: rm

caffe2/share/contrib/benchmark_harness/git_driver.py
```diff
@@ -0,0 +1,109 @@
...
+            build_android = getArgs().git_dir + "/scripts/build_android.sh"
+            processRun(build_android)
+        if getArgs().host:
+            # shutil.rmtree(getArgs().git_dir + "/build")
```
bwasti: rm

caffe2/share/contrib/benchmark_harness/platforms/android/adb.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python
+
+from utils.subprocess_with_logger import processRun
+import os.path as path
+from utils.custom_logger import getLogger
+
+class ADB(object):
+    def __init__(self, device=None):
+        self.device = device
+        self.dir = "/data/local/tmp/"
```
bwasti: might be nice to expose this to a command line flag

caffe2/share/contrib/benchmark_harness/platforms/host/host_platform.py
```diff
@@ -0,0 +1,56 @@
...
+            "--net", getArgs().net,
+            "--input", getArgs().input,
+            "--warmup", str(getArgs().warmup),
+            "--iter", str(getArgs().iter),
+            ]
+        if getArgs().program:
+            cmd.insert(0, getArgs().program)
+        elif getArgs().exec_base_dir:
+            cmd.insert(0, getArgs().exec_base_dir +
+                '/build/caffe2/share/contrib/binaries/caffe2_benchmark/binaries/caffe2_benchmark')
```
bwasti: this seems a bit fragile, is there any way to make it more robust to future changes?

caffe2/share/contrib/binaries/caffe2_benchmark/caffe2_benchmark.cc
```diff
@@ -0,0 +1,143 @@
...
+    }
+  }
+
+  // Run main network.
+  CAFFE_ENFORCE(ReadProtoFromFile(caffe2::FLAGS_net, &net_def));
+  caffe2::NetBase* net = workspace->CreateNet(net_def);
+  CHECK_NOTNULL(net);
+
+  LOG(INFO) << "Starting benchmark.";
+  caffe2::QPLConfig::initSampleRate(
```
bwasti: use of `QPLConfig` seems to be the only differentiating factor between this and `speed_benchmark.cc`.  I would recommend adding an option to use `QPLConfig` in the original file rather than copy it all over to here.

caffe2/share/contrib/observers/observer.cc
```diff
@@ -0,0 +1,142 @@
...
+    return true;
+}
+}
+
+REGISTER_CAFFE2_EARLY_INIT_FUNCTION(
+    registerGlobalTimeNetObserverCreator,
+    &registerGlobalTimeNetObserverCreator,
+    "Caffe2 net global observer creator");
+
+TimeNetObserver::TimeNetObserver(NetBase* subject_)
```
bwasti: Perhaps call this QPLObserver and template it on `<NetBase*>` and `<OperatorBase*>`?

caffe2/share/contrib/observers/observer.h
```diff
@@ -0,0 +1,59 @@
+#pragma once
```
bwasti: This file should be named something that exposes the class name.  there already exists an `observer.h` in `caffe2/core` so it gets confusing to have this

bwasti (changes requested):
looks great! a couple of questions:

why did you create a `share/contrib` directory rather than use `contrib`?

you have some changes to third_party submodules, are these necessary?

caffe2/share/contrib/binaries/caffe2_benchmark/caffe2_benchmark.cc
sf-wind: The main idea is to use the observer approach to collect the runtime metrics. This way, we can use the same mechanism to collect metrics on mobile apps. It has an added benefit that there is no need to add a benchmarking method whenever a new net type is added. My intention is to replace speed_benchmark with this new mechanism.

caffe2/share/contrib/observers/observer.cc
sf-wind: I think I need to change the name QPLObserver since it doesn't make much sense. I'm not sure about your comment on templating on <OperatorBase*>

caffe2/share/contrib/observers/observer.h
sf-wind: This is not a good name. The class name is not good either. How about I change the filename to be perf_observer.h and the class name to be PerfNetObserver?

sf-wind:
@bwasti , Thanks. I put it in share/contrib because I want to reuse it in mobile. 

I'm not sure why it changes submodules. I just did a pull from the master and push to this branch and the submodules are automatically changes. How to avoid that? Thanks.

caffe2/CMakeLists.txt
sf-wind: Can you suggest how? After I move it to the top of the share directory, the build fails because the files are not found.

PR #1264, opened 2017-09-26
Compile with MKL in conda-build

houseroad:
Problem:
Without -DBLAS=MKL, conda-build won't include MKL library into Caffe2 build. And the BLAS performance is bad on CPU.

Solution:
Explicitly add the flag. Add mkl and mkl-include as dependencies.

@ezyang @Yangqing

### DIFF
--- a/conda/build.sh
+++ b/conda/build.sh
@@ -14,7 +14,7 @@ PYTHON_ARGS="$(python ./scripts/get_python_cmake_flags.py)"
 
 mkdir -p build
 cd build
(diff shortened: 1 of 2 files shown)


ezyang:
LGTM. (in chat, I mentioned that it was surprising that mkl-include was needed for the build, since PyTorch doesn't seem to need it. Shrug.)

houseroad:
Let me check this with Pytorch team.

Yangqing:
Chatted offline and it's because PyTorch uses explicit BLAS interface only, while C2 MKL mode also uses MKLDNN and a bunch of others. In any case, thanks for the fix! :)

uddmorningsun:
@ezyang   Any ideas to support caffe2 GPU version conda-build recipe ? Thanks !

ezyang:
It's supported; at least, I have successfully used an edited version of the conda recipe to build caffe2 with GPU support.

PR #1271, opened 2017-09-29
Expose CMake options in the binary

dzhulgakov:
Useful for figuring out with people which version they built with. We can just ask for --caffe2_version gflag or get core.build_options from python.

Also adds CMAKE_INSTALL_RPATH_USE_LINK_PATH - without it wasn't building on my Mac. How should it be tested?


caffe2/core/common.cc
```diff
@@ -18,4 +18,13 @@
 
 namespace caffe2 {
 bool g_caffe2_has_cuda_linked = false;
+
+const std::map<string, string>& GetBuildOptions() {
+#ifndef CAFFE2_BUILD_STRINGS
+#define CAFFE2_BUILD_STRINGS {{"BUILD_SETTINGS", "unavailable"}}
```
bddppq: I think it's better to return an empty dict.

caffe2/python/workspace.py
```diff
@@ -55,6 +55,7 @@
 operator_tracebacks = defaultdict(dict)
 
 is_asan = C.is_asan
+build_options = C.get_build_options()
```
bddppq: I started putting some build information at https://github.com/caffe2/caffe2/blob/master/caffe2/python/build.py#L23. Can we merge them together?

Yangqing:
Random comment: it's probably a good time to consider finally writing a casse2Config.cmake file :)

lukeyeager:
Nice. I always wished C1 had this feature: https://github.com/BVLC/caffe/issues/2495.

PR #1295, opened 2017-10-04
Bump gloo

pietern:
Latest version of Gloo takes care of MPI_Init/MPI_Finalize for us, so
this commit removes handling that from caffe2/contrib/gloo. It also
imports CMake NCCL module changes from Gloo to stay consistent and
allow setting NCCL_INCLUDE_DIR and NCCL_LIB_DIR separately.


pietern:
cc @lukeyeager

lukeyeager (approved):
Tested nccl detection with cmake and mpi teardown. Looks good!

PR #1300, opened 2017-10-05
[rfc] Rename library name to lower

Yangqing:
In the past we call our libraries libCaffe2_CPU.so and libCaffe2_GPU.so that don't really match the usual linux so library naming conventions. This diff changes it to libcaffe2.so (old Caffe2_CPU) and libcaffe2_gpu.so (old Caffe2_GPU).

This might affect existing building scripts that explicitly use Caffe2_CPU and Caffe2_GPU: what do you guys think? @pietern @bwasti @slayton58


slayton58:
We don't explicitly refer to `libCaffe2_CPU.so` by name at all, so this shouldn't affect our containers :)

Issue #793, opened 2017-06-13
'caffe2.python.caffe2_pybind11_state_gpu.Workspace' object has no attribute 'is_empty'


lukeyeager:
*(previously discussed at https://github.com/caffe2/caffe2/issues/700)*

In our internal CI system, we get intermittent failures like this:

* `Exception: No gradient registered for AveragedLoss. Exception from creating the gradient op: 'caffe2.python.caffe2_pybind11_state_gpu.Workspace' object has no attribute 'is_empty'.`
* `E0531 21:27:41.782681   337 pybind_state.cc:190] Exception encountered running PythonOp function: <type 'exceptions.AttributeError'>: 'caffe2.python.caffe2_pybind11_state_gpu.Workspace' object has no attribute 'shape'`
* `E0608 15:48:40.626631   363 pybind_state.cc:190] Exception encountered running PythonOp function: <type 'exceptions.AttributeError'>: 'caffe2.python.caffe2_pybind11_state_gpu.Workspace' object has no attribute 'reshape'`

I've looked for a piece of code which exhibits something like this behavior, but haven't found it anywhere yet:
```python
blobs = []
while loop_condition:
  w = workspace()
  if common_condition:
    w = blob()
  blobs.append(w)  # could be a workspace or a blob
blobs[0].is_empty()  # fails if blobs[0] is a workspace instead of a blob
```

salexspb:
cc @tomdz

lukeyeager:
We continue to hit these failures intermittently. Here are some examples from the past few days:
```
__________________________ PythonOpTest.test_gradient __________________________
E0626 17:52:56.731902   388 pybind_state.cc:278] Exception encountered running PythonOp function: <type 'exceptions.AttributeError'>: 'caffe2.python.caffe2_pybind11_state_gpu.Workspace' object has no attribute 'reshape'

______________ PythonOpTest.test_gradient_multiple_with_indicies _______________
E0627 02:18:49.782148   388 pybind_state.cc:278] Exception encountered running PythonOp function: <type 'exceptions.AttributeError'>: 'caffe2.python.caffe2_pybind11_state_gpu.Workspace' object has no attribute 'reshape'
E0627 02:18:49.799831   388 pybind_state.cc:278] Exception encountered running PythonOp function: <type 'exceptions.AttributeError'>: 'caffe2.python.caffe2_pybind11_state_gpu.Workspace' object has no attribute 'shape'

[...]

PR #1308, opened 2017-10-09
Pybind v2.2.1

slayton58:
Bumps the pybind version from v1.8.1 to v2.2.1, resolving all compile & runtime issues that arose.

Upgrades to the API used https://github.com/pybind/pybind11/blob/master/docs/upgrade.rst as the point of reference.

This also solves a long-standing bug we had, where a type would spontaneously and intermittently change in the C++ -> Python boundary.

\cc @Yangqing

### DIFF
--- a/caffe2/python/pybind_state.cc
+++ b/caffe2/python/pybind_state.cc
@@ -241,8 +241,6 @@ PythonOpBase::PythonOpBase(
   if (!pickled.empty()) {
     py::gil_scoped_acquire g;
     try {
-      //auto pickle =
-      //    py::object(PyImport_ImportModule("pickle"), /* borrowed */ false);
       auto pickle =
           py::reinterpret_steal<py::object>(PyImport_ImportModule("pickle"));
       CAFFE_ENFORCE(pickle);
--- a/caffe2/python/pybind_state.cc
+++ b/caffe2/python/pybind_state.cc
@@ -241,6 +241,8 @@ PythonOpBase::PythonOpBase(
   if (!pickled.empty()) {
     py::gil_scoped_acquire g;
     try {
+      //auto pickle =
+      //    py::object(PyImport_ImportModule("pickle"), /* borrowed */ false);
       auto pickle =
           py::reinterpret_steal<py::object>(PyImport_ImportModule("pickle"));
       CAFFE_ENFORCE(pickle);


caffe2/CMakeLists.txt
```diff
@@ -163,6 +163,7 @@ endif()
 if (BUILD_PYTHON)
   # ---[ Python.
   add_library(caffe2_pybind11_state MODULE ${Caffe2_CPU_PYTHON_SRCS})
+  set_target_properties(caffe2_pybind11_state PROPERTIES COMPILE_FLAGS "-fvisibility=hidden")
```
Yangqing: Hmm this one is a little bit concerning as Caffe2 explicitly relies on global visibility to get things properly registered. Do you know which thing is creating offending visibilities?

slayton58: It's taken from [here](https://github.com/pybind/pybind11/blob/master/docs/upgrade.rst#stricter-enforcement-of-hidden-symbol-visibility-for-pybind11-modules) which suggests setting that flag, and it resolves some compile-time warnings.

Note that this does only apply to the pybind wrapper libs, not anything else, so the general C2 need for visibility shouldn't be affected (and this runs all our tests without issue)

lukeyeager:
> This also solves a long-standing bug we had

See https://github.com/caffe2/caffe2/issues/793

cbalint13:
@slayton58 ,

 How to deal if someone wants external pybind (not compiled with ```-fvisibility=hidden```) ? E.g. one wants relay on distro's own pybind.

slayton58:
Pybind11 is used as a header-only library, so people would be compiling based on the distro-provided headers, and `-fvisibility=hidden` should still be used.

caffe2/python/pybind_state.cc
```diff
@@ -240,8 +240,10 @@ PythonOpBase::PythonOpBase(
   if (!pickled.empty()) {
     py::gil_scoped_acquire g;
     try {
+      //auto pickle =
+      //    py::object(PyImport_ImportModule("pickle"), /* borrowed */ false);
```
pietern: Commented code

caffe2/python/pybind_state.cc
```diff
@@ -401,15 +403,16 @@ void addObjectMethods(py::module& m) {
           "tensor",
           [](Blob* blob) {
             auto t = blob->GetMutable<TensorCPU>();
-            return py::cast(t, py::return_value_policy::reference_internal);
-          })
+            return py::cast(t); //, py::return_value_policy::reference_internal);
```
pietern: Commented code

caffe2/python/pybind_state.cc
```diff
@@ -497,23 +500,25 @@ void addObjectMethods(py::module& m) {
             for (const auto& name : self->Nets()) {
               LOG(INFO) << "name: " << name;
               nets[name] = py::cast(
-                  self->GetNet(name),
-                  py::return_value_policy::reference_internal);
+                  self->GetNet(name)); //,
+                  // py::return_value_policy::reference_internal);
```
pietern: Commented code

pietern (commented):
There's a couple of instances where commented out code is lingering. Looks good to me otherwise. Thanks for working on this @slayton58!

slayton58:
@pietern Just pushed an update with those comments (and more) removed

Yangqing:
@soumith since you did https://github.com/caffe2/caffe2/commit/3f9a93da285b86b4cd43210cae26b7729bcbc8e9 , do you mind taking a look at this as well?

soumith:
everything in this PR is now in https://github.com/caffe2/caffe2/commit/3f9a93da285b86b4cd43210cae26b7729bcbc8e9 except for updating the `git submodule` of pybind11 to latest.

@slayton58 sorry for subsuming your PR, I had to do the update in-sync with some other internal changes.

soumith:
@Yangqing i didn't actually know how to update the pybind git submodule of c2 with our internal workflow, so i didn't do that. once you guys update that, travis will get back to passing status too

slayton58:
@Yangqing I can rebase this to just encompass the updated pybind commit, but that seems like it'd be a quicker fix for you (or someone else) to do from your internal tools.

If there are no objections I'll close this out due to https://github.com/caffe2/caffe2/commit/3f9a93da285b86b4cd43210cae26b7729bcbc8e9

PR #1316, opened 2017-10-10
Fix "No handlers could be found for logger"

lukeyeager:


carljm:
This change is not a good idea. Configuring logging should be handled by the top-level script/application, never as a side effect of importing some library. This change means that any application importing caffe2 and then calling `basicConfig` itself will now have its config ignored, since all calls to `basicConfig` after the first one are ignored.

dzhulgakov:
I don't think it's a good ideas either. It actually breaks logging in some internal applications at FB.

Library shouldn't configure logging (http://docs.python-guide.org/en/latest/writing/logging/)

What was the main issue that prompted this diff? I'm curious whether we can just add NullHandler to suppress warnings

In the meantime I'll revert this diff.

lukeyeager:
I was just tired of the warnings from `data_parallel_model`.

dzhulgakov:
The maybe add the main logger in the top-level training script?

PR #1317, opened 2017-10-10
TravisCI: update to cuda9

lukeyeager:


Yangqing:
Awesome :) Since cuda8 is still probably going to be popular, do we want to keep a dual cuda8-cuda9 build?

(I am asking mainly because conda uses cuda 8 still: https://anaconda.org/anaconda/cudatoolkit)

lukeyeager:
I guess that's up to you? There are downsides any way you go:

* Both - TravisCI already can't keep up with the current load
* CUDA 9 only - it sounds like you can't really recommend this yet to users who like anaconda
  * I don't know much about anaconda - perhaps they're planning an update soon?
* CUDA 8 only - no real downside, it's just kinda lame 😉

Yangqing:
Sounds much reasonable. I hope the world goes to cuda9 soon as well - cc @pietern for a double check as Pieter knows more about cuda build infra than I do.

pietern:
@lukeyeager Update: we're looking into enabling bigger build matrices

pietern:
Closing this one, since we're now building/testing both CUDA 8 and CUDA 9 on Jenkins.

PR #1318, opened 2017-10-10
Tutorial Fix

nlaptev:
Made it clear that opencv is required for ImageInput which is a common question on github & fixed a bug in the multigpu tutorial.


Yangqing:
Thanks for the fix!

PR #1328, opened 2017-10-11
Update protobuf detection

pietern:
The scripts/build_local.sh script would always build protoc from the
third_party protobuf tree and override the PROTOBUF_PROTOC_EXECUTABLE
CMake variable. This variable is used by the protobuf CMake files, so
it doesn't let us detect whether the protoc was specified by the user
or by the protobuf CMake files (e.g. an existing installation). This
in turn led to a problem where system installed headers would be
picked up while using protoc built from third_party. This only works
if the system installed version matches the version included in the
Caffe2 tree. Therefore, this commit changes the variable to specify a
custom protoc executable to CAFFE2_CUSTOM_PROTOC_EXECUTABLE, and
forces the use of the bundled libprotobuf when it is specified.

The result is that we now EITHER specify a custom protoc (as required
for cross-compilation where protoc must be compiled for the host and
libprotobuf for the target architecture) and use libprotobuf from the
Caffe2 tree, OR use system protobuf.

If system protobuf cannot be found, we fall back to building protoc
and libprotobuf in tree and packaging it as part of the Caffe2 build
artifacts.


pietern:
Tested the following configurations:

Without system protobuf:
* `scripts/build_local.sh` --> $<TARGET_FILE:protoc>
* `USE_HOST_PROTOC=1 scripts/build_local.sh` --> build_host_protoc/bin/protoc

With system protobuf:
* `scripts/build_local.sh` --> /usr/bin/protoc
* `USE_HOST_PROTOC=1 scripts/build_local.sh` --> build_host_protoc/bin/protoc

Yangqing (approved):
Looks perfect!

LMerCy:
why "fatal eror cc1189 error : this file wa generated by a newer version of protoc" still happens .I have installed protobuff by"python -m pip install protobuff==3.4" 
I also have tried uninstalled protobuff, it still goes wrong with the same error
And this only happens when i modify build_windows.bat "-DUSE_LMDB=ON ", if i choose  "-DUSE_LMDB=OFF", than i can build caffe2 success.

PR #1342, opened 2017-10-14
[fix] compilation error in pybind and perf_observer

bwasti:
Hot fix for the open source build

### DIFF
--- a/caffe2/python/CMakeLists.txt
+++ b/caffe2/python/CMakeLists.txt
@@ -2,6 +2,7 @@
 set(Caffe2_CPU_PYTHON_SRCS
     "/pybind_state.cc"
     "/pybind_state_mkl.cc"
+    "/../observers/time_observer.cc"
 )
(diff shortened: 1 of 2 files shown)


caffe2/python/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@
 set(Caffe2_CPU_PYTHON_SRCS
     "/pybind_state.cc"
     "/pybind_state_mkl.cc"
+    "/../observers/time_observer.cc"
```
Yangqing, line 5: Hmm... why is this one in the python srcs and not main srcs? Seems that it is not python specific?

bwasti: I guess the thought was not to pull in observers unless they are explicitly compiled in, whereas python exposes the interface to this one.

I am moving this change into an internal diff to test the implications of binary size when adding it to main sources

PR #1354, opened 2017-10-18
CMake changes to make Caffe2 more friendly for dependent libraries

Yangqing:
This introduces a few things:

- It enables us to create Caffe2Config.cmake that can be used down the road for building dependent libraries, so they do not need to explicitly write FindCaffe2.cmake.
- The config file will automatically figure out transitive dependency of Caffe2 as well as compiler flags.
- This diff also disables the RPATH setting since it is kind of a mess right now. In principle, we should figure out a clearer rpath setting following the typical rpath setting choices (https://cmake.org/Wiki/CMake_RPATH_handling) - I can send a follow up PR to clean this up.
- Minor: removed old gflags ang glog files.


CMakeLists.txt
```diff
@@ -140,3 +139,44 @@ if(NOT TARGET uninstall)
       COMMAND ${CMAKE_COMMAND} -P
       ${CMAKE_CURRENT_BINARY_DIR}/cmake_uninstall.cmake)
 endif()
+
+# ---[ Make configuration files for cmake to allow dependent libraries
+# easier access to Caffe2.
+
+# Specific interface link libraries for caffe2
+if ((NOT USE_GLOG) OR (NOT USE_GFLAGS) OR CAFFE2_USE_CUSTOM_PROTOBUF)
```
pietern, line 147: Not entirely true. If CAFFE2_USE_CUSTOM_PROTOBUF then Caffe2 will link libprotobuf statically, so no need for transitive dependency of libraries. However caffe2_pb.h will include google/protobuf/... so that will be a problem unless we also install those (probably not a good idea...)

pietern:
Nice!

pietern:
After this is merged we can have a caffe2-cpp project that is a single C++ file and a single CMake file that take a dependency on Caffe2, for testing that this doesn't break.

CMakeLists.txt
Yangqing: Yeah, basically if any of these is met, the exported config file is going to be buggy and I haven't found a good solution for that... Currently marking it as a warning.

Yangqing:
@pietern yep, I am using caffe2_bhtsne to do exactly this, in fact :)

This also allows us to deal with optional dependencies, such as lmdb and rocksdb, instead of having to maintaining an evergrowing USE_SOMETHING option list.

Yangqing:
All tests pass now - PTAL.

PR #1358, opened 2017-10-20
Use own benchmark and not any system pre-built ones:

Yangqing:
(1) use the cmake files of the corresponding libs
(2) allow static linkage of gtest and gbenchmark.
(3) Helps removing the temp solution in #1112 

We are yet to disable the installation of the benchmark library, and I have an open pull request at https://github.com/google/benchmark/pull/463 - once it is merged I will do submodule update.

cc @lukeyeager @pietern who had this issue before - hopefully this makes the solution cleaner.


lukeyeager:
:+1:

https://github.com/google/benchmark/pull/463 is a nice solution.

pietern (approved):
Nice!

PR #1376, opened 2017-10-24
Add cuDNN handles to CUDAContext

slayton58:
Add CUDAContext::cudnn_handle() for easier integration of single
cudnn routines into operators without requiring the weight
of CuDNNWrapper or similar, or needing to spin out a separate CuDNN*Op
version of an operator.

It was necessary to split out the cuDNN wrapper code from the base cuDNN helpers in order to resolve a circular dependency between context_gpu.h and common_cudnn.h when handles and cuDNN `#define` were added.


caffe2/core/context_gpu.h
```diff
@@ -104,10 +119,16 @@ class ThreadLocalCUDAObjects {
           CUDA_CHECK(cudaStreamDestroy(stream));
         }
       }
+      for (auto& handle : cudnn_handles_[i]) {
+        if (handle) {
+          CUDNN_CHECK(cudnnDestroy(handle));
```
ilia-cher, line 124: should we do this before destroying stream?

slayton58: Actually this is fine -- `cudnnDestroy` doesn't touch a user-passed stream, so it doesn't matter if it exists or not at this point.

caffe2/core/cudnn_wrappers.h
```diff
@@ -0,0 +1,191 @@
...
+   * Creates a cudnn wrapper associated with a CUDAContext object. Note that
+   * the CUDAContext object should outlive the CuDNNWrapper.
+   */
+  explicit CuDNNWrapper(CUDAContext* context) : context_(context) {}
+
+  /**
+   * Returns the inline cudnn handle that executes on the current
+   * thread's cuda_stream.
+   */
+  cudnnHandle_t& inline_cudnn_handle() {
```
akyrola: I might be slightly confused, but doesn't this function get deprecated with this change, because one can grab the handle directly from cudaContext? And shouldn't this call context_->cudnn_handle() instead of creating new handles?

slayton58: That would be desirable, yes.

I intended this PR as a minimal set of changes, without affecting deeper existing APIs -- I can certainly add that change if you like

akyrola:
Nice, importing it to our system

akyrola:
This is causing some build errors in fb-specific operators because they refer to CUDNNWrapper. How about importing cudnn_wrapper.h from common_cudnn.h?

slayton58:
That causes compile errors -- it introduces a circular dependency when `common_cudnn.h` is included from `context_gpu.h`

akyrola:
Shoot… Ok, I will change the fb-only op.

PR #1378, opened 2017-10-25
Making a module option for Caffe2

Yangqing:
This will help releasing models that are using Caffe2 but have their own operator implementations and extensions. More detailed docs to arrive later. Let's see what contbuild says.


Yangqing:
cc @rbgirshick FYI - Ross, this can help releasing custom ops.

caffe2/core/module.cc
```diff
@@ -0,0 +1,122 @@
+#include "caffe2/core/logging.h"
```
galv, line 17: Should this have Facebook license boiler plate here?

caffe2/core/module.cc
```diff
@@ -0,0 +1,122 @@
...
+    CAFFE_ENFORCE(handle != nullptr,
+      "Cannot load module ",
+      name,
+      " (with given filename ",
+      filename,
+      "), are you sure it is correct?");
+  } else {
+    string inferred_name = string("lib") + name + ".so";
+    handle = dlopen(
+        inferred_name.c_str(), RTLD_NOW | RTLD_GLOBAL);
```
galv, line 89: Is there a particular reason why this doesn't also use RTLD_NODELETE as its mode?

Yangqing: Android does not seem to support it. I actually placed RTLD_NODELETE but then had to remove it.

In practice we would probably never dl close a module so this is a minor point, but if issues emerge we can macro define it in non-Android cases.

caffe2/core/module.cc
Yangqing: Ah yeah, should be.

PR #1393, opened 2017-10-28
Fix bug introduced in a recent commit

Yangqing:
This is introduced in https://github.com/caffe2/caffe2/commit/8539a1e78b7fd8aa92b08a8f24b2b0f7978bf5a5 - vector<float> should not be used in Tensor shape inference.


Yangqing:
@akyrola since you are the reviewer for the original commit, would you mind taking a look?

akyrola:
oops

PR #1406, opened 2017-10-30
Fix Batched Matmul test accuracy

slayton58:
Datatypes was being handled badly in reference check, causing sporadic fails in CI. All batched mat-mul with fp16 data is performed as pseudo-fp16, with all math in fp32. Adjusted the reference implementation to reflect this.

Adjusted the gradient check threshold to the best I could get to consistently pass.


pietern:
Thanks @slayton58!

PR #1409, opened 2017-10-31
Support overriding CuDNN in CUDA directory with custom CuDNN

fpoms:
Since the current cmake configuration adds the CUDA include path before the CuDNN include path, if a CuDNN header exists in the CUDA include path it will always override any CuDNN header provided by a custom CuDNN location. This can be solved by adding the CuDNN include path before the CUDA include path in cmake.


apoms:
Please ignore, functionality works as expected.

PR #1410, opened 2017-10-31
Correct the annotation about iPhoneVersion 

103yiran:
iPhone 6 and up --> iPhone 7 and up


Maratyszcza:
Thank you for the effort, but there is no error in the annotation. iPhone 6 has reports model names `iPhone7,2`, `iPhone7,1`.

PR #1413, opened 2017-10-31
Add support for doing broadcast with single elem dimensions at both ends

bddppq:


caffe2/operators/elementwise_op.cu
```diff
@@ -272,20 +272,30 @@ bool SumReduceLikeOp<CUDAContext>::DoRunWithType() {
         "If you are doing ReduceSumLike, input1 should have "
         "a smaller number of dimensions.");
     const int axis = (axis_ == -1 ? A.ndim() - B.ndim() : axis_);
-    CAFFE_ENFORCE(
-        axis >= 0 && axis < A.ndim(),
-        "ReduceSum axis should be in the range of the number "
-        "of dimensions of the first input.");
+    CAFFE_ENFORCE(axis >= 0 && axis_ <= A.ndim() - B.ndim(),
```
dzhulgakov: hm, isn't it copypasta which we can abstract out?

caffe2/operators/elementwise_op.cc
```diff
@@ -161,26 +161,35 @@ bool SumReduceLikeOp<CPUContext>::DoRunWithType() {
...
+    while (b_dim_start < B.ndim() && B.dim(b_dim_start) == 1) {
+      ++b_dim_start;
+    }
+    int b_dim_end = B.ndim() - 1;
+    while (b_dim_end >= b_dim_start && B.dim(b_dim_end) == 1) {
+      --b_dim_end;
+    }
     size_t pre = 1, n = 1, post = 1;
-    for (int i = 0; i < axis; ++i) {
+    for (int i = 0; i < axis + b_dim_start; ++i) {
```
dzhulgakov: this really makes 'axis' argument ambiguous - it has different meaning depending on the number of leading 1 dimensions and is potentially dangerous (what if you batch size is 1?)

I think the reasonable alternative is to allow 1-based expansion logic only if `axis` is not specified and just follow numpy semantics.

dzhulgakov (changes requested):

caffe2/operators/elementwise_op.cc
bddppq: Hmm I feel it's orthogonal though. axis is necessary for supporting expanding in rhs, it indicates starting from which axis of the first Input to align B's dimension. While all the leading single elem dimensions of B are skipped, same amount of dimensions of A (starting from axis) are also skipped. So this diff hasn't changed the semantics of all previous supported cases.
`
A dims: 2 1 3 4

B dims: 1 3 4

axis: 1
`
behaves exactly like before.

dzhulgakov: Oh, my bad I got totally confused

dzhulgakov (approved):

PR #1416, opened 2017-11-01
Pass CUDA_NVCC_EXECUTABLE to NCCL build

pietern:
If this variable is set to a ccache symlink then the NCCL build will
also use the cache. The NCCL build is the slowest component of a cached
build without this change


pietern:
@caffe2bot retest this please

PR #1420, opened 2017-11-02
CMake: remove unneeded dependency with OpenBLAS

### LINKED ISSUE #763: Build with OpenBlas tries to link with cblas (and fails if you don't have one)
Hi,

I'm trying to build caffe2 (release v0.7.0) with OpenBlas support:

```
# Configure
cmake -D CMAKE_TOOLCHAIN_FILE=${TOOLCHAIN} \
<...>
-D BLAS=OpenBLAS \
-D OpenBLAS_INCLUDE_DIR=/src/dependencies/include \
-D OpenBLAS_LIB=/src/dependencies/lib/libopenblas.so \
<...>
${SRC_PATH}
```
The build fails on the stage of shared lib libCaffe2_CPU.so creation because it tries to link with cblas.
```
<...>/bin/ld.bfd: cannot find -lcblas
collect2: error: ld returned 1 exit status
```
I don't have libcblas and I'm not sure why it is needed when I have provided OpenBlas implementation which has all the stuff.

Cblas  comes from here:
cmake/Dependencies.cmake:36
```
...
32: elseif(BLAS STREQUAL "OpenBLAS")
33:   find_package(OpenBLAS REQUIRED)
34:   include_directories(SYSTEM ${OpenBLAS_INCLUDE_DIR})
35:   list(APPEND Caffe2_DEPENDENCY_LIBS ${OpenBLAS_LIB})
36:   list(APPEND Caffe2_DEPENDENCY_LIBS cblas)
...
```
If I comment out this line, then the build will produce perfectly usable library for me.

By the way, caffe1 did not need cblas when using OpenBlas lib.

christoph-conrads:
Do not try to link against `libcblas.so` when using the OpenBLAS
back-end. This fixes #763.

I briefly checked the OpenBLAS repository and as far as I can tell, the OpenBLAS build script by build never created a library called _cblas_.


pietern:
@christoph-conrads Thanks for fixing this.

@caffe2bot test this please

PR #1434, opened 2017-11-08
Don't run CUDA tests for ops without CUDA implementation

pietern:


pietern:
@caffe2bot Thanks, but next time, please include a link...

PR #1443, opened 2017-11-10
Add /usr/local/opt/python/libexec/bin to $PATH on Mac travis

Maratyszcza:
This should Travis the build failures on Mac

### DIFF
--- a/.travis/build.sh
+++ b/.travis/build.sh
@@ -38,6 +38,7 @@ fi
 if [ "$BUILD_TESTS" = 'false' ]; then
     CMAKE_ARGS+=('-DBUILD_TEST=OFF')
 fi
+CMAKE_ARGS+=$(python $ROOT_DIR/scripts/get_python_cmake_flags.py)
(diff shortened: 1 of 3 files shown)


bddppq:
I think we also need to
- do `brew install python` (or maybe even more explicit `brew install python==2.7.14`) in install.sh
- add `$(python $ROOT_DIR/scripts/get_python_cmake_flags.py)` to CMAKE_ARGS in build.sh

Maratyszcza:
Do you suggest to always add `$(python $ROOT_DIR/scripts/get_python_cmake_flags.py)` to CMAKE_ARGS? Or only on OSX?

bddppq:
Always. That script always finds include dir (which has portable logic) and only finds libpython on darwin platform.

bddppq:
Hmm brew install doesn't support `==` syntax for specifying version.

Maratyszcza:
Oops. It already landed internally, and will be pushed to GH soon

bddppq:
Well it was broken before, we have not made it worse ;)

PR #1457, opened 2017-11-11
Fixed layout of highlighted code

arogozhnikov:
Just fixing the way code looks in the manuals.

Currently caffe2.ai it looks like this on my devices:

screenshot from 2017-11-11 17-31-10


Yangqing:
Thanks so much!

PR #1461, opened 2017-11-13
Limit this fix to apple clang only

CaoZhongZ:
Use "__apple_build_version__" macro to distinguish Apple's Clang while brew installed LLVM will compile caffe2 without trouble.


Yangqing:
Thanks so much!

PR #1462, opened 2017-11-13
Bump third_party/protobuf to v3.4.1

pietern:
This version fixes warnings when compiling on macOS.

It is generally discouraged to use the bundled version of protobuf and
rather use the system installed version. CMake, by default, will use
the system installed version. When cross compiling Caffe2 (e.g. for
mobile), it needs to use the bundled version of protobuf.


pietern:
Verified it fixes the compile warnings:

https://ci.pytorch.org/jenkins/job/caffe2-builds/job/caffe2-macos-build/4/console

Yangqing:
Hmm protobuf seems to be failing for Android.

pietern:
It has to do with some js_embed binary. Never heard of it. Hopefully it can just be disabled...

pietern:
See google/protobuf#2658 and google/protobuf#3876

Yangqing:
Proper solution is probably something like https://github.com/google/protobuf/pull/3878 but not sure when it will land...

Yangqing:
#1466 should do the bump and also address build temporarily.

PR #1463, opened 2017-11-13
Adding zstd to build

Yangqing:
This is in order for us to share compression ops to oss.


pietern:
Why enable by default? Are we using it somewhere already?

Yangqing:
Yep, in an upcoming diff. But good call - we shouldn't enable it in default.

PR #1466, opened 2017-11-14
Disable protobuf libprotoc and protoc build for cross compilation.

Yangqing:
Also bumped third_party/protobuf to v3.4.1 similar to #1462 . cc @pietern


pietern:
This doesn't include the bump yet, LGTM otherwise

PR #1467, opened 2017-11-14
Fix: template caused a compilation error

KangLin:
Fix：compilation error


Yangqing:
Thanks! This is already included in the PR #1455 that is being landed, so I'll close this one for now.

PR #1470, opened 2017-11-14
disable travis Linux host build, as we are moving to using jenkins for contbuild.

Yangqing:
I only commented them out so that if we want to run Travis in the future, we could potentially still enable them.


Yangqing:
(just testing water, will merge later when things are more full fledged)

Yangqing:
Closing as @pietern will deal / have dealt with this.

PR #1480, opened 2017-11-16
Add setup.py

bddppq:
Take environment variable CMAKE_ARGS as extra cmake flags

tested `pip install` `pip wheel` `python setup.py build` `python setup.py install`


dzhulgakov (commented):
Can you share what the options are by default if you build it like this? (either CMake output or after installation `from caffe2.python import build; build.build_options`)?

I'm curious whether we want to include MKL into build dependencies (otherwise performance would be bad).

Also, does CUDA support work? Maybe makes sense to test on gpu server to double-check.

In general I'm not an expert on setuptools, so would defer to others to take a look

bddppq:
It invokes cmake to do the actual build, so everything should work exactly the same as using plain cmake. If we want to enable MKL in some distributions, we should do it through the environment variable "CMAKE_ARGS", which will be converted to cmake flags to be passed to cmake invocation command. The default cmake flags this setup.py passes to cmake is just to prune the unusable stuffs in python (c++ executables, tests, benchmarks), and passing some python information from the python runtime to cmake so cmake doesn't need to figure that out.
I think build options will depend on the build machines (as directly invoking cmake), on my linux desktop it's
 ```
$ python -c 'from caffe2.python import build; from pprint import pprint; pprint(build.build_options)'
WARNING:root:This caffe2 python run does not have GPU support. Will run in CPU only mode.
WARNING:root:Debug message: No module named caffe2_pybind11_state_gpu
{u'BLAS': u'Eigen',
 u'BUILD_TYPE': u'Release',
 u'CXX_FLAGS': u' -Wno-deprecated -std=c++11 -O2 -fPIC -Wno-narrowing',
 u'FORCE_FALLBACK_CUDA_MPI': u'',
 u'GIT_VERSION': u'v0.8.1-688-g03808cd-dirty',
 u'HAS_MKL_DNN': u'',
 u'HAS_MKL_SGEMM_PACK': u'',
 u'PERF_WITH_AVX': u'1',
 u'PERF_WITH_AVX2': u'1',
 u'UNIQUE_LONG_TYPEMETA': u'',
 u'USE_ACCELERATE': u'',
 u'USE_ATEN': u'OFF',
 u'USE_CUDA': u'OFF',
 u'USE_EIGEN_FOR_BLAS': u'1',
 u'USE_GFLAGS': u'OFF',
 u'USE_GLOG': u'OFF',
 u'USE_GLOO': u'',
 u'USE_LITE_PROTO': u'',
 u'USE_MKL': u'',
 u'USE_MPI': u'OFF',
 u'USE_NCCL': u'OFF',
 u'USE_NNPACK': u'ON',
 u'USE_NVTX': u'',
 u'USE_OPENMP': u'OFF'}
```

File: setup.py [hunk 1]
```diff
@@ -0,0 +1,201 @@
...
+    description='Caffe2',
+    ext_modules=ext_modules,
+    cmdclass=cmdclass,
+    packages=packages,
+    install_requires=install_requires,
+    setup_requires=setup_requires,
+    tests_require=test_requires,
+    author='jiayq',
+    author_email='jiayq@fb.com',
+    url='https://github.com/caffe2/caffe2',
```
pietern: https://caffe2.ai

setup.py
```diff
@@ -0,0 +1,201 @@
...
+    git_version = subprocess.check_output(['git', 'rev-parse', 'HEAD'],
+                                          cwd=TOP_DIR).decode('ascii').strip()
+except subprocess.CalledProcessError:
+    git_version = None
+
+with open(os.path.join(TOP_DIR, 'VERSION_NUMBER')) as version_file:
+    VersionInfo = namedtuple('VersionInfo', ['version', 'git_version'])(
+        version=version_file.read().strip(),
+        git_version=git_version
+    )
```
pietern, line 47: Instead of a call to git and a VERSION_FILE, you can use just a single git command to get a descriptive version. Then the VERSION_FILE can be removed.

```
$ git describe --tags HEAD
v0.8.1-687-g08574bb
```

setup.py
```diff
@@ -0,0 +1,201 @@
...
+        py_exe = sys.executable
+        py_inc = sysconfig.get_python_inc()
+        cmake_args = shlex.split(os.environ.get('CMAKE_ARGS', ''))
+        log.info('CMAKE_ARGS: {}'.format(cmake_args))
+
+        cwd = os.getcwd()
+        os.chdir(build_dir)
+        self.compiler.spawn([
+            'cmake',
+            '-DBUILD_SHARED_LIBS=OFF',
```
pietern: I suppose this ensures you get 1 big Python extension .so. Is there a way to use shared libs?

pietern (commented):
Nice work @bddppq.

Not a big fan of calling out to cmake from here. We have the build scripts in `./scripts` which we ask folks to use and would like this to use those as well. Then there is only a single point where we deal with these invocations. Or better yet, it would also be nice to have an option to point setup.py to a build directory and have it just package it. Then we can make a Python wheel out of every build we do, and not have to special case this script for every environment.

setup.py
bddppq: We used to do that in onnx, but later change to this way because source distribution, it's not a git repo. Also it's helpful to have the version in a plain file when doing importing.

setup.py
bddppq: Yep, this will create one big giant python extension.
I started with using shared libs, but it's not easy (yep there is a way, will describe later) to get it work. The problem is at the build stage, we don't know where will the final installation dir (setuptools installation dir, not the cmake installation dir) be, so we can not use an absolute rpath. Note since we want to create wheels, we can not put these cmake steps to the install* commands. The possible way to make it work is using relative rpath, but this relative value will be different between pybind_state.so and the so in $cmake_install_dir/lib directory (imagine pybind_state.so -> libcaffe2.so -> libxxx.so, the first one is at caffe2/python directory, and the latter two are in lib/ directory), so we need to do patchelf for individual so outside of cmake.
But regardless, I think it's better to do static libs/one giant so when doing deployment, it makes the deployment story much easier.

bddppq:
@pietern It takes an environment variable CMAKE_BUILD_DIR to point to a previous existing cmake build directory, if users has done build there before, invoking cmake (and make) again should do nothing. Note we shouldn't directly package stuffs from the cmake build directory, it should be from the cmake install directory. I can change it to if user has specified the cmake install directory (via CMAKE_INSTALL_DIR), then we skip the cmake invocation.
I dont' think we should make setup.py a packaging-only script, that will make us unable to do python package source distribution.

pietern:
Good point re: source distribution.

You could consider running scripts/build_local.sh (you can pass additional CMake args that are passed through). Then we can only have one path for defaults and for example ccache configuration. Not required of course.

bddppq:
@pietern Yeah makes sense, I will change it to use the build scripts instead of directly invoking cmake.

setup.py
pietern: I see, makes sense. Perhaps we can use the git way if it's a git repo and fall back to VERSION if it doesn't work? It's nice to have both the tag + distance + sha IMO.

bddppq: We can add as many entries as we want in the VersionInfo (say git_describe=$(git describe --tags)) :-)

Update: 
Actually it's already doing "try doing git, if failed then pass". But instead we can replace the git_version value with the output of $(git describe --tags) (right now it's $(git rev-parse HEAD))

PR #1484, opened 2017-11-16
Make macOS build use ccache via CMAKE_C*_COMPILER

pietern:


pietern:
Time spent: 6:00 -> 1.30

PR #1496, opened 2017-11-20
Allow specifying cmake build directory in the build scripts

bddppq:


pietern:
LGTM! You can ignore the clang38 failure -- a fix is pending.

PR #1497, opened 2017-11-21
Use git version info as python package version

bddppq:


galv:
In the spirit of don't-repeat-yourself, shouldn't you remove VERSION_NUMBER?

More importantly, I don't think this is quite right, locally for me, `git describe --tags HEAD` reports `v0.8.1-699-g9d600253`

Your version numbers need to match a specific regex: `[N!]N(.N)*[{a|b|rc}N][.postN][.devN]`, as described [here](https://www.python.org/dev/peps/pep-0440/#version-scheme).

In my experience packaging python projects in the past, nothing will crash if you use a non-compliant version id locally. However, you will get warnings. Also, I'm not sure if this will work on pypi. In fact, I suspect it won't work.

bddppq:
Ok good point, and the current code won't work in sdist (in which there is no git). I should compute the version number as (sorry for mixing bash and python code below):

```
version = $(cat VERSION_NUMBER)
if git_is_available:
    version += '.dev{}'.format($(git rev-parse --short HEAD))
```

Regarding to the VERSION_NUMBER file, having the version number in a plain text file is easier for applications that don't invoke python command.

galv:
That looks reasonable to me.

bddppq:
Hmm actually by looking closer, my previous proposal still doesn't fulfill the standard, it should be

```
version = $(cat VERSION_NUMBER)
if git_is_available:
    version += '+git.{}'.format($(git rev-parse --short HEAD))
```

galv:
Oh I see. I assume you're referring to to the fact that the N in .devN must be a decimal number, not a hexadecimal number. (I verified this in appendix B of PEP440). Just note that you can't put a version like that onto pypi, if that's your goal. You'd have to have a "release" workflow as well that omits the plus sign and the hash, I believe.

By the way, this answer suggests a way to incorporate the number of commits (part of `git describe`'s output) since the last tag into the dev field along with the current commit's hash, if you want that:  https://stackoverflow.com/a/35522080/3469721

pietern:
With the current approach you won't be able to create release wheels without first removing the .git directory. Would be good to think about this before we need to do another release

PR #1520, opened 2017-11-26
git clone from the master branch in the docker files because branch v0.8.1 does not exist

vtomole:


burningion:
This worked for me to build Caffe2 on Docker, fixed the broken references in the v0.8.1 and v0.8.0 branches. But I'm not sure it's a permanent fix, the v0.8.0 and v0.8.1 releases are both broken because of the Cub repo references are gone that the submodules point to.

orionr:
Thank you @vtomole and @burningion for this contribution. Bringing it in for now. We might use some other Docker containers we've been creating for Jenkins as our base in the future, but the broken downstream reference makes this an immediate fix.

PR #1527, opened 2017-11-27
[WIP] Bring back python3 builds

bddppq:


pietern:
@bddppq I think we can abort this PR, since we're testing a bunch of py3 environments on Jenkins now.

Why do you need to set the Python_ADDITIONAL_VERSIONS variable btw?

bddppq:
Yes let's abort it for now.
`Python_ADDITIONAL_VERSIONS` is for instructing cmake what python versions to look for. Right now caffe2 cmake build system will only look for python2. I guess in the python3 jenkins CI you are explicitly passing `PYTHON_EXECUTABLE` to cmake?

PR #1530, opened 2017-11-28
Build Redis support on Linux

pietern:
Builds can then execute rendezvous where a shared file system is not available.


Yangqing:
@pietern one of the failure is due to an incorrect hypothesis assertion - I will send an internal diff to fix that.

pietern:
Since this is in master, this should now be in better shape, although still requires #1536 to be 100% fixed.

@caffe2bot retest this please

Issue #1441, opened 2017-11-09
resnet50_trainer.py + OpenMPI failing with unhandled type error


kramimus:
Running resnet50_trainer.py with OpenMPI (just over TCP) across 2 nodes on the master branch and hitting this error:

```
$ mpirun.openmpi -machinefile $HOME/machinefile.gpu -npernode 1 python resnet50_trainer.py --train_data null
/usr/bin/ssh: /program/caffe2/0.8.2dev-9.0-7.0-py27/env/lib/libcrypto.so.1.0.0: no version information available (required by /usr/bin/ssh)
/usr/bin/ssh: /program/caffe2/0.8.2dev-9.0-7.0-py27/env/lib/libcrypto.so.1.0.0: no version information available (required by /usr/bin/ssh)
Ignoring @/caffe2/caffe2/contrib/nccl:nccl_ops as it is not a valid file.
Ignoring @/caffe2/caffe2/contrib/gloo:gloo_ops as it is not a valid file.
Ignoring @/caffe2/caffe2/contrib/gloo:gloo_ops_gpu as it is not a valid file.
Ignoring @/caffe2/caffe2/distributed:file_store_handler_ops as it is not a valid file.
Ignoring @/caffe2/caffe2/distributed:redis_store_handler_ops as it is not a valid file.
Caffe2 build configuration: 
  BLAS                      : Eigen
  BUILD_TYPE                : Release
  CXX_FLAGS                 :  -fopenmp -Wno-deprecated -std=c++11 -O2 -fPIC -Wno-narrowing
  FORCE_FALLBACK_CUDA_MPI   : 1
  GIT_VERSION               : v0.8.1-600-gbd025ea
  HAS_MKL_DNN               : 
  HAS_MKL_SGEMM_PACK        : 
  PERF_WITH_AVX             : 1
  PERF_WITH_AVX2            : 1
  UNIQUE_LONG_TYPEMETA      : 
  USE_ACCELERATE            : 
  USE_ATEN                  : OFF
  USE_CUDA                  : ON
  USE_EIGEN_FOR_BLAS        : 1
  USE_GFLAGS                : ON
  USE_GLOG                  : OFF
  USE_GLOO                  : 
  USE_LITE_PROTO            : 
  USE_MKL                   : 
  USE_MPI                   : ON
  USE_NCCL                  : ON
  USE_NNPACK                : ON
  USE_NVTX                  : 
  USE_OPENMP                : ON
[...]

PR #1537, opened 2017-11-29
Handle broadcast of ints on CPU side

pietern:


pietern:
This should fix #1441

PR #1539, opened 2017-11-30
Fix linkage error with external nccl

dzhulgakov:
If nccl is installed in the system, binary linkage was failing as we don't pass dependent libs to binary targets (we do it for tests and rest of the targets)


pietern:
This should have shown up in the new images that include a system install of NCCL, but haven't seen it.

Where did you find this issue?

@caffe2bot retest this please

pietern:
For example in [this build](https://ci.pytorch.org/jenkins/job/caffe2-builds/job/py2-cuda9.0-cudnn7-ubuntu16.04-build/53/consoleFull) you see it picks up libnccl.so from a system directory and compiles the contrib/nccl files just fine. I'm curious what is different here vs in the situation where you observed this.

orionr:

Issue #1517, opened 2017-11-23
bi-directional encoder in seq2seq translation mode example fails to get created


fmetze:
I am getting the following error when trying to build a bi-directional encoder for the seq2seq translation example. If I am not using the bi-directional encoder, things seem to work, I can turn on attention, train on 2 GPUs, etc - and I am getting recognizable results, but I'd like to try the bidirectional encoder.

This is on an Azure Ubuntu 16.04 Deep Learning VM, with CUDNN 6.0, using the current (as of now) master branch.

metze@DeepUbuntu:~/caffe2-install/caffe2/python/examples$ python2 ../models/seq2seq/train.py --source-corpus /home/metze/training/news-commentary-v8.es-en.es --target-corpus /home/metze/training/news-commentary-v8.es-en.en --source-corpus-eval /home/metze/dev/newstest2009.es --target-corpus-eval /home/metze/dev/newstest2009.en --use-bidirectional-encoder
Ignoring @/caffe2/caffe2/contrib/nccl:nccl_ops as it is not a valid file.
Ignoring @/caffe2/caffe2/contrib/gloo:gloo_ops as it is not a valid file.
Ignoring @/caffe2/caffe2/contrib/gloo:gloo_ops_gpu as it is not a valid file.
Source vocab size 7211
Target vocab size 6877
Number of training batches 5436
Number of eval batches 79
Initializing Seq2SeqModelCaffe2 from scratch: Start
Traceback (most recent call last):
  File "../models/seq2seq/train.py", line 784, in <module>
    main()
  File "../models/seq2seq/train.py", line 779, in main
    max_gradient_norm=args.max_gradient_norm,
  File "../models/seq2seq/train.py", line 667, in run_seq2seq_model
    model_obj.initialize_from_scratch()
  File "../models/seq2seq/train.py", line 518, in initialize_from_scratch
    self._build_model(init_params=True)
  File "../models/seq2seq/train.py", line 129, in _build_model
    loss_blobs = self.model_build_fun(model)
  File "../models/seq2seq/train.py", line 251, in model_build_fun
    num_gpus=self.num_gpus,
  File "/home/metze/caffe2-install/caffe2/python/models/seq2seq/seq2seq_util.py", line 314, in build_embedding_encoder
    scope=get_layer_scope(scope, 'encoder', i),
  File "/home/metze/caffe2-install/caffe2/python/models/seq2seq/seq2seq_util.py", line 161, in rnn_bidirectional_layer
[...]

PR #1541, opened 2017-11-30
make sure we do integer division

fmetze:
when building a bidirectional encoder, this may end up giving us a float, rather than an integer
fixes issue #1517


pietern:
Thanks for your contribution. Is there any way you can add a test case for this?

fmetze:
sorry - right now i do not have the time to figure out how to do this.

orionr:

PR #1544, opened 2017-11-30
Update tutorial of AICamera demo

thundertrick:
There is no submodule in this project now. And `app/libs/caffe2`does not exist either. The project can be compiled directly. No need to update submodules.


orionr:
Thanks @thundertrick for your changes and sorry for the slow merge. Taking care of it now.

PR #1549, opened 2017-11-30
should have submitted this together with #1541

fmetze:


pietern:
Hi, thanks for your contrib. You can add this to #1541 instead of having a separate PR for it. Please add an additional commit to the branch you used for #1541 to include it there.

fmetze:
sorry - right now i do not have the time to figure out how to do this. i tried, and couldn’t do it quickly, so i submitted this separately.

PR #1566, opened 2017-12-04
Always check cuDNN support in test_convolution_gradients

pietern:
Regardless of device checker/gradient checker we cannot run a
backwards pass with cuDNN when NHWC is used.


pietern:
@Yangqing I'm landing this to unbreak some tests.

PR #1571, opened 2017-12-05
Stod

Yangqing:
This is in order for Android to pass - Android support for string related functions is quite limited.


pietern:
https://ci.pytorch.org/jenkins/job/caffe2-builds/job/py2-android-ubuntu16.04-trigger-build/8/

Triggered manually since the Android test is not triggered automatically at the moment.

PR #1576, opened 2017-12-05
[caffe2][script] Implement typed numeric literals

jamesr66a:
1 is an int32
1LL is an int64
1f is a float

Still need:
Parsing out numbers such as 1.0 as integer. 1.0f should work, though


pietern:
FYI pr/py3-gcc5-ubuntu16.04-test failure is benign.

PR #1579, opened 2017-12-06
Adding anaconda dockerfiles

pjh5:
These are not expected to work correctly right now. They will be used as repros of build problems when anaconda exists alongside system libraries.


pjh5:
A PR request will be made from another branch

PR #1581, opened 2017-12-06
[caffe2][script] Misc. fixes and improvements

zdevito:
* condition if
* True/False literals
* and, or, not
* 0-output expressions, like print
* _ is given a fresh name
* x.foo(...) is desugared to foo(x,...)
* +=, *=

### DIFF
--- a/caffe2/contrib/script/compiler.cc
+++ b/caffe2/contrib/script/compiler.cc
@@ -24,6 +24,12 @@ struct DefCompiler {
     }
     emitStatements(def.statements());
   }
+  void emitExpressionStatement(TreeRef stmt) {
+    // expression with no used outputs
+    auto r = emit(stmt);
+    // remove the implicit single output
+    r->clear_output();
+  }
   void emitStatements(const ListView<TreeRef>& statements) {
     for (auto stmt : statements) {
       switch (stmt->kind()) {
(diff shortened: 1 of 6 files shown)


caffe2/contrib/script/compiler.cc
```diff
@@ -251,6 +276,9 @@ struct DefCompiler {
     } else if (type_ident == "LL") {
       dtype->set_i(TensorProto_DataType_INT64);
       value->set_i(v);
+    } else if (type_ident == "b") {
```
jamesr66a: Is this only an implementation detail in the compiler or in the language can I pass, for ex, 1234b and that will emit a ConstantFill(type=bool, value=1i)?

jamesr66a (approved):
LGTM! Thanks!

caffe2/contrib/script/compiler.cc
zdevito: It is an implementation detail, True generates (TK_CONST 1 b). I will probably remove this tomorrow in favor of making it (TK_CONST "True" b), (TK_CONST "1.0" "f"), (TK_CONST "34" "I"), etc.

pietern:
@zdevito The android failure is real in this case. Just like stoi and stod it needs to use the caffe2 wrapper for those as they don't exist in Android's C++11.

PR #1587, opened 2017-12-07
Parameterize jenkins user uid/gid in docker build

pietern:


pietern:
@caffe2bot retest this please

PR #1589, opened 2017-12-07
[caffe2][script] Builtins for {zeros,ones}{,_like}

jamesr66a:


caffe2/contrib/script/compiler.cc
```diff
@@ -344,6 +348,76 @@ struct DefCompiler {
       env; // map from name in Def to name in NetDef
   std::vector<NetDef*> net_def_stack;
   int next_fresh = 0;
+
+ private:
+  OperatorDef* emitFillOp(const Apply& apply) {
+    auto builtin_type = apply.name().name();
+    auto values = getValues(apply.inputs());
+    if (values.size() > 1) {
+      throw std::runtime_error(
```
zdevito: These errors should be

```
throw ErrorReport(apply) << "the " << "error" << "message";
```

Otherwise you don't get context info.

zdevito (approved):
Looks good, just switch the error reporting to ErrorReport.

PR #1595, opened 2017-12-08
Don't use CMake generator expression for in-tree protoc build

pietern:
This fixes the in-tree protoc build on CentOS 7 (that ships with super old protobuf version).


pietern:
Support for generator expressions in the DEPENDS clause was added in Kitware/CMake@bbffccca42d4f209220e833e1a86e735a5c83339

Issue #805, opened 2017-06-15
Compile error with cpu-only build - gcc7


dbermond:
I'm getting this compile error when trying to build caffe2 without gpu support (cpu-only build):
```
[ 64%] Building CXX object caffe2/CMakeFiles/Caffe2_CPU.dir/operators/remove_data_blocks_op.cc.o
In file included from /home/user/abs/caffe2-cpu-git/src/caffe2-cpu-git/caffe2/operators/recurrent_network_op.cc:1:0:
/home/user/abs/caffe2-cpu-git/src/caffe2-cpu-git/caffe2/operators/recurrent_network_op.h: In member function ‘bool caffe2::AccumulateInputGradientOp<T, Context>::RunOnDevice()’:
/home/user/abs/caffe2-cpu-git/src/caffe2-cpu-git/caffe2/operators/recurrent_network_op.h:761:50: error: cannot call member function ‘const T& caffe2::OperatorBase::Input(int) [with T = caffe2::Tensor<caffe2::CPUContext>]’ without object
         OperatorBase::Input<Tensor<CPUContext>>(0).template data<int32_t>()[0];
                                                  ^
make[2]: *** [caffe2/CMakeFiles/Caffe2_CPU.dir/build.make:2895: caffe2/CMakeFiles/Caffe2_CPU.dir/operators/recurrent_network_op.cc.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/Makefile2:427: caffe2/CMakeFiles/Caffe2_CPU.dir/all] Error 2
make: *** [Makefile:141: all] Error 2
```

System Information:
**OS:** Arch Linux x86_64
**Compiler:** gcc 7.1.1
**Caffe2:** git master (currently at 0.7.0.r434.g4c76b747)

It builds fine when enabling gpu support and using gcc 5.4.0 (due to cuda).

Steps to reproduce and complete output are attached.

[caffe2-steps-and-output.txt](https://github.com/caffe2/caffe2/files/1076418/caffe2-steps-and-output.txt)

cuepi:
This is happening to me in a Mac too. 

System Information:
**OS:** Mac OSX
**Compiler:** gcc-7 (Homebrew GCC 7.1.0 --without-multilib) 7.1.0
**Caffe2:** git master (a57251bb0ad5339cb54cecd66ed615535f7e334b)

[...]

PR #1602, opened 2017-12-11
Fix compilation on GCC 7

pietern:
Thanks to @BrettRyland for the initial fix in #805.


pietern:
Manually triggered GCC 7 build [here](https://ci.pytorch.org/jenkins/job/caffe2-builds/job/py2-gcc7-ubuntu16.04-trigger-build/6/).

Adding GCC 7 to the default set of builds to run after this is merged.

pietern:
Everything looks good.

PR #1603, opened 2017-12-11
Adding Ubuntu Anaconda environments

pjh5:


pietern:
If you make anaconda a step in the Ubuntu builders then you don't have to dup the base install of packages and it can be shared. Also it makes the CUDA + anaconda builder work out of the box.

docker/jenkins/common/install_anaconda.sh
```diff
@@ -0,0 +1,88 @@
...
+apt-get install -y --no-install-recommends \
+      build-essential \
+      cmake \
+      git \
+      libgoogle-glog-dev \
+      libprotobuf-dev \
+      protobuf-compiler \
+      python-dev \
+      python-pip
+pip install numpy protobuf
```
pietern: These are all installed by previous build steps (see install_base.sh) so this can be removed.

docker/jenkins/common/install_anaconda.sh
```diff
@@ -0,0 +1,88 @@
...
+      libgtest-dev \
+      libiomp-dev \
+      libleveldb-dev \
+      liblmdb-dev \
+      libopencv-dev \
+      libopenmpi-dev \
+      libsnappy-dev \
+      openmpi-bin \
+      openmpi-doc \
+      python-pydot
```
pietern: Some of these are installer earlier as well.

Issue #1547, opened 2017-11-30
test_remove_padding is flaky


pietern:
See: https://ci.pytorch.org/jenkins/job/caffe2-builds/job/caffe2-linux-trusty-cuda8-cudnn6-test/247

pietern:
Tracking in #1608.

PR #1608, opened 2017-12-12
Revert "Skip remove_padding test because it is flaky"

pietern:
This reverts commit 787102f017e7692efb43daae5f0e7079cf5df680.

Also see #1547.

Creating the PR to kick off tests and see it fail.


pietern:
@caffe2bot retest this please

PR #1610, opened 2017-12-12
Add readme for docker/jenkins directory

pietern:
This also removes the `bin/{build.sh,test.sh}` scripts that are now
located in `.jenkins/{build.sh,test.sh}`. The rationale for this is
that these scripts don't care about Docker specifically and are also
run for, for example, macOS builds.


pjh5 (approved):

pietern:
@pjh5 Thanks!

PR #1627, opened 2017-12-15
Fix for installation of exported targets

Yangqing:
cc @houseroad @pietern

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,6 @@
 cmake_minimum_required(VERSION 2.8.12 FATAL_ERROR)
+#cmake_policy(SET CMP0022 NEW)
+#cmake_policy(SET CMP0023 NEW)
 
 # ---[ Project and semantic versioning.
 project(Caffe2 CXX C)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -139,27 +141,38 @@ endif()
 
 # ---[ Make configuration files for cmake to allow dependent libraries
 # easier access to Caffe2.
+# Since one would use glog, gflags and protobuf functionalities, any
```
pietern: oxford comma

CMakeLists.txt
```diff
@@ -139,27 +141,38 @@ endif()
...
+if (USE_GLOG)
+  list(APPEND CAFFE2_INTERFACE_LIBS glog)
+endif()
+if (USE_GFLAGS)
+  list(APPEND CAFFE2_INTERFACE_LIBS gflags)
+endif()
+if (NOT CAFFE2_USE_CUSTOM_PROTOBUF)
+  list(APPEND CAFFE2_INTERFACE_LIBS protobuf)
+endif()
+# message(STATUS "Caffe2 interface libs: " ${CAFFE2_INTERFACE_LIBS})
```
pietern: comment

pietern:
@Yangqing Nice. The only difference is that for BUILD_SHARED=OFF you now don't declare the generate CMake files right? The CAFFE2_INTERFACE_LIBS stuff is sugar?

Yangqing:
@pietern Yep. Basically for OSS I think we should use shared library as the standard deployment path, as static library + imported dependency is literally something unmanageable :)

For Android use cases, this means that users have to hand pick what static libs they want to link against, but that is probably the best because it consciously makes users aware of binary size.

CMakeLists.txt
Yangqing: done

CMakeLists.txt
Yangqing: done

pietern (approved):
There's the cmake policy at the start, do you want to keep those around?

Yangqing:
Yep, the policy is there so we don't accidentally mix old style and new style target_link_libraries.

PR #1632, opened 2017-12-16
Fix the ATen static building issue on CUDA

houseroad:
@Yangqing @pietern 

With https://github.com/caffe2/caffe2/pull/1627, Caffe2 can statically built with USE_ATEN=ON and USE_CUDA=OFF. But the function deleterFor defined in aten_op_template.h causes duplicated symbols in libcaffe2.a and libcaffe2_gpu.a.

I checked at only one place we call this function, so directly manually inline it into the caller. Later when we use it at other places, we can just extract it again, and put the implementation in aten_op.cc.

### DIFF
--- a/caffe2/contrib/aten/aten_op_template.h
+++ b/caffe2/contrib/aten/aten_op_template.h
@@ -34,14 +34,6 @@ namespace caffe2 {
 
 using at::Half; // for AT_FORALL_SCALAR_TYPES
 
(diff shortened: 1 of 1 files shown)


pietern:
LGTM

cc @zdevito

zdevito:
Looks good. `static inline` does the same thing in this case, but since we only use it once this fix is fine.

Yangqing:
LGTM :)

PR #1640, opened 2017-12-19
Adding conda specific script to macos builds

pjh5:


harouwu:
Hi @pjh5 , thanks for working on this. A noob question here, is this PR focusing on supporting conda install for Caffe2?

conda/build.sh
```diff
@@ -1,21 +1,65 @@
...
 fi
+echo "Installing caffe2 to ${PREFIX}"
 
 # conda build will copy everything over, including build directories.
-# Don't let this pollute hte build!
+# Don't let this pollute the build!
 rm -rf build || true
 
+# Default leveldb from conda-forge doesn't work
+pip install leveldb==0.18
```
pietern: Does this mean that a default conda install of the package won't work?

pjh5: @pietern We can just turn leveldb off. I will remove this, as we do turn leveldb off later.

@harouwu Sorry I didn't notice your question earlier. These are the first steps to supporting conda install, but this PR won't enable that just yet

PR #1658, opened 2017-12-24
cmake: handle CUDA 9.1 in GCC version check

dbermond:
GCC version check is currently being skipped when using the
newly released CUDA 9.1.

This will also handle other CUDA 9.x minor releases if any,
reducing our work if there are such releases like 9.2. This
assumes that the next major CUDA version will be 10.0,
needing adjustment only after such major version is
released.

### DIFF
--- a/cmake/Dependencies.cmake
+++ b/cmake/Dependencies.cmake
@@ -326,13 +326,14 @@ endif()
 if(USE_CUDA)
   include(cmake/Cuda.cmake)
   if(HAVE_CUDA)
-    # CUDA 9.0 requires GCC version <= 6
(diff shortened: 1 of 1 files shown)


pietern:
Nice, thanks @dbermond!

pietern:
FYI, since we use [nvidia/cuda](https://hub.docker.com/r/nvidia/cuda/) as base image for our Docker builds, and it doesn't have a CUDA 9.1 version yet, we don't run tests for CUDA 9.1 yet. Once this is added, we can easily add a build for CUDA 9.1.

cc @slayton58

dbermond:
Thank you @pietern for working on this.

PR #1661, opened 2017-12-26
Make Jenkins test directory less strict.

ezyang:
I need this change for my plans in CI side, where I plan to mount a
directory onto the test directory so that I can easily transfer the test
files back to a host Jenkins instance.

Signed-off-by: Edward Z. Yang <ezyang@fb.com>


ezyang:
CC @pietern

ezyang:
Actually turns out I don't need this.

Issue #1676, opened 2017-12-29
caffe2 python module should install to the site-packages / dist-packages directory


Erotemic:
When building caffe2 I set my `CMAKE_INSTALL_PREFIX` to `~/venv3`, which correctly installed most of the files (headers, shared libs, binaries, cmake share files) to the correct locations in: `$PREFIX/include`, `$PREFIX/lib`, `$PREFIX/bin`, and `$PREFIX/share`). 

However, the `caffe2` (and `caffe`) python module was simply installed into `$PREFIX`. As such python has no way of finding this and the `PYTHONPATH` needs modification to use the package. 

In my case, the more canonical behavior would have been to install the `caffe2` module to `~/venv3/lib/python3.5/site-packages/`. More generally, it should be placed in a `site-packages` (or `dist-packages) directory depending on the platform. 

The CMake code to find the correct path for installing python modules is as follows: 

This first bit just defines some helper functions: 
```cmake

###
# Removes common indentation from a block of text to produce code suitable for
# setting to `python -c`, or using with pycmd. This allows multiline code to be
# nested nicely in the surrounding code structure.
#
# This function respsects PYTHON_EXECUTABLE if it defined, otherwise it uses
# `python` and hopes for the best. An error will be thrown if it is not found.
#
# Args:
#     outvar : variable that will hold the stdout of the python command
#     text   : text to remove indentation from
#
function(dedent outvar text)
  # Use PYTHON_EXECUTABLE if it is defined, otherwise default to python
  if ("${PYTHON_EXECUTABLE}" STREQUAL "")
    set(_python_exe "python")
  else()
    set(_python_exe "${PYTHON_EXECUTABLE}")
  endif()
  set(_fixup_cmd "import sys; from textwrap import dedent; print(dedent(sys.stdin.read()))")
  # Use echo to pipe the text to python's stdinput. This prevents us from
  # needing to worry about any sort of special escaping.
  execute_process(
    COMMAND echo "${text}"
    COMMAND "${_python_exe}" -c "${_fixup_cmd}"
    RESULT_VARIABLE _dedent_exitcode
    OUTPUT_VARIABLE _dedent_text)
  if(NOT ${_dedent_exitcode} EQUAL 0)
    message(ERROR " Failed to remove indentation from: \n\"\"\"\n${text}\n\"\"\"
    Python dedent failed with error code: ${_dedent_exitcode}")
[...]

PR #1677, opened 2017-12-29
cmake: python packages now install to the cannonical directory

Erotemic:
Addresses issue #1676

Now when `make install` is run, the `caffe2` (and `caffe`) python modules will be installed into the correct site-packages directory (relative to the prefix) instead of directly in the prefix.


Yangqing:
Thanks for the fix!

cc @ezyang @pjh5 @bddppq who have been looking into python builds - does this work for e.g. pip and conda?

ezyang:
For the conda package, we can get rid of the hack where we were manually copying the Python files to the right place. So... definitely an improvement :)

ezyang:
It seems like the patch needs to do a lot of work to get its grubby hands on the site-packages directory as per Python. Is this really the cmake-recommended method of doing this? I did some brief Googling but this does seem like a fairly broken corner of cmake, so maybe this really is the only way to manage it.

Erotemic:
@ezyang As far as I'm aware there is no "CMake recommended way" of getting the site packages directory (I work for the company that develops CMake). However, there is a "CMake recommended way" of getting the `PYTHON_EXECUTABLE` (via `find_packages`). Calling the python executable via the command line is the way I have done it in my projects thus-far. IMHO it seems rather elegant to ask python where we should place its modules. 

The helper functions I dropped in ` cmake/Utils.cmake` are CMake functions I wrote to make executing python scripts much easier from within CMake. The actual code in `caffe2/CMakeLists.txt` is really quite small. You *could* use `execute_process` directly to call python, but its a bit uglier and requires that you are careful with your initial indentation. The helper functions make the real essence of the code clear and concise.

bddppq:
This looks good to me. But the [setup.py](https://github.com/caffe2/caffe2/blob/master/setup.py#L129) needs to be updated as well.

bddppq:
@caffe2bot test this please

bddppq:
@caffe2bot test this please

pjh5 (approved):

Yangqing:
@caffe2bot test this please.

Erotemic:
@bddppq I'm not sure what lines 129 through 131 are trying to do. It looks like its taking things from the install directory and moving it to the build directory. That makes no sense to me. `copy_tree` does takes the args in the order `src, dst` right? Is `build_lib` not the build directory? 

Do these 3 lines just need to be removed? Likewise should the `get_outputs` function change?

bddppq:
@Erotemic It's copying python and libs files from cmake install directory to distutils build directory, in later step they will be copied to users' python installation (by distutils/setuptools). `get_outputs` are to tell setuptools what have been installed (and so when users do uninstall it can cleanly remove them). I think you just need to change from where it copies the caffe and caffe2 directories, the rest should be the same.

Erotemic:
Ah, I see. The setup.py script is an entry point and python is actually running cmake. That makes more sense. I think my new changes should resolve that issue. 

I added a few more changes now that I understand how setup.py is being used. 

* First, I added comments to hopefully make things a bit more clear.
 * Second there was an issue with CMAKE_ARGS, if any of these were specified then the build failed because the TOP directory needs to be the last argument to cmake. 
* Lastly, I work on a machine that is shared by a bunch of people, so the default option of building with all available CPUs is not desirable. I added the `--jobs` command line argument to setup.py build, which is then passed to make.

bddppq:
@Erotemic Nice! 
Two minor comments:
1. `build_ext` is old-style class, so we can not use `super`
2. I feel the jobs count thing should be dealt in bulid_local.sh instead of in setup.py.

bddppq:
@caffe2bot test this please

Erotemic:
I will make the super change. As for handling -j the setup.py needs to intercept it or it will raise an unexpected argument error. Technically you can circumvent this by exporting the CAFFE_NCPUS variable, so build_local does handle it in a way.

bddppq:
@caffe2bot test this please

bddppq:
@Erotemic I think you need to update the CI test scripts as well:
https://github.com/caffe2/caffe2/blob/master/.jenkins/build.sh: the code at the bottom of this file tries to make caffe2 importable in python without setting PYTHONPATH, which is not needed thanks to your work
https://github.com/caffe2/caffe2/blob/master/.jenkins/test.sh: the test command hardcodes the relative path `caffe2/python` which has been changed in this PR.

cc @pietern

.jenkins/test.sh
```diff
@@ -63,18 +62,25 @@ if [[ "$BUILD_ENVIRONMENT" == *-cuda* ]]; then
   EXTRA_TESTS+=(caffe2/contrib/nccl)
 fi
 
+# Add the site-packages in the caffe2 install prefix to the PYTHONPATH
+SITE_DIR=$($PYTHON -c "from distutils import sysconfig; print(sysconfig.get_python_lib(prefix=''))")
+export PYTHONPATH="${PYTHONPATH}:/usr/local/caffe2/$SITE_DIR"
```
bddppq: I think with your changes there is no need to export PYTHONPATH

bddppq:
@caffe2bot test this please

.jenkins/test.sh
Erotemic: I thought so at first, but the default python points to `/usr/local/lib/pythonX.Y/site-packages`. So, if path to the caffe python module was `/usr/local/lib/pythonX.Y/site-packages/caffe2`, then it would work. 

However, looking at the script, it seems like your installing to to `/usr/local/caffe2` instead of `/usr/local`. This means the python module will exist at `/usr/local/caffe2/lib/pythonX.Y/site-packages/caffe2`, so it wont be in the `PYTHONPATH` by default (which I think is the correct behavior when install to `/usr/local` with a project-name-prefix).

To make this clear maybe its best to replace `/usr/local/caffe2` with a variable named `CAFFE2_INSTALL_PREFIX`? Then it would be easy to change where you install if you didn't want to install to the jenkins system to run tests.

pjh5:
@Erotemic I think you need to remove this line https://github.com/caffe2/caffe2/blob/master/conda/build.sh#L67 to make it work on the conda builds.

bddppq:
@caffe2bot test this please

.jenkins/test.sh
bddppq: @Erotemic The default cmake install prefix should be /usr/local, thus with your changes, the caffe2 files will be installed into /usr/local/lib/pythonX.Y/site-packages/caffe2. I believe you can remove the PYTHONPATH export line.

orionr:
Hi @Erotemic - any chance you can address @bddppq's comment about the PYTHONPATH? Once that's done we should be able to merge this. Thanks!

Erotemic:
@orionr I will look into it either tonight or tomorrow.

.jenkins/test.sh
Erotemic: If you look in build.sh the CMAKE_INSTALL_PREFIX is set to "/usr/local/caffe2", so this line would be necessary. Due to this I did create a variable to make it clear where the prefix is in test.sh. However, there still is the issue of two failing tests. I'm not 100% sure what's going on or if its the fault of this branch.

.jenkins/test.sh
```diff
@@ -64,7 +67,7 @@ fi
 
 # Add the site-packages in the caffe2 install prefix to the PYTHONPATH
 SITE_DIR=$($PYTHON -c "from distutils import sysconfig; print(sysconfig.get_python_lib(prefix=''))")
-export PYTHONPATH="${PYTHONPATH}:/usr/local/caffe2/$SITE_DIR"
+export PYTHONPATH="${PYTHONPATH}:${INSTALL_PREFIX}/${SITE_DIR}"
```
bddppq: @Erotemic I believe with your changes there is no need to export `PYTHONPATH`

bddppq:
@caffe2bot test this please

Erotemic:
I see now the failing test is lib/python2.7/dist-packages/caffe2/python/operator_test/ngram_ops_test.py::TestNGramOps::test_ngram_from_categorical_op FAILED 

There is one value in an array that differs

```
02:05:27 AssertionError: 
02:05:27 Arrays are not equal
02:05:27 (mismatch 10.0%)
02:05:27  x: array([68, 12, 50, 59, 65, 33, 16, 28,  0, 51, 65, 28, 34, 16, 56, 10, 56,
02:05:27         8, 32, 74])
02:05:27  y: array([68,  0, 50, 59, 65, 33, 16, 16,  0, 51, 65, 28, 34, 16, 56, 10, 56,
02:05:27         8, 32, 74], dtype=int32)
```

I this file has recently been changed. Perhaps a rebase would case the pr dashboards to become green?

bddppq:
@Erotemic yeah I think this recent master commit fixes that test: https://github.com/caffe2/caffe2/commit/10518848a06ed154d2fe20aa0cd16d557d95167e . Please rebase with master.

Erotemic:
@bddppq just rebased.

bddppq:
@onnxbot test thisbolease

bddppq:
@caffe2bot test this please

bddppq:
@caffe2bot test this please

bddppq:
@caffe2bot test this please

bddppq:
@caffe2bot test this please

bddppq:
@caffe2bot test this please

bddppq:
@caffe2bot retest this please

bddppq:
@Erotemic Thanks for your contribution :-)

Erotemic:
@bddppq Was this merged? It looks like it was simply closed?

orionr:
Hi @Erotemic - it was committed through our internal system, which might have required a few tweaks to get it landed and so might not match your branch 100%. However, it was definitely committed at

https://github.com/caffe2/caffe2/commit/c20584e07c4a75d27f829fd60bf7a37c97c01523

Again, thank you, and I've verified those changes are in master.

Erotemic:
Ah, make sense. Thanks for letting me know. Glad to have helped.

PR #1681, opened 2017-12-31
Fix RNN scoping situation

salexspb:
There is a long lasting problem of scoping which was introduced in original python wrappers early in H1. Basically each RNNCell implemented has to manually scope outputs of each of the operators. If somebody forgets, then there could be weird bugs with layers etc.

Approach is the following. User has to explicitly specify current scope when using  apply_over_sequence function and others if the function is going to be called several times (like for stacking layers). This way we use Caffe2 native scoping approach instead of inventing one extra API people have to use (i.e. passing scope name as an argument to the RNNCell constructor).

### DIFF
--- a/caffe2/python/models/seq2seq/seq2seq_util.py
+++ b/caffe2/python/models/seq2seq/seq2seq_util.py
@@ -579,7 +579,6 @@ def build_embedding_decoder(
             input_size = decoder_cells[-1].get_output_dim()
 
         cell = rnn_cell.LSTMCell(
-            name=get_layer_scope(scope, 'decoder', i),
             forward_only=forward_only,
             input_size=input_size,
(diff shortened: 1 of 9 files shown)


pietern:
@salexspb I think @caffe2bot is not happy that your commit shows up as authored by `Alexander Sidorov <salex@Tophers-Desktop.local>`. Can you `git config --global user.email` and `git commit --amend --reset-author`? Then you can force push to update the commit here.

salexspb:
@pietern , I actually have put this  as  FYI for now, need to fix some use-cases internally as well. The e-mail was fixed after I created this commit, I think. As another PR worked fine (the one you already merged)

salexspb:
@caffe2bot retest this please

Issue #1638, opened 2017-12-18
difference between macro USE_METAL, CAFFE2_USE_MPSCNN & CAFFE2_MOBILE 


xhsoldier:
For iOS system, do I need to enable all these macros?
USE_METAL, 
CAFFE2_USE_MPSCNN
CAFFE2_MOBILE

pietern:
@bwasti Do you know this?

jerryzh168:
USE_METAL is deprecated.
If you just use CPU, you can enable CAFFE2_MOBILE.
If you want to use iOS GPU, you should enable both CAFFE2_MOBILE and CAFFE2_USE_MPSCNN.

xhsoldier:
i enable the both CAFFE2_MOBILE and CAFFE2_USE_MPSCNN. then compile with a error:
caffe2/caffe2/mobile/contrib/ios/mpscnn/mpscnn.mm:24:10: fatal error: 'caffe2/mobile/fb/aloha/headers.h' file not found
#include "caffe2/mobile/fb/aloha/headers.h"

sasikiran:
@jerryzh168 I couldn't get the master branch compile with flags CAFFE2_USE_MPSCNN and CAFFE2_MOBILE. It only seems to accept USE_METAL flag though.

jerryzh168:
There are problems with header files. We're looking into it, will give an update when it is resolved.

Yangqing:
Should be fixed in #1690 now - feel free to reopen if it does not work.

xhsoldier:
the libcaffe2.a was succussfully built. But when I use the lib, I got a error:
```
Undefined symbols for architecture arm64:
  "_OBJC_CLASS_$_MPSCNNSoftMax", referenced from:
      objc-class-ref in libcaffe2.a(mpscnn.mm.o)
  "_OBJC_CLASS_$_MPSCNNPoolingMax", referenced from:
      objc-class-ref in libcaffe2.a(mpscnn.mm.o)
  "_OBJC_CLASS_$_MPSCNNPoolingAverage", referenced from:
      objc-class-ref in libcaffe2.a(mpscnn.mm.o)
  "_OBJC_CLASS_$_MPSCNNNeuronTanH", referenced from:
      objc-class-ref in libcaffe2.a(mpscnn.mm.o)
  "caffe2::utils::ComputeAllAnchors(caffe2::Tensor<caffe2::CPUContext> const&, int, int, float)", referenced from:
[...]

Issue #1162, opened 2017-09-03
code not found in mpscnn.mm


hpsoar:
```
#include "caffe2/fb/aloha/generate_proposals_op.h"
#include "caffe2/fb/aloha/generate_proposals_op_util_boxes.h"
```

Yangqing:
Should be fixed in #1690 now - feel free to reopen if it does not work.

Issue #298, opened 2017-04-19
iOS Example


sbryfcz:
This is really just a question/enhancement. I'd love to see an example of what needs to happen to build an XCode/iOS project using this library. I was able to build_ios just fine, but its unclear what the next steps are. Thanks for the great work.

Yangqing:
Yup, would love to create examples as follow-ups in the next weeks; cc'ing @bwasti who's our mobile expert.

mjohnst:
I have started attempting an integration of the C++ API into an Objective-C++ project. I can post more when I've gotten something running inference on a device.

BrianOn99:
@mjohnst Thanks in advance!  Please show your result once you get something working!

abjurato:
@bwasti, another couple of iOS-related questions:
1. are you planning Metal support? If yes, would it be based on MPSCNN?
2. are you planning any Swift wrapper?

RobertBiehl:
@mjohnst Hey, I'm attempting the same thing.
I built the static library with build_ios.sh and am now trying to build a pod on top of that.

My problem is when referencing the header
`#include "caffe2/core/predictor.h"` in an _Objective C/Objective C ++_  file, this leads to the following errors:
<img width="939" alt="screen shot 2017-04-22 at 11 08 05" src="https://cloud.githubusercontent.com/assets/446619/25303078/0307ec8e-274c-11e7-8fc8-6a650b28eefd.png">

I believe this is due to the fact that the required header has a chain dependency to a header `#include "caffe2/core/net.h"` which contains actual implementation and references c++ libs which might collide with what's been used when building the static lib.

RobertBiehl:
@abjurato A Swift wrapper is what I'm trying to build. Just a simple plug and play pod ;)
It seems that I've become soft when it comes to integrating unwieldy c++ libs...

I'll post when I get there...

jquave:
I've been able to get this up and running: [caffe2 on iOS tutorial](http://jamesonquave.com/blog/caffe2-on-ios-deep-learning-tutorial/).

RobertBiehl:
Thanks @jquave.
I hope I'll have simple to use solution soon.

```
2017-04-22 23:30:29.201545+0200 Caffe2Test[4036:2762929] Inference results:
0: birdhouse - 31.4025%
1: carton - 20.0337%
2: crossword puzzle, crossword - 2.15863%
3: perfume, essence - 1.46789%
4: perfume, essence - 1.46789%
```

[...]

PR #1690, opened 2018-01-03
Properly include the generate proposal headers.

Yangqing:
The header files will be committed separately from fbcode.

cc @pietern


Yangqing:
This should close #1652 #1162 #1638 and #298 .

PR #1709, opened 2018-01-05
Remove build status from caffe2.ai install pages.

orionr:
* Not relevant for installs and link seems to fail even though they are succeeding on the homepage.
* Also switched Android typo to iOS.


JoelMarcey (commented):
Approved -- even if already merged :)

PR #1711, opened 2018-01-06
Lock hypothesis in Docker image to 3.44.6

pietern:
Version 3.44.7 doesn't work with the ancient version of setuptools
that is included in Ubuntu Trusty (setuptools 3.3).

Also see HypothesisWorks/hypothesis-python@eadd62e467d6cee6216e71b391951ec25b4f5830.

@Yangqing @ezyang Should we lock versions of all packages here?

### DIFF
--- a/docker/jenkins/common/install_python.sh
+++ b/docker/jenkins/common/install_python.sh
@@ -88,10 +88,13 @@ pushd pip-9.0.1
 popd
 rm -rf pip-9.0.1*
 
-# Install pip packages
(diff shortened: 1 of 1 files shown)


ezyang:
If you want a reproduceable build, yes you should lock everything. However, I find that people tend to forget to ever upgrade consequently. But for a Trusty dockerfile that seems fine.

pietern:
This was fixed upstream. No longer an issue.

PR #1725, opened 2018-01-10
Use GCC to compile Android Caffe2

sf-wind:
It seems GCC performs better. Always use it to compile.


Yangqing:
For whomever on the oss side wondering - we have observed NNPACK performance to be better on gcc compared to clang, due to the former generating higher performance assembly. cc @Maratyszcza FYI.

Maratyszcza:
LGTM

pietern:
@caffe2bot test this please

PR #1612, opened 2017-12-12
Import the FindCUDA module from CMake

pietern:
This addresses a problem where a generator expression for include
directories was passed to NVCC, where the generator expression itself
was prefixed by -I. As the NNPACK include directory generator
expression expands to multiple directories, the second and later ones
were not prefixed by -I, causing nvcc to return an error.

This was first fixed in CMake 3.7 (see Kitware/CMake@7ded655f).

To fix this, we can import and track FindCUDA from CMake upstream.

The most recent change to these files is Kitware/CMake@21806e6f.

This fixes #1397.

PR #1727, opened 2018-01-10
Backport FindCUDA functionalities from CMake

Yangqing:
This is in principle similar to #1612 and is tested on Windows 2017. CMake passes, although there are still bugs in the MSVC compiler that prevents cuda to compile properly.

The difference between this and #1612 is that this diff explicitly puts the CMake files into a separate folder and uses a MiscCheck.cmake chunk of code to test whether we need to include them. See README.txt for more details.


pietern:
Just to confirm, you pulled this from CMake master? Or a released version? It could be helpful to include the SHA1 of the CMake commit you pulled this from so we know where it came from.

This also implies that if you're using Visual Studio, you must be using CMake 3.0, which is fine IMO.

pietern:
What prevents us from always using these files? If it is only the CMake 3.0 requirement then I think we should bite the bullet and dump support for 2.8.

Yangqing:
@pietern yep this is pulled from master (hash in the README.txt: Kitware/CMake@bc88329e) I think in the future we should always upstream fixes to cmake instead of having local copies out of sync with cmake master - like, in the pytorch case we more or less lost track of what local changes we made.

RE always using this - we can defer the choice to a later date when we officially deprecate cmake 2.8.x. Also, we might want to not invade into the user's default cmake settings in case they have their own FindCUDA settings etc?

Yangqing:
And yep - for VS2015/2017 you will be using cmake 3.x so the version should be fine.

Yangqing:

cmake/MiscCheck.cmake
```diff
@@ -159,3 +159,15 @@ endif()
 
 # ---[ Create CAFFE2_BUILD_SHARED_LIBS for macros.h.in usage.
 set(CAFFE2_BUILD_SHARED_LIBS ${BUILD_SHARED_LIBS})
+
+# ---[ Check if we will need to include the local Modules_CUDA_fix folder.
+# Add your conditions here if needed.
+if (MSVC)
```
yinghai, line 165: Recently I found that with this diff, we need `set(CAFFE2_CMAKE_USE_LOCAL_FINDCUDA ON)` for linux systems with cmake <3.7 to avoid the nvcc error. Maybe add a cmake version check here too?

PR #1728, opened 2018-01-10
Adapting conda build to work for ubuntu and adding a flag to control precedence of Anaconda include dirs

pjh5:
This should fix Protobuf version problems on all Anaconda builds by putting include directories under Anaconda before all other include directories.


pietern:
@pjh5 In principle this looks good to me. I'm wondering if we should flip it around though. Revert default behavior to just doing whatever is needed, and then opting OUT of Anaconda include dirs, instead of opting IN for Anaconda include dirs.

scripts/build_local.sh
```diff
@@ -22,7 +22,7 @@ fi
 
 # Use special install script with Anaconda
 if [ -n "${USE_ANACONDA}" ]; then
-  conda build "$CAFFE2_ROOT/conda"
+  CONDA_CMAKE_ARGS=CMAKE_ARGS conda build "$CAFFE2_ROOT/conda"
```
pietern: Needs to be CONDA_CMAKE_ARGS="${CMAKE_ARGS}"

caffe2/CMakeLists.txt
```diff
@@ -102,7 +102,6 @@ target_link_libraries(caffe2 PRIVATE ${Caffe2_DEPENDENCY_LIBS})
 target_include_directories(caffe2 INTERFACE $<INSTALL_INTERFACE:include>)
 target_compile_options(caffe2 INTERFACE "-std=c++11")
 install(TARGETS caffe2 EXPORT Caffe2Targets DESTINATION lib)
-link_directories(${CMAKE_INSTALL_PREFIX}/lib)
```
Yangqing, line 105: Thanks for catching this :)

cmake/Utils.cmake
```diff
@@ -264,10 +266,22 @@ endfunction()
 #
 function(caffe2_include_directories)
   foreach(path IN LISTS ARGN)
-    if (${path} MATCHES "/anaconda")
-      include_directories(AFTER SYSTEM ${path})
+    if (${DEPRIORITIZE_ANACONDA})
```
Yangqing, line 269: So I think we should probably going to enforce "if conda, only conda" to simplify this chunk, but I guess it is fine for now. Mixing system headers and conda headers is a bit error prone.

pietern: Totally agree.

PR #1729, opened 2018-01-10
Move Android.mk

pietern:
cc @sf-wind


pietern:
Merged in 0808040f9e03997defec86a884d19adbb2efe573.

Maratyszcza:
Do we even need it? Lets not pretend that we support `ndk-build`, and just remove this file.

PR #1730, opened 2018-01-10
NNPACK: Use new bindings and custom thread pool

Maratyszcza:
This change should dramatically (~10X) improve performance of convolution with NNPACK engine


Maratyszcza:
@caffe2bot test this please

bddppq:
@caffe2bot add to whitelist

PR #1738, opened 2018-01-11
Adapting conda for ubuntu and fixing most protobuf problems for conda builds

pjh5:
The other PR wasn't being imported to FB correctly so I'm making a copy


pjh5:
The other PR worked

PR #1739, opened 2018-01-11
fix gru_cell bug

anderspapitto:
the fc needs to be in the output_gate_t scope so it can find its input
weights correctly


anderspapitto:
this fixes the failure in https://github.com/onnxbot/onnx-fb-universe/pull/168

PR #1740, opened 2018-01-11
Use android.cmake.toolchain from Android NDK

Maratyszcza:
The android.cmake.toolchain file we use from a submodule is unmaintained and not updated since 2015.
It causes numerous problems in Caffe2 build:
- Caffe2 can't be built for Android ARM64, because gcc toolchain for ARM64 doesn't support NEON-FP16 intrinsics, and the android.cmake.toolchain we use doesn't allow us specify clang-5.0 from NDK r15c
- Caffe2 can't be built with Android NDK r16 (the most recent NDK version)
- Caffe2 can't be built for Android with Ninja generator

This change updates the build script to use $ANDROID/build/cmake/android.cmake.toolchain instead, which is maintained by Android team, and synchronized with Android NDK version.
As this toolchain file doesn't support "armeabi-v7a with NEON FP16" ABI, I had to disable mobile OpenGL backend, which requires NEON-FP16 extension to build. With some work, it can be re-enabled in the future.


Yangqing (commented):
I feel that we can backport the most recent android toolchain cmake file from ndk and patch it locally, similar to what we did with the cmake cuda fix in #1727 ?

Maratyszcza:
@Yangqing I don't think it is worth it. New NDK versions come with CMake toolchain files, lets just use those.

Maratyszcza:
@caffe2bot retest this please

Yangqing:
Sounds good, although I am still concerned that this will not be able to build fp16, meaning it completely blocks the Android opengl gpu path - that sounds suboptimal? cc @bwasti

Maratyszcza:
@Yangqing I can fix the OpenGL backend build with Android NDK toolchain file, it just requires more work.

Maratyszcza:
@caffe2bot retest this please

Maratyszcza:
@caffe2bot retest this please

Maratyszcza:
@caffe2bot retest this please

pietern:
The images have been updated and the Android one now has CMake 3.6.3:

```
$ cmake --version
cmake version 3.6.3
```

@caffe2bot retest this please

Maratyszcza:
@caffe2bot retest this please

ca1773130n:
@Yangqing I've made a workaround for NDK to compile with OpenGL. Check [this](http://cameleonx.com/blog/2018/03/27/how-to-enable-gpu-acceleration-in-caffe2-for-android/) out.
@Maratyszcza I'm just wondering why Facebook is going to deprecate OpenGL backend. Could you explain a little about the reason?

Maratyszcza:
@ca1773130n The OpenGL 3.0 backend is no longer used internally at Facebook, and no one on Caffe2 team maintains it. It something breaks down, you are on your own to fix.
The path forward for GPU inference on mobile is Qualcomm SNPE backend for Qualcomm Adreno GPUs, ARM ComputeLibrary GLES backend for ARM Mali GPUs, and Metal/MPSCNN backend for Apple GPUs.

ca1773130n:
@Maratyszcza Thank you for your explanation. Great!

PR #1741, opened 2018-01-11
Use protoc path to figure out its install prefix 

pietern:
[x] Have to rebase
[x] Have to ensure this works on macOS + Anaconda


pjh5 (approved):
LGTM! We can do it for darwin too. I could look into it if you want

pietern:
This is now good to go. @Yangqing can you take a look as well to see if all looks good?

After merging this I also want to change the CUSTOM_PROTOBUF variable to indicate that it's intended to be used for cross compilation and nothing else.

pietern:
Removed the helper script and now use get_filename_component to figure out the path.

It's probably safe to assume that we can find the libraries in PREFIX/lib if protoc is installed in PREFIX/bin.

No need to query the RPATH.

gokceneraslan:
`protoc` command cannot find libprotobuf.so when I compile bundled protobuf from caffe within a conda environment.

PR #1742, opened 2018-01-11
Add ElmanCell and ElmanRNN

anderspapitto:


caffe2/python/rnn_cell.py
```diff
@@ -234,6 +234,71 @@ def create_states(self, model):
...
+        brew.fc(
+            model,
+            hidden_t_prev,
+            gates_t,
+            dim_in=self.hidden_size,
+            dim_out=self.hidden_size,
+            axis=2,
+        )
+        brew.sum(model, [gates_t, input_t], gates_t)
+        if self.activation == 'tanh':
```
dzhulgakov, line 283: throw if non of the expected ones? and check in the constructor?

caffe2/python/rnn_cell.py
```diff
@@ -234,6 +234,71 @@ def create_states(self, model):
         ]
 
 
+# based on http://pytorch.org/docs/master/nn.html#torch.nn.RNNCell
+class ElmanCell(RNNCell):
```
dzhulgakov: Hm, I haven't seen it called Elman before - it's usually just "rnn" (which is taken) or "simple" or "basic".

Maybe BasicRNNCell then? (similar to TF)

caffe2/python/rnn_cell.py
```diff
@@ -1405,6 +1470,7 @@ def _LSTM(
 
 
 LSTM = functools.partial(_LSTM, LSTMCell)
```
dzhulgakov, line 1480: hm, this function is not doing just LSTM any more, maybe we should rename it

caffe2/python/rnn_cell.py
anderspapitto: the pytorch docs describe it as "a multi-layer Elman RNN", but yeah I can switch to Basic

caffe2/python/rnn_cell.py
anderspapitto: to _RNN?

caffe2/python/rnn_cell.py
anderspapitto: done

caffe2/python/rnn_cell.py
anderspapitto: done

PR #1747, opened 2018-01-12
Install CMake 3.6.3 in base image for Android build

pietern:
This is needed for #1740.

Verified that `./build.sh py2-android-ubuntu16.04` builds an Android base image with CMake 3.6.3.

### DIFF
--- a/docker/jenkins/build.sh
+++ b/docker/jenkins/build.sh
@@ -69,6 +69,10 @@ fi
 
 if [[ "$image" == *-android-* ]]; then
   ANDROID=yes
+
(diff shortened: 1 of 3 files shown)


Maratyszcza:
Thanks @pietern!

Maratyszcza:
LGTM. Two things to check:
- Is the new CMake in $PATH?
- Do we still use the old CMake for non-Android builds?

pietern:
@Maratyszcza Yes and yes. We don't rev CMake for the other builds, since the system default version is what people will interact with first, so we should ensure it works.

Merging this change and rebuilding the base images.

PR #1762, opened 2018-01-18
Semi-automatically generate scripts out of our tutorials

salexspb:
The idea is the following. We are going to automatically generate .py files using a jupyter post-save hook. Also, there is a script to generate these for all the tutorials. The script is also used from Jenkins test.sh. So if you don't run the sync anyhow, test will complain. 

In this diff I include the framework itself + .py files generated for all tutorials. They live under a separate folder.


orionr:
Nicely done! I'll import it.

PR #1765, opened 2018-01-18
Adding a separate script for anaconda builds

pjh5:
Lots of unwanted stuff here that shouldn't be in this branch. I just need to make a PR so I can test it


pietern:
LGTM

PR #1766, opened 2018-01-18
Increase lower bound of values for values in div test

pietern:
This should translate to an 1% error margin. The gradient checker uses a .5% threshold.

### DIFF
--- a/caffe2/python/hypothesis_test.py
+++ b/caffe2/python/hypothesis_test.py
@@ -203,7 +203,7 @@ class TestOperators(hu.HypothesisTestCase):
             return (x / y, )
 
         def non_zero(x):
-            return abs(x) > 10e-5
+            return abs(x) > 1e-2
 
         def div_dtypes():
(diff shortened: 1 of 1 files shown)


pietern:
FWIW this failed after updating the base images. I think the seed is influenced by new binaries and that we don't randomize them at all (and we probably should do that).

See https://ci.pytorch.org/jenkins/job/caffe2-builds/job/py2-gcc5-ubuntu16.04-test/401/

Maratyszcza:
LGTM

Maratyszcza (approved):

Maratyszcza:
As for seed, I suggest we use static seed for PR builds (and maybe a random seed for nightly builds). When tests fail due to unfortunate seed, it is still a bug, but I'd prefer to separate this signal from signal on PR changes.

pietern:
@Maratyszcza I agree wrt static seed. We can run master + PRs on a fixed seed and then have a periodic build with a random seed. Then we can deal with the flakiness asynchronously.

PR #1767, opened 2018-01-19
Assorted small change to conda scripts

pjh5:
More changes to be added later. I need to make a PR so that I can point jenkins to this


conda/meta.yaml
```diff
@@ -15,29 +15,28 @@ build:
...
+    - cmake
     - future
     - glog
-    - numpy
-    - six
-    - python
-    - protobuf {{ protobuf }}
-    - cmake
     - lmdb
-    - snappy ==1.1.6
```
pietern, line 26: No snappy?

docker/jenkins/common/install_anaconda.sh
```diff
@@ -26,22 +27,14 @@ case "$ANACONDA_VERSION" in
   ;;
 esac
 /bin/bash ~/anaconda.sh -b -p /opt/conda
-rm ~/anaconda.sh
+rm -f ~/anaconda.sh
```
pietern: Why?

pietern (commented):
Can you move the cuda_conda to a subdirectory of "conda"? Then we don't pollute the root dir as much and keep all the conda stuff in a single dir.

docker/jenkins/common/install_anaconda.sh
pjh5: I thought it might be the cause of another problem. It wasn't, but it didn't seem like a bad thing to do anyway

pietern: It's not needed though

pietern:
Good to go when the mac failure is fixed. I see the last build took 37 mins. Any clue why it takes this long? IMO we should aim for 10 mins or less.

pjh5:
Anaconda builds always take a really long time :/

I don't actually know what's causing this build error though. They seemed to happen on all the conda mac builds recently

conda/cuda/build.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+# Install script for Anaconda environments with CUDA on linux
+# This script is not supposed to be called correctly, but should be run by:
```
orionr: Nit: correctly -> directly

orionr (approved):

PR #1770, opened 2018-01-19
add test forward pass

pengbo-learn:
Forward pass in test model should fix running mean and running var in batch normalization op.
So another forward pass is needed.


orionr:

PR #1793, opened 2018-01-24
Better support for windows

Yangqing:
Basically, this should make windows {static_lib, shared_lib} * {static_runtime, shared_runtime} * {cpu, gpu} work. A few highlights:

(1) Updated newest protobuf.
(2) use protoc dllexport command to ensure proper symbol export.
(3) various code updates to make sure that C2 symbols are properly shown
(4) cmake file changes to make build proper
(5) option to choose static runtime and shared runtime similar to protobuf
(6) revert to visual studio 2015 as current cuda and msvc 2017 do not play well together.


willyd:
@Yangqing Thanks for bringing caffe2 to Windows. Is the shared library build really working for you? Last time I tried to implement this via WINDOWS_EXPORT_ALL_SYMBOLS I got a  [LNK1189](https://msdn.microsoft.com/en-us/library/at5879hx.aspx) error.

Yangqing:
@willyd thanks for checking this! Yeah, I thought /bigobj resolved that issue:

https://msdn.microsoft.com/en-us/library/ms173499.aspx

which is in https://github.com/caffe2/caffe2/blob/bc3da2301d58548a721a2317514536e844adfc7b/CMakeLists.txt#L109 . Which Windows MSVC version is generating this error?

willyd:
@Yangqing `/bigobj` increases the number of symbols you can have in an object file but apparently does not change the number of objects you can export from a DLL. I just tested your windows_cuda branch with a shared library build and I still get the LNK1189 error. I think the only way to work around this issue is to stop using the `WINDOWS_EXPORT_ALL_SYMBOLS` and add the required `CAFFE2_API` decorators everywhere it is needed.

I am using MSVC 2017.

BTW, you might want to include this [commit](https://github.com/willyd/caffe2/commit/d4789b18b301a11c6bdab93363afeb9f9de3cd99) in this PR to correctly support glog on Windows.

Yangqing:
@willyd thanks for the head-up! Yeah, I also realize that the large build with windows cuda does not work - it ended up getting the same error. We should probably mark things under core/ to be CAFFE2_API and the rest hidden. Would be a good next step to do. Right now I guess we can only do static linking when/if we want cuda.

Yangqing:
@caffe2bot

willyd:
> Yeah, I also realize that the large build with windows cuda does not work - it ended up getting the same error.

I was not using CUDA but I was using glog, gflags and lmdb so this may explain the difference if you are not seeing the error on CPU only builds.

> We should probably mark things under core/ to be CAFFE2_API and the rest hidden. Would be a good next step to do. Right now I guess we can only do static linking when/if we want cuda.

Makes sense.

pietern:
@caffe2bot retest this please

Satge96:
@Yangqing 
Me and a friend have huge problems installing caffe2 on our windows machines. And we an't find a real step-by-step guide (not written or in video form).
Could you provide something like that?

Yangqing:
As it got reverted due to internal build reasons - new PR is at #1827 .

Yangqing:
@Satge96 - currently scripts/build_windows.bat is your friend... we haven't got a complete guide yet but feel free to send issues our way - we appreciate that!

PR #1803, opened 2018-01-25
add opencv dependencies to caffe2

orionr:
The video input op requires additional opencv packages. This is to add them to
cmake so that it can build

### DIFF
--- a/cmake/Dependencies.cmake
+++ b/cmake/Dependencies.cmake
@@ -207,7 +207,7 @@ endif()
 # ---[ OpenCV
 if(USE_OPENCV)
   # OpenCV 3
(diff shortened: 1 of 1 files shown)


orionr:
I'm not sure if the checkpoint changes should have gotten in here. @pietern, thoughts?

pietern:
@orionr That looks like a copy of 4532cd3506c4168b0f84c3a41bc70174eba49947. Perhaps you raced with this commit being merged?

orionr:
Yup - looks like it. I'll check our new script as well. Builds / tests are being kicked off here, though, which is good!

PR #1817, opened 2018-01-26
no_bias in resnet32x32

pengbo-learn:
ResNetBuilder needs at least 4 parameters


harouwu (approved):
Looks good. Thanks for the fix

PR #1819, opened 2018-01-26
Modernizing the gflags dependency in cmake.

Yangqing:
Historically, for interface dependent libraries (glog, gflags and protobuf), exposing them in Caffe2Config.cmake is usually difficult.

New versions of glog and gflags ship with new-style cmake targets, so one does not need to use variables. New-style targets also make it easier for people to depend on them in installed config files.

This diff modernizes the gflags library, and still provides a fallback path if the installed gflags does not have cmake config files coming with it.

It does change one behavior of the build process though - when one specifies -DUSE_GFLAGS=ON but gflags cannot be found, the old script automatically turns it off but the new script crashes, forcing the user to specify USE_GFLAGS=OFF.


Yangqing:
@caffe2bot retest this please

CMakeLists.txt
```diff
@@ -196,7 +196,7 @@ if (USE_GLOG)
   list(APPEND CAFFE2_INTERFACE_LIBS ${GLOG_LIBRARIES})
 endif()
 if (USE_GFLAGS)
-  list(APPEND CAFFE2_INTERFACE_LIBS ${GFLAGS_LIBRARIES})
+  list(APPEND CAFFE2_INTERFACE_LIBS gflags)
```
pietern, line 199: I expect this will be interpreted as the gflags **CMake target**, not -lgflags, correct?

pietern (approved):
Nice work!

pietern:
@caffe2bot retest this please

pietern:
(there was a CI failure for pr/py2-cuda9.0-cudnn7-centos7-build)

CMakeLists.txt
Yangqing: Yep, this is going to be interpreted as the gflags target, which is what public/gflags.cmake does: newer gflags ships with the target definition, for older gflags we construct that target for cmake.

Yangqing:
(there are three errors: 
pr/py2-gcc5-ubuntu16.04-test showing coredump in pytest (known issue)
pr/py2-ios-macos10.13-build showing github account issue
pr/py2-system-macos10.13-build showing github account issue)

@caffe2bot retest this please

PR #1833, opened 2018-01-29
add back cuda auto arch detection

Yangqing:
This was removed in an earlier version. Anyway, I suspect this will make jenkins a bit unhappy (do we use gpu instances for building as well?) so firing a PR to test.


pietern:
@Yangqing FYI we set `-DCUDA_ARCH_NAME=Maxwell` for these builds.

PR #1834, opened 2018-01-29
remove "-s" compilation flag from clang when build for Android

dzyong:
Now we use **clang** to build Caffe2 for Android with arm64-v8a ABI, but clang doesn't support "-s" compilation flag. If we append this flag to clang, it will report a warning:

> clang++: warning: argument unused during compilation: '-s' [-Wunused-command-line-argument]

This submit will check we use gcc or clang to build Caffe2 for Android.


Yangqing:
@caffe2bot test this please

@dzyong thanks for the fix!

PR #1835, opened 2018-01-29
enable USE_MOBILE_OPENGL by default

dzyong:
iOS is also depend on USE_MOBILE_OPENGL, so I think we should only disable it for Android.


dzyong:
Hi @Maratyszcza , would you please help me to review this commit : )

This commit is trying to minimize the impact of temporary disable USE_MOBILE_OPENGL.

Maratyszcza:
LGTM. We are going to deprecate the OpenGL backend soon, but wouldn't hurt to make it usable while it lasts.

PR #1839, opened 2018-01-29
Enable rnn_cell_test in jenkins

salexspb:


pietern:
@salexspb Looks like another one is flaky as well.

PR #1841, opened 2018-01-29
GRU sequence lengths: allow unspecified sequence lengths

anderspapitto:
modeled after the earlier change for LSTM

### DIFF
--- a/caffe2/operators/gru_unit_op.cc
+++ b/caffe2/operators/gru_unit_op.cc
@@ -19,7 +19,7 @@
 namespace caffe2 {
 REGISTER_CPU_OPERATOR(GRUUnit, GRUUnitOp<float, CPUContext>);
 OPERATOR_SCHEMA(GRUUnit)
-    .NumInputs(4)
+    .NumInputs(3, 4)
     .NumOutputs(1)
     .SetDoc(R"DOC(
 GRUUnit computes the activations of a standard GRU,
@@ -35,30 +35,36 @@ value at X[t][n] >= seqLengths[n].
         "drop_states",
         "Bool to determine if hidden state is zeroes or passed "
         "along for timesteps past the given sequence_length.")
-    .Input(0, "hidden_prev", "The previous GRU hidden state.")
-    .Input(
-        1,
-        "gates",
-        "Unactivated gate outputs from forget, update, "
-        "and output gates, pre-activation.")
-    .Input(
-        2,
-        "seq_lengths",
-        "Array of sequence lengths.  "
-        "len(seq_lengths) should equal batch size N.")
-    .Input(3, "t", "The timestep for this operation.")
+    .Arg(
+        "sequence_lengths",
+        "When false, the sequence lengths input is left out, "
+        "and all following inputs are shifted left by one.")
     .Output(0, "hidden", "The new GRU hidden state calculated by this op.");
 REGISTER_CPU_OPERATOR(GRUUnitGradient, GRUUnitGradientOp<float, CPUContext>);
-OPERATOR_SCHEMA(GRUUnitGradient).NumInputs(6).NumOutputs(2);
+OPERATOR_SCHEMA(GRUUnitGradient)
+    .NumInputs(5, 6)
+    .NumOutputs(2)
+    .Arg(
+        "sequence_lengths",
+        "When false, the sequence lengths input is left out, "
+        "and all following inputs are shifted left by one.");
 
 class GetGRUUnitGradient : public GradientMakerBase {
   using GradientMakerBase::GradientMakerBase;
   vector<OperatorDef> GetGradientDefs() override {
-    return SingleGradientDef(
-        "GRUUnitGradient",
-        "",
-        vector<string>{I(0), I(1), I(2), I(3), O(0), GO(0)},
-        vector<string>{GI(0), GI(1)});
+    if (GetFlagArgument(def_, "sequence_lengths", true)) {
+      return SingleGradientDef(
+          "GRUUnitGradient",
+          "",
+          vector<string>{I(0), I(1), I(2), I(3), O(0), GO(0)},
+          vector<string>{GI(0), GI(1)});
+    } else {
+      return SingleGradientDef(
+          "GRUUnitGradient",
+          "",
+          vector<string>{I(0), I(1), I(2), O(0), GO(0)},
+          vector<string>{GI(0), GI(1)});
+    }
   }
 };
 REGISTER_GRADIENT(GRUUnit, GetGRUUnitGradient);
--- a/caffe2/operators/gru_unit_op.h
+++ b/caffe2/operators/gru_unit_op.h
@@ -46,7 +46,7 @@ void GRUUnit(
     T* H,
     Context* /*context*/) {
   for (int n = 0; n < N; ++n) {
-    const bool valid = t < seqLengths[n];
+    const bool valid = seqLengths == nullptr || t < seqLengths[n];
 
     for (int d = 0; d < D; ++d) {
       if (!valid) {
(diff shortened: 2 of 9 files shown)


caffe2/operators/gru_unit_op.cc
```diff
@@ -35,6 +35,7 @@ value at X[t][n] >= seqLengths[n].
         "drop_states",
         "Bool to determine if hidden state is zeroes or passed "
         "along for timesteps past the given sequence_length.")
+    .Arg("no_sequence_lengths", "Ignore the sequence lengths input")
```
salexspb: usually, we do simple optional inputs like this one by changing number of inputs. I think, all other inputs are required. So you could always assume that if there are 3 inputs, they are prev, gates and t. 

Having a graph with 4 inputs but ignoring one of them is not good. It can, for example, confuse graph rewriting tools and gradient generators.

caffe2/operators/gru_unit_op.cc
```diff
@@ -49,7 +50,10 @@ value at X[t][n] >= seqLengths[n].
     .Input(3, "t", "The timestep for this operation.")
     .Output(0, "hidden", "The new GRU hidden state calculated by this op.");
 REGISTER_CPU_OPERATOR(GRUUnitGradient, GRUUnitGradientOp<float, CPUContext>);
-OPERATOR_SCHEMA(GRUUnitGradient).NumInputs(6).NumOutputs(2);
+OPERATOR_SCHEMA(GRUUnitGradient)
```
salexspb: ditto

salexspb (changes requested):

caffe2/operators/gru_unit_op.cc
anderspapitto: so, in this case, I would prefer to keep consistency with LSTMUnit (you can see that I made almost character-for-character the same diff to LSTMUnit here https://github.com/caffe2/caffe2/pull/1764 - and in that case it was the only option), because these two features are already 90% copy-paste from each other, and gratuitously diverging them makes it harder to maintain.

If you feel strongly that it's better to diverge them, I'll make the change you've requested.

caffe2/operators/gru_unit_op.cc
anderspapitto: ditto

pietern:
@caffe2bot retest this please

anderspapitto:
something weird is going on with the tests, where submodules are not available

```
21:02:39 CMake Error at third_party/pthreadpool/CMakeLists.txt:105 (ADD_SUBDIRECTORY):
21:02:39   add_subdirectory given source
21:02:39   "/var/lib/jenkins/workspace/third_party/fxdiv" which is not an existing
21:02:39   directory.
```

that's not my error

anderspapitto:
@caffe2bot retest this please

anderspapitto:
@caffe2bot retest this please

caffe2/python/operator_test/rnn_cell_test.py
```diff
@@ -36,17 +36,20 @@
 import unittest
 
 
-def lstm_unit(hidden_t_prev, cell_t_prev, gates,
-              seq_lengths, timestep, forget_bias=0.0,
-              drop_states=False, no_sequence_lengths=False):
+def lstm_unit(*args, **kwargs):
```
salexspb, line 39: why not make seq_lengths be None in a case it is not provided? Then you don't need this kwargs parsing. I.e. you would need just to rearrange inputs order here a bit, put seq_lengths in the optional section

caffe2/python/operator_test/rnn_cell_test.py
```diff
@@ -126,11 +132,11 @@ def layer_norm_lstm_reference(
         )
 
         hidden_t, cell_t = lstm_unit(
```
salexspb, line 134: May be we should keep kwargs usage, it is less error prone. it is also related to the previous comment

caffe2/python/operator_test/rnn_cell_test.py
```diff
@@ -1569,34 +1575,36 @@ def lstm_with_attention(
            d=st.integers(1, 10),
            t=st.integers(1, 10),
            dtype=st.sampled_from([np.float32, np.float16]),
-           no_sequence_lengths=st.booleans(),
+           sequence_lengths=st.booleans(),
```
salexspb: nit: use_sequence_lengths

caffe2/python/operator_test/rnn_cell_test.py
```diff
@@ -1569,34 +1575,36 @@ def lstm_with_attention(
...
            **hu.gcs)
     def test_lstm_unit_recurrent_network(
```
salexspb: this is not our main test, actually. Could you add use_sequence_lengths to lstm_base, please?

salexspb (changes requested):
Looks like some tests are failing.

caffe2/python/operator_test/rnn_cell_test.py
anderspapitto: not possible - the interface here has to exactly match the interface of the C++ op, I can't choose where to put the args

caffe2/python/operator_test/rnn_cell_test.py
anderspapitto: same answer as above - my hand is forced by how the test machinery uses this function

caffe2/operators/gru_unit_op.h
```diff
@@ -132,10 +132,16 @@ class GRUUnitOp : public Operator<Context> {
       : Operator<Context>(operator_def, ws),
         drop_states_(OperatorBase::template GetSingleArgument<bool>(
             "drop_states",
-            false)) {}
+            false)),
+        sequence_lengths_(OperatorBase::template GetSingleArgument<bool>(
```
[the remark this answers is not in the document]
anderspapitto: no idea - I'm matching the style/usage of the surrounding code. This diff has enough things going on without sprinkling in extra unrelated refactorings :)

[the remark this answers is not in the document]
anderspapitto: i.e. look at drop_states on line 133

caffe2/python/operator_test/rnn_cell_test.py
anderspapitto: sure

caffe2/python/operator_test/rnn_cell_test.py
anderspapitto: sure

caffe2/python/operator_test/gru_test.py
```diff
@@ -184,13 +196,21 @@ def generate_input_state(n, d):
...
 
     with scope.NameScope("test_name_scope"):
-        hidden_t_prev, gates_t, seq_lengths, timestep = \
-            model.net.AddScopedExternalInputs(
-                "hidden_t_prev",
-                "gates_t",
-                'seq_lengths',
-                "timestep",
-            )
+        if sequence_lengths:
```
dzhulgakov, line 200: nit: you can have several calls to AddScopedExternalInputs to avoid duplication

dzhulgakov (approved):
Looks good, even though tests become (forcedly) a bit messy.

Run clang-format before landing?

caffe2/python/operator_test/rnn_cell_test.py
salexspb: so, use_sequence_lengths? :)

caffe2/python/operator_test/rnn_cell_test.py
salexspb: yeah, makes sense

salexspb (approved):
Looks like args, kwargs business in testing code is necessary indeed, should be good to go. Please check nits (from me and Dima).

PR #1842, opened 2018-01-29
Adding gflags to default dependency of conda builds

pjh5:


salexspb:
what about manual installation? Can we update install scripts / documentation page for those not using conda build ?

Yangqing (approved):

pjh5:
@salexspb Sure, but that will be another PR because that's on a different branch

PR #1843, opened 2018-01-29
implement sequence length support for BasicRNN

anderspapitto:


caffe2/python/operator_test/basic_rnn_test.py
```diff
@@ -0,0 +1,139 @@
...
+
+        seq_lengths_data = np.random.randint(1, seq_length + 1, size=(batch_size,)).astype(np.int32)
+        input_blob_data = np.random.randn(seq_length, batch_size, input_size).astype(np.float32)
+        initial_h_data = np.random.randn(batch_size, hidden_size).astype(np.float32)
+        gates_t_w_data = np.random.randn(hidden_size, hidden_size).astype(np.float32)
+        gates_t_b_data = np.random.randn(hidden_size).astype(np.float32)
+        i2h_w_data = np.random.randn(hidden_size, input_size).astype(np.float32)
+        i2h_b_data = np.random.randn(hidden_size).astype(np.float32)
+
+        with core.DeviceScope(gc):
```
salexspb: could you use assertReferenceChecks? It would free you from doing temp workspace and manual blob feeding. 

This is a common abstraction we normally use for testing operators in caffe2 when using hypothesis. There are several rnn specific examples in rnn_cell_test.

caffe2/python/operator_test/basic_rnn_test.py
```diff
@@ -0,0 +1,139 @@
...
+    if seq_lengths is not None:
+        seq_lengths = (np.ones(shape=(N, D)) *
+                       seq_lengths.reshape(N, 1)).astype(np.int32)
+
+    ret = []
+
+    hidden_prev = hidden_initial
+
+    for t in range(T):
+        input_fc = np.dot(   input[t],  i2h_w.T) +  i2h_b
```
salexspb: nit: formatting seems to be off here :)

salexspb (changes requested):

caffe2/python/rnn_cell.py
```diff
@@ -288,6 +289,27 @@ def _apply(
             raise RuntimeError(
                 'BasicRNNCell with unknown activation function (%s)'
                 % self.activation)
+        if seq_lengths is not None:
+            # TODO If this codepath becomes popular, it may be worth
```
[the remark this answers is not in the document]
anderspapitto: I'd like to get it pushed in this form, and open a task to revisit it in the future. I'm really trying to get https://github.com/pytorch/pytorch/pull/4695 pushed, and the only thing holding it up is these smaller caffe2 diffs.

caffe2/python/operator_test/basic_rnn_test.py
anderspapitto: I can't currently, because assertReferencChecks only works with a single operator, it doesn't know how to compare a multi-op chunk of a graph

caffe2/python/rnn_cell.py
```diff
@@ -288,6 +289,31 @@ def _apply(
...
+            # taking a look at optimizing it - for now a simple
+            # implementation is used to round out compatibility with
+            # ONNX.
+            if scope.CurrentDeviceScope().device_type == caffe2_pb2.CUDA:
```
dzhulgakov: hm, it's quite hacky (and sometimes people move it to GPU after the fact). The problem is that it's literally 1 scalar but you'd need to do a memcopy for it :-\

But in this case you can utilize `CopyFromCPUInput` - it's a magic op which would do the right copy depending on the device you're running on

dzhulgakov (commented):
@salexspb - I think it's fine to land it and have a follow up task (as Anders mentioned it's blocking an outstanding diffs in other repos).

caffe2/python/operator_test/basic_rnn_test.py
salexspb: Did you check rnn_cell_test? It handles this by looking into net.op[-1] for most of the RNNCell types.

salexspb:
Sure, the diff is fine, go ahead if you feel blocked. I am just a bit concerned about test-code consistency with the rest of the RNN logic.

FWIW, using assertReferenceCheck works fine with RNNs by using this technic: 

https://github.com/caffe2/caffe2/blob/3eb636c0edc9337fb348ec554e366b1020604c68/caffe2/python/operator_test/rnn_cell_test.py#L1136

Also, _prepare_rnn method is used by other cell tests

https://github.com/caffe2/caffe2/blob/7d4c7690d64487f71cb08be2a30f32380a11d07a/caffe2/python/rnn/rnn_cell_test_util.py#L35

Among others it allows to easily test multi-layer. With current RNNCell setup it is possible to make a mistake with scoping and break multi-layer. Which I am trying to fix in PR 1681

caffe2/python/operator_test/basic_rnn_test.py
anderspapitto: yes, I did try. The problem is that because I don't have a single C++ op, net.op[-1] is not BasicRNNCell, it's something like one of the constituent FCs

anderspapitto:
@salexspb The first thing I tried was actually to use all the same machinery (prepare_rnn, assertReferenceChecks, etc), and it was giving me a lot of trouble.

caffe2/python/rnn_cell.py
anderspapitto: switched to CopyFromCPUInput

caffe2/python/operator_test/basic_rnn_test.py
salexspb: I don't understand this actually. One of the net's ops has to be RecurrentNetwork operator. Which will have the whole cell inside it. If this is not -1 in your case it can be just another index. You don't get access into the cell itself by navigating net.op

salexspb (approved):

salexspb:
@anderspapitto , let me know if you would like me to walk your through testing tools we use for RNNs in caffe2

salexspb:
Upd, @anderspapitto  and I had a chat and there could've be some missunderstanding.

For future reference - everything comming from RNNCell is just one single operator. Except prepare_input part which is input projection for the whole sequence.

Also there is https://github.com/salexspb/caffe2/blob/rnn_tutorial/caffe2/python/tutorials/Custom_RNNCell.ipynb (WIP) which gives some visualizations of how RNNs in Caffe2 work from python APY point of view. prepare_input is not covered yet. 

Also note that this approach works for comparing a single op, not the whole results of a cell (with prepare_input part). If we want to compare the whole things, one could use network comparison tools. 

Here we compare with unrolled version, this is not strictly nessasary for simple cells. One can manually create a target Caffe2 net and compare it with what cell produced. Unroll automates this process - it unrolles an RNN into the main net and compares it with the net having RecurrentOperator.

Relevant example of usage:

https://github.com/caffe2/caffe2/blob/55a489f6d2bcf43f58abd14643a978fa79c8b52f/caffe2/python/operator_test/rnn_cell_test.py#L958

anderspapitto:
I've started working on improving the test, but I'm going to do it asynchronously while I push this to unblock the related pytorch PR.

I've opened a task to track getting this fixed, https://our.intern.facebook.com/intern/tasks/?t=25797344

And while I'm at it I've also opened https://our.intern.facebook.com/intern/tasks/?t=25797388 to track the perf improvement of a dedicated C++ op, though I don't intend to work on that one immediately

PR #1848, opened 2018-01-30
[experimental] ONNX While Operator

jamesr66a:
Special While loop operator that follows the semantics of While in ONNX: https://github.com/jamesr66a/onnx/blob/controlflow/docs/Operators.md#experimental-loop

Stuff that's missing:

- Lexical scoping enforced via child workspaces
- Double-buffering on forward

Further possible enhancements:
- Full parallelism when there are no loop-carried dependencies
- Diagonal execution
- More optimized scan_outputs shaping via static shape inference provided in ONNX (coming sometime)
- GPU support (probably just some tensor value management stuff)
- Gradient support (likely low-pri right now)

### DIFF
--- a/caffe2/operators/onnx_while_op.cc
+++ b/caffe2/operators/onnx_while_op.cc
@@ -1,0 +1,83 @@
+/**
+ * Copyright (c) 2016-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "caffe2/operators/onnx_while_op.h"
+
+namespace caffe2 {
+
+REGISTER_CPU_OPERATOR(ONNXWhile, ONNXWhileOp<CPUContext>);
+
+OPERATOR_SCHEMA(ONNXWhile)
+    .NumInputs(2, INT_MAX)
+    .NumOutputs(0, INT_MAX)
+    .SetDoc(R"DOC(
+*** EXPERIMENTAL. This operator is a work-in-progress. No assumption should be
+made about the stability or correctness of this op. ***
+
+Generic Looping construct confirming to the ONNX Loop operator spec. This loop
+has multiple termination conditions:
+
+1. Trip count. Iteration count specified at runtime. Set by specifying the
+    input M. Optional. Set to empty string to omit. Note that a static trip
+    count (specified at graph construction time) can be specified by passing
+    in a constant node for input M.
+2. Loop termination condition. This is an input to the op that determines
+    whether to run the first interation and also a loop-carried dependency for
+    the body graph. The body graph must yield a value for the condition
+    variable, whether this input is provided or not.
+
+This table summarizes the operating modes of this operator with equivalent
+C-style code:
+
+Operator inputs defined as (max_trip_count, condition_var). Omitted optional
+inputs are represented as empty string. Concretely, in this caffe2 op an input
+is marked as omitted by setting its 'has_{name}' argument to False.
+
+    input ("", ""):
+        for (int i=0; ; ++i) {
+          cond = ... // Note this value is ignored, but is required in the body
+        }
+
+    input ("", cond) // Note this is analogous to a while loop
+        bool cond = ...;
+        for (int i=0; cond; ++i) {
+          cond = ...;
+        }
+
+    input ("", 1) // Note this is analogous to a do-while loop
+        bool cond = true
+        for (int i=0; cond; ++i) {
+          cond = ...;
+        }
+
+    input (trip_count, "") // Note this is analogous to a for loop
+        int trip_count = ...
+        for (int i=0; i < trip_count; ++i) {
+          cond = ...; // ignored
+        }
+
+    input (trip_count, cond)
+        int trip_count = ...;
+        bool cond = ...;
+        for (int i=0; i < trip_count && cond; ++i) {
+          cond = ...;
+        }
+    )DOC")
+    .Arg("loop_net", "Net executed on each iteration")
+    .Input(0, "condition", "Scalar boolean condition")
+    .AllowInplace([](int in, int out) -> bool { return true; });
+
+} // namespace caffe2
--- a/caffe2/operators/onnx_while_op.h
+++ b/caffe2/operators/onnx_while_op.h
@@ -1,0 +1,233 @@
+/**
+ * Copyright (c) 2016-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef CAFFE2_OPERATORS_ONNX_WHILE_OP_H_
+#define CAFFE2_OPERATORS_ONNX_WHILE_OP_H_
+
+#include "caffe2/core/context.h"
+#include "caffe2/core/logging.h"
+#include "caffe2/core/operator.h"
+
+namespace caffe2 {
+
+template <class Context>
+class ONNXWhileOp final : public Operator<Context> {
+ public:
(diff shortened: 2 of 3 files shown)


caffe2/operators/onnx_while_op.cc
```diff
@@ -0,0 +1,37 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "caffe2/operators/onnx_while_op.h"
+
+namespace caffe2 {
+
+REGISTER_CPU_OPERATOR(ONNXWhile, ONNXWhileOp<CPUContext>);
```
orionr, line 21: Is there a reason why we're putting `ONNX` in the name here? Could we just go with `While` instead?

caffe2/operators/onnx_while_op.cc
```diff
@@ -0,0 +1,37 @@
...
+
+OPERATOR_SCHEMA(ONNXWhile)
+    .NumInputs(2, INT_MAX)
+    .NumOutputs(0, INT_MAX)
+    .SetDoc(R"DOC(
+'While' control operator, first input is a scalar boolean blob that stores loop's
```
orionr: Nice doc - thank you!

jamesr66a: Whoops, this is actually out of date :p

caffe2/operators/onnx_while_op.cc
jamesr66a: This is more of a working name to distinguish this from the existing `While` operator. I wanted to get this in code to make sure the operator semantics for the ONNX spec were sane from a backend point of view. We probably should merge this with the existing operator at some point

caffe2/operators/onnx_while_op.h
```diff
@@ -0,0 +1,224 @@
...
+  //  Inputs: iteration number, condition, loop-carried dependencies
+  //  Outputs: condition, loop-carried dependencies, scan_outputs
+  bool RunOnDevice() override {
+    constexpr int64_t num_inputs_before_lcds = 2;
+    // First input is the maximumt trip count. Second input is the condition
+    // variable (for the first iteration). The rest of the inputs are
+    // loop-carried dependencies.
+    int num_loop_carried_deps = InputSize() - num_inputs_before_lcds;
+    int64_t max_trip_count = *Input(0).template data<int64_t>();
+    const int64_t first_iter_condition = *Input(1).template data<int64_t>();
```
ilia-cher: I assume there's a conscious decision not to use bool?

caffe2/operators/onnx_while_op.h
```diff
@@ -0,0 +1,224 @@
...
+          return (bool)*condition_var_ptr;
+        }
+      } else {
+        return true;
+      }
+    };
+
+    // Allocate scan_outputs for zero-iteration case
+    for (int i = 0; i < num_scan_outputs; ++i) {
+      Output(i + num_loop_carried_deps)->Resize(0);
```
ilia-cher: isn't there a number of iterations first according to the comment above? (then lcd, then scan outputs)

caffe2/operators/onnx_while_op.h
```diff
@@ -0,0 +1,224 @@
...
+      } else {
+        break;
+      }
+      *iteration_var_ptr += 1ll;
+    }
+
+    if (*iteration_var_ptr > 0) {
+      // Copy out final loop-carried dependencies
+      for (int i = 0; i < num_loop_carried_deps; ++i) {
+        Output(i)->CopyFrom(*lcd_tensors_[i]);
```
ilia-cher, line 208: do we output number of iterations first?

caffe2/operators/onnx_while_op.h
```diff
@@ -0,0 +1,224 @@
...
+            void *dst_data =
+                (char *)scan_output_target->raw_mutable_data(sot_meta) +
+                timestep_size * scan_output.itemsize() * itr;
+            memcpy(dst_data, src_data, timestep_size * scan_output.itemsize());
+          }
+        }
+      } else {
+        break;
+      }
+      *iteration_var_ptr += 1ll;
```
ilia-cher, line 201: to confirm - we do not update body net's input condition after each iteration?

ilia-cher (changes requested):
It seems this is not ready yet, because of putting body's local blobs (incl. lcds) into the same workspace as a loop op.

On the higher level, we need to figure how to reconcile this with existing While. Original While op is quite generic and, with some modifications, could be used to implement the same spec. In addition to that, Do operator already supports nested workspaces.

upd: re. 'implement the same spec' - meaning, it would be possible to write a Python function outputting While+Do that does equivalent job, if you need a C2 op with precisely the same spec as in ONNX that's a bit trickier

jamesr66a:
@ilia-cher so one thing i realized this morning was that the existing `While` op has a separate cond_net, whereas the ONNX op takes a condition input and a condition output from the body_net, as requested by partners. This brings up a few issues:

1) If we were to do a translation into `Do + While` we would have to hoist ops that contribute to the condition value out into a separate net. This isn't too bad, but we'd probably have to be careful about if this value is computed based on values that somebody uses later. Partners indicated that this was the case often, so we decided on the combined body net schema
2) Is the current `While` op compatible with all the operating modes described in the ONNX op? In particular the do-while loop. Is this something that the `Do + While` op combination can achieve?
3) Thinking down the road, can we generalize the While op to employ various amounts of parallelism based on op parameters? In particular, if there are no LCD's and we're only using the trip count (i.e. `for (int i=0; i<trip_count; ++i) { <body with no LCDs> }`) we can treat this as essentially a `map()` operation and issue iterations in parallel. This could serve as a powerful runtime meta-programming construct, for e.x. if i want to run N models in parallel, where N is determined at runtime. Another example is diagonal execution, which we've synced up with nvidia about and is a good thing to have

caffe2/operators/onnx_while_op.h
```diff
@@ -0,0 +1,224 @@
...
+        has_trip_count_(
+            OperatorBase::GetSingleArgument<int64_t>("has_trip_count", 0)),
+        has_cond_(OperatorBase::GetSingleArgument<int64_t>("has_cond", 0)) {
+    CAFFE_ENFORCE(this->template HasSingleArgumentOfType<NetDef>("body"),
+                  "body net must be specified in ONNXWhile operator");
+    body_net_def_ = this->template GetSingleArgument<NetDef>("body", NetDef());
+
+    // Create loop-carried deps in Workspace
+    for (int i = 2; i < body_net_def_.external_input_size(); ++i) {
+      Blob *b = ws_->CreateBlob(body_net_def_.external_input()[i]);
```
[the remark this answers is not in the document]
jamesr66a: Yeah, this is the main reason this is marked WIP

caffe2/operators/onnx_while_op.h
jamesr66a: There was some compilation issue i ran into when i put bool here, let me try to repro

caffe2/operators/onnx_while_op.h
jamesr66a: Good catch, that comment above is stale and left-over from before a recent spec change. I'll update it

caffe2/operators/onnx_while_op.h
jamesr66a: The condition variable is a pure LCD (i.e. we don't intervene like here), so we can just treat it with the rest of them. Let me update the code to actually do that

caffe2/operators/onnx_while_op.h
jamesr66a: see above

caffe2/operators/onnx_while_op.h
```diff
@@ -0,0 +1,224 @@
...
+
+namespace caffe2 {
+
+template <class Context> class ONNXWhileOp final : public Operator<Context> {
+public:
+  ONNXWhileOp(const OperatorDef &operator_def, Workspace *ws)
+      : Operator<Context>(operator_def, ws), ws_(ws),
+        has_trip_count_(
+            OperatorBase::GetSingleArgument<int64_t>("has_trip_count", 0)),
+        has_cond_(OperatorBase::GetSingleArgument<int64_t>("has_cond", 0)) {
```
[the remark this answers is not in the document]
jamesr66a: Can you provide a code example of what you mean? This seems like the most direct way to do an optional argument with a default

ilia-cher (approved):

caffe2/operators/onnx_while_op.h
```diff
@@ -0,0 +1,233 @@
...
+            dims.insert(dims.begin(), 1);
+            scan_output_target->Resize(dims);
+            scan_output_target->CopyFrom(scan_output);
+          } else {
+            auto dims = scan_output.dims();
+            CAFFE_ENFORCE_EQ(
+                dims,
+                scan_outputs_sizes[i],
+                "Size of scan output changed across iterations");
+            dims.insert(dims.begin(), itr);
```
artemrakhov-glow, line 182: Is this line a no-op? I don't think dims are ever used after.

caffe2/operators/onnx_while_op.h
```diff
@@ -0,0 +1,233 @@
...
+            scan_output_target->Extend(1, 2.0f, &context_);
```
artemrakhov-glow, line 183: 2.0f is "resize percentage" here. I assume you wanted to do 2x resize here, like vector does? Then it should be 100

artemrakhov-glow (commented):
@jamesr66a sorry if the comments don't make sense. I'm new to Caffe2 codebase, trying to understand the code.

PR #1853, opened 2018-01-31
Fix for glog on windows

willyd:
These changes are required to use glog on Windows.

@Yangqing Please consider merging them as they were removed when PR #1793 was reverted.


Yangqing:
Thanks! That is my fault. Merging.

willyd:
Thanks! If you can have a look at this PR #1854 which fixes an issue with protobuf.

Yangqing:
Yep, thanks for that as well @willyd - this is really helpful as we don't have full windows coverage set up yet. Will merge that one.

PR #1871, opened 2018-02-02
Redefine print for Python 3

Shuolongbj:
Compatible with pyrhon3


caffe2/python/tutorials/Loading_Pretrained_Models.ipynb
```diff
@@ -70,7 +70,7 @@
     "from matplotlib import pyplot\n",
     "import os\n",
     "from caffe2.python import core, workspace, models\n",
-    "import urllib2\n",
+    "import urllib.request\n",
```
orionr, line 73: Thanks for starting this! We'll need backwards compatibility with python 2, so we'll need to follow

http://python-future.org/compatible_idioms.html

For example, you'll need to replace this with

```
from future.standard_library import install_aliases
install_aliases()
import urllib.request
```

Can you also please add this to the very top?

```
from __future__ import unicode_literals
from __future__ import absolute_import
from __future__ import print_function
from __future__ import division
```
Also please test on python2 and python3. Thanks again! This is a great start.

orionr (changes requested):

caffe2/python/tutorials/Loading_Pretrained_Models.ipynb
Shuolongbj: LGTM！thank you

PR #1875, opened 2018-02-03
Allow for all of the names we have in our model zoo.

orionr:
* We now allow subdirectories as well as numbers in the name.
* Also fixed an error case.


orionr:
Also fixes https://github.com/caffe2/caffe2/issues/1882

PR #1880, opened 2018-02-03
Fixed CAFFE2_API decoration for caffe2/proto when building static libraries

willyd:
This was forgotten in #1854.

cc @Yangqing


Yangqing:
Thanks! It was due to a race condition between we landing the code and the PR update. Landing this one now.

Issue #1882, opened 2018-02-04
keyError when install model using"python -m caffe2/python/models/download -i resnet50"


blateyang:
If this is a build issue, please fill out the template below.

### System information

* Operating system: ubuntu16.04
* Compiler version:gcc 5.4.0
* CMake version:
* CMake arguments:
* Relevant libraries/versions (e.g. CUDA):CUDA 8.0

### CMake summary output

```
******** Summary ********
<please paste summary output here>
```
When I use "python -m caffe2.python.models.download -i resnet50" to install resnet50 model, I ran into following problem:
`Traceback (most recent call last):
  File "/usr/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/home/ygj/caffe2/build/caffe2/python/models/download.py", line 176, in <module>
    print("'{model}' is not a valid model name.".format(model))
KeyError: u'model'`
I find it is because there is a small bug in line 157 of  download.py,i.e `if not re.match("^[a-zA-Z_]+$", name):`. '$' should be remove, otherwise the model name with num ending can't be correctly install.

pjh5:
The bug is actually that it should be 
`print("...".format(model=model))`
instead of
`print("...".format(model))`

Thanks for pointing this out. It'll be faster for you to change this in your code, but we'll fix this soon.

pjh5:
Actually, my earlier comment describes another bug. But you are talking about an earlier issue. I think you're right in the bug you describe.

PR #1884, opened 2018-02-04
Fixing download.py regex bug

pjh5:
Fix for https://github.com/caffe2/caffe2/issues/1882


pjh5:
Another PR fixed this

PR #1899, opened 2018-02-06
Multidimensional slice operator

smessmer:
- Multidimensional slice operator for CPU
- The GPU algorithm for slice just calls into the CPU one. This is inperformant and should be replaced with a better GPU algorithm.
- Test cases for multi dimensional slice
- Also replace use of random module in test cases with drawing from hypothesis framework. This should allow for better shrinking.

depends on https://github.com/caffe2/caffe2/pull/1903

### DIFF
--- a/caffe2/operators/slice_op.h
+++ b/caffe2/operators/slice_op.h
@@ -63,10 +63,14 @@ bool SliceImpl(
       if (end < 0) {
         end = data.dims()[i] + 1 + end;
       }
+      if (start > data.dims()[i]) {
+        start = data.dims()[i];
+      }
+      if (end > data.dims()[i]) {
+        end = data.dims()[i];
+      }
       CAFFE_ENFORCE_GE(start, 0);
       CAFFE_ENFORCE_GE(end, 0);
-      CAFFE_ENFORCE_LT(start, data.dims()[i]);
-      CAFFE_ENFORCE_LE(end, data.dims()[i]);
       CAFFE_ENFORCE_GE(end, start);
       starts_idx[i] = start;
       ends_idx[i] = end;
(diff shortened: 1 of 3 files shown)


caffe2/operators/slice_op.h
```diff
@@ -25,12 +25,85 @@ namespace caffe2 {
 
 namespace {
 
+template<class SIndex, class Context>
+void slice_(
```
dzhulgakov, line 29: naming convention is usually with CamelCase and without underscores

caffe2/operators/slice_op.h
```diff
@@ -44,14 +117,28 @@ bool SliceImpl(
   CAFFE_ENFORCE_GE(data.ndim(), starts.size());
   CAFFE_ENFORCE_EQ(starts.size(), ends.size());
 
+  // Compute start and end indices
+  const int last_nonfull_dimension = _last_nonfull_dimension<SIndex>(data.dims(), starts, ends);
```
dzhulgakov, line 121: you probably want to collapse several "full" dimensions in a row - that would be more efficient

also, even though it's correct - move it after the disambiguation of dimensions below? that way it's more understandable.

caffe2/operators/slice_op.h
dzhulgakov, line 29: this is cool. For GPU I think the right model would be instead of recursion doing the iteration in the 'variable number system'. I.e. if you have 3 dimensions [a,b,c] after collapsing, run the loop and do `x = i % a; y = i/a%b; z = i/a/b`. On GPU that's run in parallel and execute super-fast. The current implementation does pretty much the same but only for one dimension.

dzhulgakov (commented):
Great stuff!

orionr:

PR #1919, opened 2018-02-09
Fix typo

ppwwyyxx:


orionr:
Thanks @ppwwyyxx for your correction!

PR #1925, opened 2018-02-09
Update NNPACK submodule to fix build with Python >= 3.6

Maratyszcza:
enum34 dependency of PeachPy conflicts with built-in enum package on Python >= 3.6
This commit brings in NNPACK change to avoid using enum34 on Python >= 3.4


bddppq (approved):
LG

PR #1926, opened 2018-02-09
Experimental support for setup.py develop mode install

bddppq:
`python setup.py develop` / `pip install -e .`


setup.py
```diff
@@ -94,106 +115,130 @@ class build_ext(setuptools.command.build_ext.build_ext):
     to `setup.py build`.  By default all CPUs are used.
     """
     user_options = [
-        ('jobs=', 'j', 'Specifies the number of jobs to use with make')
+        (str('jobs='), str('j'), str('Specifies the number of jobs to use with make'))
     ]
 
+    built = False
```
bddppq, line 122: This is to avoid invoking cmake build more than once.

orionr:
That's great for libcaffe2.so. However, how about for Detectron and GPU? For example, I see the following on my devgpu:

```
libcaffe2_detectron_ops_gpu.so    libcaffe2_module_test_dynamic.so
libcaffe2_gpu.so                  libcaffe2.so
```
Would all of those roll into `caffe2_pybind_state.so`?

bddppq:
Need to fix the python extension suffix first: https://github.com/caffe2/caffe2/pull/1946

PR #1930, opened 2018-02-09
Explicitly pass data on run call for caffe2.ai.

orionr:
* Otherwise we are relying on the input order for the network which might not be consistent.
* Note that this requires that the input be called data, but we use this convention in our model zoo.


houseroad (approved):
Thanks for also fixing these.

PR #1947, opened 2018-02-12
Include newer Python 3 versions in base image builder

pietern:
cc @bddppq


pietern:
@bddppq Confirmed that numpy doesn't work with 3.7 at the moment.

Didn't investigate. Just disabled it in the set of images to build.

PR #1949, opened 2018-02-12
Optionally build with sccache

pietern:
Work in progress to start using sccache


pietern:
No need to reinstall after #1952 is merged.

PR #1935, opened 2018-02-10
Fix cmake function to resolve libraries correctly

yinghai:
Previous behavior may fail to resolve the correct library name

PR #1950, opened 2018-02-13
Fix cmake function to resolve libraries correctly

yinghai:
Previous behavior may fail to resolve the correct library name. A rework of https://github.com/caffe2/caffe2/pull/1935 as it was messed up in the rebase...


yinghai:
@caffe2bot retest this please

PR #1951, opened 2018-02-13
Minor change in tutorial

suryafyi:
The author of the tutorial must have meant the codes of ImageNet classes. Not AlexNet since it is a model, not a dataset.


orionr (approved):
You are exactly right! Thank you for this. I'll be bringing this in shortly.

PR #1953, opened 2018-02-13
Skip system Python if Anaconda is used

pietern:
We don't care about a particular system Python when building Anaconda images.

Rebasing later to remove the sccache change once it is merged (#1952).

### DIFF
--- a/docker/jenkins/build.sh
+++ b/docker/jenkins/build.sh
@@ -23,10 +23,10 @@ valid_images=(
   py2-android-ubuntu16.04
 
   # Builds for Anaconda
-  py2-conda-ubuntu16.04
-  py2-conda-cuda9.0-cudnn7-ubuntu16.04
-  py3-conda-ubuntu16.04
-  py3-conda-cuda9.0-cudnn7-ubuntu16.04
+  conda2-ubuntu16.04
+  conda2-cuda9.0-cudnn7-ubuntu16.04
+  conda3-ubuntu16.04
+  conda3-cuda9.0-cudnn7-ubuntu16.04
 )
 
 image="$1"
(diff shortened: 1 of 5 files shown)


docker/jenkins/build.sh
```diff
@@ -23,10 +23,10 @@ valid_images=(
...
 
   # Builds for Anaconda
-  py2-conda-ubuntu16.04
-  py2-conda-cuda9.0-cudnn7-ubuntu16.04
-  py3-conda-ubuntu16.04
-  py3-conda-cuda9.0-cudnn7-ubuntu16.04
+  conda2-ubuntu16.04
+  conda2-cuda9.0-cudnn7-ubuntu16.04
+  conda3-ubuntu16.04
+  conda3-cuda9.0-cudnn7-ubuntu16.04
```
pjh5, line 29: Are these actually used anywhere? I think we can just delete them

docker/jenkins/centos-cuda/Dockerfile
```diff
@@ -15,7 +15,8 @@ RUN bash ./install_ccache.sh && rm install_ccache.sh
 # Install Python
 ARG PYTHON_VERSION
 ADD ./install_python.sh install_python.sh
-RUN bash ./install_python.sh && rm install_python.sh
+RUN if [ -n "${PYTHON_VERSION}" ]; then bash ./install_python.sh; fi
+RUN rm install_python.sh
```
pjh5, line 19: We'll only really need these if we also make conda builds for centos, or if we start having other centos builds without python. I don't have plans for that right now

pjh5 (approved):
A few changes that probably aren't necessary but otherwise LGTM

docker/jenkins/centos-cuda/Dockerfile
pietern: You're right. Changed everything to keep it in sync.

docker/jenkins/build.sh
pietern: They aren't, but can serve as documentation.

PR #1957, opened 2018-02-14
Fix Android build with binaries

Maratyszcza:
After we removed android-cmake submodule and switched to android.cmake.toolchain from Android NDK, the code that builds cpufeatures dependency is no longer valid. This commit fixes it.


Maratyszcza:
@caffe2bot retest this please

sf-wind (approved):
Let's ship it.

Maratyszcza:
@caffe2bot retest this please

PR #1962, opened 2018-02-14
Disable support of using ninja in setup.py

bddppq:
Our cmake files have some issue when using using ninja as the generator to build with cuda


Maratyszcza:
LGTM

Maratyszcza (approved):

PR #1966, opened 2018-02-15
Add Cmake dependencies for NVTX

syed-ahmed:
Summary:
- Commit c9c24d5 adds the cmake stuff necessary for compiling with USE_NVTX flag. [convnet_benchmarks.py](https://github.com/caffe2/caffe2/blob/master/caffe2/python/convnet_benchmarks.py) is  the only front-end file that utilizes the `--caffe2_use_nvtx` flag. `AsyncDAGNet` on GPU is currently using `ProfiledRanges` that utilizes this flag (CAFFE2_USE_NVTX). And somehow cmake didn't have the necessary dependencies to build caffe2 with nvtx. Hence, this commit fixes that issue (can be cherry picked).

- Commit  7ee4731 is what I had to do to get markers and ranges for a simple convnet. I'm including it in this PR just for reference and it is not suitable for a merge. I am just wondering if `nvprof` will be supported and if yes, `ProfiledRanges` could be abstracted in a better way. (May be there is already plans for it in `contrib/prof`?


Yangqing:
@pietern  would you mind taking a look?

orionr:

PR #1968, opened 2018-02-15
Fix setup.py

yinghai:
There is a typo in the setup.py.


bddppq:
I don't understand the part "cause incomplete install". This flag is to turn off building benchmarks, if the flag was not working then it should build more stuffs?

yinghai:
Agree. My observation is due to change of behavior of setup.py, not relevant of this typo.

PR #1970, opened 2018-02-16
Fix warning in net_test

Maratyszcza:
Fixes an annoying warning when building for Android with tests enabled.


Maratyszcza:
@caffe2bot retest this please

PR #1921, opened 2018-02-09
Move onnx-caffe2 inside caffe2

yinghai:
This PR moves all the python source code from https://github.com/onnx/onnx-caffe2 to caffe2. It helps stage follow-up PRs to have C++ onnx-caffe2 backend and control the size of the diff.  There will a couple of PRs after this one:
- Pull onnx into third-party and cmake support
- C++ onnx-caffe2 backend with Python wrapper to replace majority functions of the backend.py.

PR #1974, opened 2018-02-16
Update onnx version for onnx-caffe2 test

yinghai:
We recently added `onnx.optimize` from onnx-caffe2 to onnx. So we need a newer version of onnx to run the tests add in https://github.com/caffe2/caffe2/pull/1921.


bddppq (approved):
@pietern

PR #1976, opened 2018-02-16
Remove gf library in MKL

sf-wind:
Do not use gf library in MKL. Even after I install the entire MKL library it is still not found. After removing it, the MKL code can still run. I'm not sure why that library is included. 

Does it make sense?

Test Plan: Build Caffe2 with MKL linked statically. using the following flag:

scripts/build_local.sh -DBUILD_BINARY=ON -DBUILD_SHARE_DIR=ON -DBUILD_OBSERVERS=ON -DUSE_ZSTD=ON -DBLAS=MKL -DMKL_USE_STATIC_LIBS=ON -DMKL_USE_SINGLE_DYNAMIC_LIBRARY=OFF

Run caffe2_benchmark with MKL and some models can run successfully


ajtulloch:
sgtm

pietern:
MKL build is passing. Merging.

PR #1979, opened 2018-02-16
Allow more backends in caffe2_benchmark

sf-wind:
Test plan: run the MKL backend and the new backend is picked up.


caffe2/share/contrib/binaries/caffe2_benchmark/caffe2_benchmark.cc
```diff
@@ -13,9 +13,9 @@
 
 CAFFE2_DEFINE_string(
     backend,
-    "default",
+    "builtin",
```
bwasti, line 16: how is builtin different from default?

caffe2/share/contrib/binaries/caffe2_benchmark/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 set(__target "caffe2_benchmark")
 add_executable(${__target} "${CMAKE_CURRENT_SOURCE_DIR}/caffe2_benchmark.cc")
 add_dependencies(${__target} ${Caffe2_MAIN_LIBS_ORDER})
-caffe_add_whole_archive_flag(Caffe2_CPU_OBSERVER Caffe2_CPU_OBSERVER_LINK)
+caffe_add_linker_flag(Caffe2_CPU_OBSERVER Caffe2_CPU_OBSERVER_LINK)
```
bwasti, line 4: does add_linker_flag preserve whole_archive?

sf-wind: I will use -DBUILD_SHARED_LIBS=OFF so that only static library is built for benchmarking purpose. Otherwise it uses dynamic library, which is fine in most cases.

caffe2/share/contrib/binaries/caffe2_benchmark/caffe2_benchmark.cc
sf-wind: Default is basically the default backend, with engine field to be "". Builtin is the builtin backend baked in the model.

bwasti (approved):

Issue #1543, opened 2017-11-30
caffe2_pybind11_state.so:  undefined symbol: _ZN6caffe26script15CompilationUnit6defineERKSs


xiaobogaga:
Build successfully, but when run with:

```
python -c "from caffe2.python import core" 2>/dev/null && echo "success" || echo failure
```

shows error:

```
WARNING:root:This caffe2 python run does not have GPU support. Will run in CPU only mode.
WARNING:root:Debug message: No module named caffe2_pybind11_state_gpu
CRITICAL:root:Cannot load caffe2.python. Error: caffe2/python/caffe2_pybind11_state.so: 
undefined symbol: _ZN6caffe26script15CompilationUnit6defineERKSs
```

I don't know why, although I find little same issue like #197, #346, their solution seems not useful.
help wanted...

platform information:
* centos 7
* python 2.7.5
* gcc 4.8.5

landscape-ooo:
@pietern @Maratyszcza 
` python -m caffe2.python.operator_test.relu_op_test
`

```
WARNING:root:This caffe2 python run does not have GPU support. Will run in CPU only mode.
WARNING:root:Debug message: /home/work/caffe2_bin/caffe2/python/caffe2_pybind11_state_gpu.so: undefined symbol: _ZN6caffe26script15CompilationUnit6defineERKSs
CRITICAL:root:Cannot load caffe2.python. Error: /home/work/caffe2_bin/caffe2/python/caffe2_pybind11_state.so: undefined symbol: _ZN6caffe26script15CompilationUnit6defineERKSs

```
platform information:

- centos 7
- python 2.7.14 （virutabl_env --no-site-packages ）then (pip install -r  req.txt) where req.txt  is copyfrom   [req.txt](https://caffe2.ai/docs/getting-started.html?platform=centos&configuration=cloud#null__python-dependencies)
- gcc 4.8.5
- cuda 8.0.61  cudnn 6.0.21
- cmake 3.10 
======cmake config 
 `
[...]

PR #1987, opened 2018-02-19
Make sure libcaffe2.so does not require executable stack

Maratyszcza:
This commit updates python-peachpy submodule to bring in the fix.

In #1543 @samarjeet reported that importing caffe2 from Python fails on his system with the error "CRITICAL:root:Cannot load caffe2.python. Error: libcaffe2.so: cannot enable executable stack as shared object requires: Invalid argument". I investigated and found that this is caused by libcaffe2.so being marked as requiring executable stack, which itself was caused by assembly (PeachPy) files in NNPACK not specifying whether they need an executable stack (by default, linker assumes execstack needed). I patched PeachPy to add ".note.GNU-stack" section to generated ELF files, which makes the linker mark libcaffe2.so as NOT needing executable stack. See Maratyszcza/PeachPy#89 for details.


Yangqing:
Looks good. Test failure seems transient.

PR #1993, opened 2018-02-21
wip: cuda targets

Yangqing:
(this is in order to modernize the cuda targets and allow caffe2_gpu library to be relocatable with cmake. Not sure if this builds yet. my local mac passes cmake configuration but not sure about the rest.)


pjh5 (approved):
Do you know if we should ever also set compiler flags like this example
"
  set_target_properties(Foo::Foo PROPERTIES
    INTERFACE_COMPILE_OPTIONS "${PC_Foo_CFLAGS_OTHER}"
    INTERFACE_INCLUDE_DIRECTORIES "${Foo_INCLUDE_DIR}"
  )
" on https://cmake.org/cmake/help/v3.5/manual/cmake-developer.7.html#find-modules ?

Yangqing:
@pjh5 yep. For compiler flags that should be used in host compilers, we should totally set so. In this case for the NVCC flags though, it is added to nvcc compilers so it needs to be custom.

Yangqing:
I am almost sure that the test errors are not due to this change (it is purely CMake), but I'll hold on till they get resolved first.

pjh5:
@Yangqing these test failures have come up sporadically for a while now. I suggest you do not wait for them

pietern:
`blobs_queue_db_test.py::BlobsQueueDBTest::test_create_blobs_queue_db_string` is hanging for the CUDA builds, but this has been the case on master for a while... This should be fixed ASAP but is unrelated to your changes here.

Yangqing:
Sounds great, thanks guys - landing.

pjh5:
Sweet

PR #1997, opened 2018-02-21
Adding guards around adding protobuf targets

pjh5:


cmake/public/protobuf.cmake
```diff
@@ -18,7 +18,9 @@ elseif(Protobuf_FOUND OR PROTOBUF_FOUND)
   # content.
   message(STATUS "Caffe2: Found protobuf with old-style protobuf targets.")
   if(PROTOBUF_LIBRARY)
-    add_library(protobuf::libprotobuf UNKNOWN IMPORTED)
+    if (NOT TARGET protobuf::libprotobuf)
+      add_library(protobuf::libprotobuf UNKNOWN IMPORTED)
+    endif()
     set_target_properties(protobuf::libprotobuf PROPERTIES
```
Yangqing: You might want to put this inside the if statement as well?

Yangqing: (same for below)

yinghai:
This test seems to be failing consistently: 
```
20:39:52 lib/python2.7/dist-packages/caffe2/python/operator_test/blobs_queue_db_test.py::BlobsQueueDBTest::test_create_blobs_queue_db_string Build timed out (after 30 minutes).
```

pjh5:
@yinghai this is a known problem that we've been having for a while. They seemed to appear after we bumped the docker versions to 43.

yinghai:
@pjh5 Were you able to reproduce this with the dockerfile? Because I was not able to reproduce...

pjh5:
@yinghai these failures are not consistent, and we don't know their cause yet. But I am quite certain that they are unrelated to this change

orionr (approved):
So long as no more tests fail (we need to get those tests passing again!) and things look good on Mac and conda I think you're good.

PR #1998, opened 2018-02-21
Add onnx as a submodule

yinghai:
Adding onnx as third_party for better integration, especially on C++ level. 

ONNX version is pinned at https://github.com/onnx/onnx/commit/c894dd753eab0d14c8a512dd21501ac932dc8040.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -61,3 +61,6 @@
 [submodule "third_party/python-six"]
 	path = third_party/python-six
 	url = https://github.com/benjaminp/six.git
+[submodule "third_party/onnx"]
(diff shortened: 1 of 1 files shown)


bddppq:
Maybe also change to install the onnx package from submodule instead of from github? Note since we are going to bump the onnx version quite often, onnx installation should not be embed in the docker images.

yinghai:
Makes sense. I'll do that!

bddppq (approved):

yinghai:
After some discussion, I'll do the pip install change in a separate PR to keep the change localized.

PR #2000, opened 2018-02-22
Support convolution without bias in NNPACK bindings

Maratyszcza:
NNPACK interface always requires bias. If a convolution doesn't have one, Caffe2 should create a dummy bias. This problem pops up in models converted from PyTorch through ONNX, because by default they convert PyTorch convolution into Conv (without bias) + Add


jonmorton (approved):
tyty

Maratyszcza:
@caffe2bot retest this please

Maratyszcza:
@caffe2bot retest this please

PR #2004, opened 2018-02-22
Fixing an error building with CUDA on windows

pjh5:


Yangqing:
Actually, with #1993 this should not be needed any more. Should I land that diff?

pjh5:
I think this may still be necessary for Windows. I still get the same error as before (lots of unresolved external symbols in cuda and cuDNN libraries) after rebasing onto #1993.

cmake/Utils.cmake
```diff
@@ -244,7 +244,11 @@ function(caffe2_binary_target target_name_or_src)
   endif()
   add_executable(${__target} ${__srcs})
   add_dependencies(${__target} ${Caffe2_MAIN_LIBS_ORDER})
-  target_link_libraries(${__target} ${Caffe2_MAIN_LIBS} ${Caffe2_DEPENDENCY_LIBS})
+  if (USE_CUDA)
```
Yangqing, line 247: I assume this is no longer needed with the interface library we have?

pjh5: It actually was on Windows

Yangqing: nvm, saw your note above.

Yangqing:
LGTM. Test failure inspected separately. Merging.

PR #2008, opened 2018-02-22
added debug info to OperatorDef

bwasti:
this might be useful for net construction debugging.

### DIFF
--- a/caffe2/python/core.py
+++ b/caffe2/python/core.py
@@ -35,6 +35,7 @@ import caffe2.python._import_c_extension as C
 import pickle
 import numpy as np
 import sys
+import traceback
 
 # Mac os specific message
 if (sys.platform == 'darwin' and 'leveldb' in C.registered_dbs()):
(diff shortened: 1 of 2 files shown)


xiaomengy:
@caffe2bot retest this please

bddppq:
onnx has a similar field called [doc_string](https://github.com/onnx/onnx/blob/5641b4abd364d32710ec19d22f3ae321282fda9b/onnx/onnx.proto#L169). I think debug_info is a more appropriate name, but to align with onnx maybe we should use doc_string?

caffe2/python/core.py
```diff
@@ -321,6 +322,12 @@ def CreateOperator(
     registered with Caffe2.
     """
     operator = caffe2_pb2.OperatorDef()
+    stack = traceback.extract_stack()
+    debug_info = ""
+    for frame in stack[:-1]:
+      debug_info += "{}:{} {} -> {}\n".format(*frame)
```
bddppq: https://docs.python.org/2/library/traceback.html#traceback.format_exc?

bwasti:
`doc_string` is a very confusing name for a stacktrace

bddppq (approved):

bwasti:
fix is here :) https://github.com/caffe2/caffe2/pull/2013

PR #2014, opened 2018-02-23
Relax comparing ops in gradient tests

bddppq:
Fix CI breakage introduced in https://github.com/caffe2/caffe2/pull/2008


bddppq:
Oops crossed fire with https://github.com/caffe2/caffe2/pull/2013

PR #2015, opened 2018-02-23
Caffe2 ARM ComputeLibrary integration

jerryzh168:


CMakeLists.txt
```diff
@@ -25,6 +25,7 @@ cmake_dependent_option(
     "NOT BUILD_SHARED_LIBS" OFF)
 option(BUILD_OBSERVERS "Build performance observers/loggers in caffe2/share/observers directory" OFF)
 option(BUILD_TEST "Build C++ test binaries (need gtest and gbenchmark)" ON)
+option(USE_ARM_COMPUTE "Use ARM Compute Library" ON)
```
bwasti: lets turn this off by default, as it has a scons dependency at the moment

caffe2/mobile/contrib/CMakeLists.txt
```diff
@@ -1,5 +1,8 @@
 add_subdirectory(ios)
 add_subdirectory(opengl)
+if (USE_ARM_COMPUTE AND ANDROID)
+	add_subdirectory(arm-compute)
```
Maratyszcza: Please convert to spaces

caffe2/mobile/contrib/arm-compute/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+if(USE_ARM_COMPUTE AND ANDROID)
```
Maratyszcza: Typically we just disable options on unsupported platforms, e.g. see https://github.com/caffe2/caffe2/blob/1f9e79f920a22aed1a370bf5faec1ce5e1ef73d7/cmake/Dependencies.cmake#L460-L465

.gitmodules
```diff
@@ -61,6 +61,9 @@
 [submodule "third_party/python-six"]
 	path = third_party/python-six
 	url = https://github.com/benjaminp/six.git
+[submodule "third_party/ComputeLibrary"]
+	path = third_party/ComputeLibrary
+	url = https://github.com/ARM-software/ComputeLibrary
```
bwasti: `.git` for consistency

caffe2/mobile/contrib/arm-compute/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Build
+
+To build, clone and install scons
+
+```
+brew install scons
+```
+
+use the build\_android.sh:
```
Maratyszcza: typo?

scripts/build_android.sh
```diff
@@ -52,24 +52,35 @@ CMAKE_ARGS=()
...
 CMAKE_ARGS+=("-DBUILD_SHARED_LIBS=OFF")
+if $USE_ARM_COMPUTE;then
+CMAKE_ARGS+=("-DANDROID_TOOLCHAIN_NAME=arm-linux-androideabi-4.9")
+else
 CMAKE_ARGS+=("-DANDROID_TOOLCHAIN=gcc")
+fi
 
 # Disable unused dependencies
 CMAKE_ARGS+=("-DUSE_CUDA=OFF")
-CMAKE_ARGS+=("-DUSE_GFLAGS=OFF")
```
bwasti: can you add this back?

scripts/build_android.sh
```diff
@@ -52,24 +52,35 @@ CMAKE_ARGS=()
...
+  if [ $arg == "-DUSE_ARM_COMPUTE=ON" ];then
+    USE_ARM_COMPUTE=true
+  fi
+done
 # Use android-cmake to build Android project from CMake.
 CMAKE_ARGS+=("-DCMAKE_TOOLCHAIN_FILE=$ANDROID_NDK/build/cmake/android.toolchain.cmake")
 
 # Don't build artifacts we don't need
-CMAKE_ARGS+=("-DBUILD_TEST=OFF")
+CMAKE_ARGS+=("-DBUILD_TEST=ON")
```
bwasti: I think this is on by default, either way the comment above doesn't make sense anymore

scripts/build_android.sh
```diff
@@ -52,24 +52,35 @@ CMAKE_ARGS=()
 # target architecture and need an exact version match.
 CMAKE_ARGS+=("-DCAFFE2_CUSTOM_PROTOC_EXECUTABLE=$CAFFE2_ROOT/build_host_protoc/bin/protoc")
 
+CMAKE_ARGS+=($@)
+USE_ARM_COMPUTE=false
+for arg in ${CMAKE_ARGS[@]};do
+  if [ $arg == "-DUSE_ARM_COMPUTE=ON" ];then
```
bwasti: there are a couple other ways to set things "true" in cmake, it might be good to check for all of those

scripts/build_android.sh
```diff
@@ -78,36 +89,19 @@ fi
...
-if $USE_GCC;then
-  CMAKE_ARGS+=("-DCMAKE_CXX_FLAGS=-s")
-fi
+CMAKE_ARGS+=("-DCMAKE_CXX_FLAGS=-s")
 
 cmake "$CAFFE2_ROOT" \
     -DCMAKE_INSTALL_PREFIX=../install \
     -DCMAKE_BUILD_TYPE=Release \
-    "${CMAKE_ARGS[@]}"
+    -G Ninja \
```
bwasti: this should not be a default

caffe2/mobile/contrib/arm-compute/core/operator.cc
```diff
@@ -0,0 +1,9 @@
+#include "operator.h"
+
+namespace caffe2 {
+
+CAFFE_DEFINE_REGISTRY(GLOperatorRegistry, OperatorBase, const OperatorDef &,
+                      Workspace *);
+CAFFE_REGISTER_DEVICE_TYPE(DeviceType::OPENGL, GLOperatorRegistry);
```
Maratyszcza, line 7: This probably collides with our OpenGL ES 3.0 backend

scripts/build_android.sh
```diff
@@ -78,36 +89,19 @@ fi
...
+    "${CMAKE_ARGS[@]}" \
+    "$@"
 
 # Cross-platform parallel build
-if [ "$(uname)" == "Darwin" ]; then
-  cmake --build . -- "-j$(sysctl -n hw.ncpu)"
-else
-  cmake --build . -- "-j$(nproc)"
-fi
+ninja
```
bwasti: lets leave this bit

caffe2/CMakeLists.txt
```diff
@@ -10,6 +10,7 @@ set(Caffe2_GPU_SRCS)
 # files for CPU and GPU tests respectively.
 set(Caffe2_CPU_TEST_SRCS)
 set(Caffe2_GPU_TEST_SRCS)
+set(Caffe2_MOBILE_TEST_SRCS)
```
Maratyszcza: Why do we need `Caffe2_MOBILE_TEST_SRCS`?  IIUC, `Caffe2_GPU_TEST_SRCS` exist because it is built with a different compiler

scripts/build_android.sh
Maratyszcza: I like the idea of switching to `ninja`, but lets leave it for a separate PR

Maratyszcza:
Could you keep changes to CMakeLists to a minimum?

caffe2/mobile/contrib/arm-compute/core/operator.cc
jerryzh168: Is OpenGL ES 3.0 backend removed already?

caffe2/mobile/contrib/arm-compute/README.md
jerryzh168: Which word? build\_android.sh? that is probably for markdown..

caffe2/mobile/contrib/arm-compute/core/operator.cc
Maratyszcza: No, its still in the source tree, and potentially even being built.

caffe2/CMakeLists.txt
jerryzh168: So that we can build only those tests when we build on mobile? I'm not sure, @bwasti what do you think?

caffe2/mobile/contrib/arm-compute/core/operator.cc
jerryzh168: Should we change it to something else? I think DeviceType::OPENGL is a pre-defined device type.

scripts/build_android.sh
jerryzh168: -DUSE_ARM_COMPUTE=1? what else?

scripts/build_android.sh
jerryzh168: Should we turn on the test by default?

jerryzh168:
@Maratyszcza @bwasti could you review again? it would be great if we can merge this today

caffe2/CMakeLists.txt
Maratyszcza: cc @Yangqing as well

scripts/build_android.sh
Maratyszcza: Better keep it off. When needed, tests can be built with `scripts/build_android.sh -DBUILD_TEST=ON`

caffe2/mobile/contrib/arm-compute/core/operator.cc
Maratyszcza: cc @hlu1 to check if this can conflict with OpenGL ES 3.0

caffe2/CMakeLists.txt
```diff
@@ -169,7 +182,7 @@ if (BUILD_TEST)
           ${test_name} ${Caffe2_MAIN_LIBS} ${Caffe2_DEPENDENCY_LIBS}
           gtest_main)
     endif()
-    if (${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION} GREATER 3.0)
+    if (${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION} GREATER 3.0 AND NOT ANDROID)
```
Maratyszcza: Why the change? The sources still use `cxx_range_for` when building for android. `target_compile_features` tells CMake to set appropriate flags. Without it, compiler may default to c++98.

jerryzh168: `target_compile_features` is not recognized before for some reason.. resolved now.

jerryzh168:
remaining issue `device_type` and `Caffe2_MOBILE_TEST_SRCS`, but I think we can modify these in another diff as long as it doesn't break the ci?

caffe2/CMakeLists.txt
bwasti: yea, we can chuck everything into CPU_TEST_SRCS.  semantically, we could rename them to CC_SRCS and NVCC_SRCS huh?

caffe2/mobile/contrib/arm-compute/core/operator.cc
bwasti:
```
#ifdef CAFFE2_OPENGL_BACKEND
#error Can only build one OpenGL backend at a time.
#else
#define CAFFE2_OPENGL_BACKEND
#endif
```

CMakeLists.txt
```diff
@@ -25,6 +25,7 @@ cmake_dependent_option(
     "NOT BUILD_SHARED_LIBS" OFF)
 option(BUILD_OBSERVERS "Build performance observers/loggers in caffe2/share/observers directory" OFF)
 option(BUILD_TEST "Build C++ test binaries (need gtest and gbenchmark)" ON)
+option(USE_ARM_COMPUTE "Use ARM Compute Library" OFF)
```
Maratyszcza: The name is ambiguous: users may think they have to enable it for any ARM builds. Please change to `USE_ACL`, `USE_ARMCL`, or `USE_ARM_COMPUTE_LIBRARY`

scripts/build_android.sh
```diff
@@ -52,15 +52,27 @@ CMAKE_ARGS=()
...
 
 # Don't build artifacts we don't need
-CMAKE_ARGS+=("-DBUILD_TEST=OFF")
+CMAKE_ARGS+=("-DBUILD_TEST=ON")
 CMAKE_ARGS+=("-DBUILD_BINARY=OFF")
 CMAKE_ARGS+=("-DBUILD_PYTHON=OFF")
 CMAKE_ARGS+=("-DBUILD_SHARED_LIBS=OFF")
+
+if $USE_ARM_COMPUTE;then
+CMAKE_ARGS+=("-DANDROID_TOOLCHAIN_NAME=arm-linux-androideabi-4.9")
```
Maratyszcza: This line have the same effect as `-DANDROID_TOOLCHAIN=gcc`. See https://android.googlesource.com/platform/ndk/+/master/build/cmake/android.toolchain.cmake#103

scripts/build_android.sh
```diff
@@ -103,7 +115,7 @@ fi
 cmake "$CAFFE2_ROOT" \
     -DCMAKE_INSTALL_PREFIX=../install \
     -DCMAKE_BUILD_TYPE=Release \
-    "${CMAKE_ARGS[@]}"
+    "${CMAKE_ARGS[@]}" \
```
Maratyszcza: Revert?

caffe2/CMakeLists.txt
Maratyszcza: agree, `CC_SRCS` + `NVCC_SRCS` would make more sense

scripts/build_android.sh
```diff
@@ -52,6 +52,13 @@ CMAKE_ARGS=()
 # target architecture and need an exact version match.
 CMAKE_ARGS+=("-DCAFFE2_CUSTOM_PROTOC_EXECUTABLE=$CAFFE2_ROOT/build_host_protoc/bin/protoc")
 
+CMAKE_ARGS+=($@)
+USE_ARM_COMPUTE=false
```
Maratyszcza: Please remove this snippet, its not needed

cmake/Dependencies.cmake
```diff
@@ -444,6 +444,43 @@ if (USE_MOBILE_OPENGL)
...
+  list(APPEND ARM_COMPUTE_INCLUDE_DIRS "third_party/ComputeLibrary/")
+  list(APPEND ARM_COMPUTE_INCLUDE_DIRS "third_party/ComputeLibrary/include")
+  caffe2_include_directories(${ARM_COMPUTE_INCLUDE_DIRS})
+  string (REPLACE ";" " -I" ANDROID_STL_INCLUDE_FLAGS "-I${ANDROID_STL_INCLUDE_DIRS}")
+  set (ARM_COMPUTE_SRC_DIR "${PROJECT_SOURCE_DIR}/third_party/ComputeLibrary/")
+  set (ARM_COMPUTE_LIB "${CMAKE_CURRENT_BINARY_DIR}/libarm_compute.a")
+  set (ARM_COMPUTE_CORE_LIB "${CMAKE_CURRENT_BINARY_DIR}/libarm_compute_core.a")
+  set (ARM_COMPUTE_LIBS ${ARM_COMPUTE_LIB} ${ARM_COMPUTE_CORE_LIB})
+
+  add_custom_command(
```
Maratyszcza, line 470: The custom command doesn't have `DEPENDS`, so it won't rebuild the sources e.g. after a submodule update. Hopefully, ARM CL would just add CMakeLists, and we don't have to monkey-patch it.

cmake/Dependencies.cmake
```diff
@@ -464,6 +501,14 @@ if (USE_METAL)
   endif()
 endif()
 
+
+if (USE_ARM_COMPUTE)
```
Maratyszcza: This should be above the code that imports arm compute library targets from Scons. The motivation is to check only `if (USE_ARM_COMPUTE)`, so if we need to enable this backend on other platforms, we need to patch only one snippet.

Maratyszcza (approved):
Please do the following before merging:
- Remove `Caffe2_MOBILE_TEST_SRCS`, use `Caffe2_CPU_TEST_SRCS`
- Remove code that sets and uses `USE_ARM_COMPUTE` from `scripts/build_android.sh`
- Replace `if (USE_ARM_COMPUTE AND ANDROID)` with `if (USE_ARM_COMPUTE)`. Only the snippet that disables ARMCL on non-Android platforms should check the platform.
- Rename `USE_ARM_COMPUTE` option to something less ambiguous. Current name may suggest that this option is required for all builds for ARM.
- Check that is builds with ARM Compute Library for ARMv7 and ARM64, and with -DBUILD_TEST=ON as well.

I didn't look much at C++ code, assume @bwasti understands it.

caffe2/mobile/contrib/arm-compute/core/operator.cc
hlu1: OpenGL ES 3.0 backend does not use DeviceType. It prefixes operator names with OpenGL instead

caffe2/mobile/contrib/arm-compute/core/context.h
```diff
@@ -0,0 +1,324 @@
...
+#include "arm_compute/runtime/BlobLifetimeManager.h"
+#include "arm_compute/runtime/MemoryManagerOnDemand.h"
+#include "arm_compute/runtime/PoolManager.h"
+#include "utils/Utils.h"
+#include "include/half/half.hpp"
+
+namespace caffe2 {
+
+typedef half_float::half half;
+typedef half DataType;
```
hlu1, line 32: Do you need to check if float16 is supported on device?

bwasti (approved):
Let's rename the PR to be a bit more informative?

PR #2020, opened 2018-02-23
Remove build_host_protoc.bat

Yangqing:
It is no longer needed after protobuf cmake update - when we are not doing cross compilation, Windows can directly use the binary built with build_windows.bat.


pjh5 (approved):

Yangqing:

PR #2021, opened 2018-02-23
Update CMake min requirement to 3, and use interface library for cuda libs.

Yangqing:
Firing a diff to see if this actually works across platforms.


Yangqing:
Interesting finding. This only works with cmake3 because INTERFACE library is not available in 2.8.

Yangqing:
@pietern Maybe we should enforce cmake 3 :)

pjh5 (approved):

Yangqing:

Yangqing:
Ubuntu cmake3 install is pending a docker build trigger. Rest is passing properly (per centos7 build).

pietern:
:heart:

PR #2025, opened 2018-02-23
Scope MultiRNN blobs with name as well as layers

pooyadavoodi:
Currently state names are scoped only by layer number. This PR adds self.name to the scope, so it becomes like `name/layer_1/state_name`.

Also don't double scope MultiRNN in case of multiple layers if using `rnn_cell.LSTM`.
Currently the scope is applied to both LSTMCell and MultiRNNCell which causes to have scopes like this: `name/layer_1/name/blobname`. This PR changes that to `name/layer_1/blobname`.

### DIFF
--- a/caffe2/python/memonger_test.py
+++ b/caffe2/python/memonger_test.py
@@ -483,10 +483,10 @@ class MemongerTest(hu.HypothesisTestCase):
                 "cell_init_{}".format(i)
             )
             init_blobs.extend([hidden_init, cell_init])
-        model.param_init_net.ConstantFill([], ["input"], shape=[T, 4, 10])
+        input_blob = model.param_init_net.ConstantFill([], ["input"], shape=[T, 4, 10])
         output, last_hidden, _, last_state = rnn_cell.LSTM(
             model=model,
-            input_blob="input",
+            input_blob=input_blob,
             seq_lengths=seq_lengths,
             initial_states=init_blobs,
             dim_in=10,
--- a/caffe2/python/memonger_test.py
+++ b/caffe2/python/memonger_test.py
@@ -483,10 +483,10 @@ class MemongerTest(hu.HypothesisTestCase):
                 "cell_init_{}".format(i)
             )
             init_blobs.extend([hidden_init, cell_init])
-        input_blob = model.param_init_net.ConstantFill([], ["input"], shape=[T, 4, 10])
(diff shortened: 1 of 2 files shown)


caffe2/python/memonger_test.py
```diff
@@ -483,10 +483,10 @@ def test_rnn(self):
                 "cell_init_{}".format(i)
             )
             init_blobs.extend([hidden_init, cell_init])
-        model.param_init_net.ConstantFill([], ["input"], shape=[T, 4, 10])
+        input_blob = model.param_init_net.ConstantFill([], ["input"], shape=[T, 4, 10])
```
salexspb: why do you have to change it? Did it stop working with string blob names?

salexspb (commented):
The RNNCell change makes sense to me. Could you share an  example how state names looked like before and how they look now? Also I am not sure why the memonger RNN test started failing. Could you elaborate on the test change, please?

pooyadavoodi:
If we call `rnn_cell.LSTM` with `scope='test'` (like in `RNNCellTest.test_multi_lstm`, then some of the blobs change by this PR as follows:
`test/layer_0/test/initial_hidden_state` --> `test/layer_0/initial_hidden_state`
`test/layer_1/test/i2h_w` --> `test/layer_1/i2h_w`

caffe2/python/memonger_test.py
pooyadavoodi: Yes, it stopped working because of this change: https://github.com/caffe2/caffe2/pull/2025/files#diff-def726fb1d234a67aeaf10e07aa06da7R995

What happens is that we scope `self.cells[0].prepare_input` with `self.name` (not with layer id) and that causes all the inputs and outputs of `FC` in `prepare_input` to be scoped unless they are given as strings instead of BlobReference. If we want to keep `input_blob` out of the scope of the LSTM, then we have to use a BlobReference, otherwise the test will crash unless we scope https://github.com/caffe2/caffe2/blob/master/caffe2/python/memonger_test.py#L486 as well.

salexspb: Ah, I see. So a trick I did before when applying scopes is the following. If any input is a string, we rectify it before entering a scope. Here is an example. https://github.com/caffe2/caffe2/blob/master/caffe2/python/rnn_cell.py#L182

In your case you just need to take care of one single input_blob :)

salexspb:
@pooyadavoodi , thanks so much!

PR #2035, opened 2018-02-23
[auto] onnx-pr-555

onnxbot-worker-1:
This is an auto PR managed by onnxbot for tracking

https://github.com/onnx/onnx/pull/555 Create unique proto filename based on ONNX_NAMESPACE


bddppq:
@caffe2bot retest this please

PR #2037, opened 2018-02-24
Add eigen version check

Yangqing:
From an internal discussion we found a case where the unsupported/ header is not found and the error message was cryptic. This makes it more explicit that we require 3.3.0+ eigen version.

### DIFF
--- a/caffe2/operators/conv_op_eigen.cc
+++ b/caffe2/operators/conv_op_eigen.cc
@@ -18,6 +18,12 @@
 #include "caffe2/core/operator.h"
 #include "caffe2/operators/conv_pool_op_base.h"
 
+#include "Eigen/Core"
+
+#if !EIGEN_VERSION_AT_LEAST(3, 3, 0)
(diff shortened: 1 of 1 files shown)


caffe2/operators/conv_op_eigen.cc
```diff
@@ -18,6 +18,12 @@
 #include "caffe2/core/operator.h"
 #include "caffe2/operators/conv_pool_op_base.h"
 
+#include "Eigen/Core"
+
+#if EIGEN_VERSION_AT_LEAST(3, 3, 0)
```
orionr: I think you'd actually want

```
#if !EIGEN_VERSION_AT_LEAST(3, 3, 0)
#error "Caffe2 requires Eigen to be at least 3.3.0.";
#endif
```
Otherwise, yes please!

Yangqing: lol yes, realized that error :P

PR #2040, opened 2018-02-24
After nnpack update, cpufeatures is no longer needed.

Yangqing:


Yangqing:
Nah. NVM, still needed.

PR #2042, opened 2018-02-24
[cmake] Move nccl to modern cmake, and avoid using EXTERNAL_DEPENDENCIES

Yangqing:


Yangqing:
@caffe2bot retest this please. Due to cmake version bump and centos. Expectation is that centos will pass and ubuntu 14.04 will fail. Ubuntu 14.04 will come back normal after we add cmake3 to docker image build.

PR #2043, opened 2018-02-24
[cmake] Move nnpack to modern cmake and avoid using EXTERNAL_DEPENDENCIES.

Yangqing:


Maratyszcza:
Removing EXTERNAL_DEPENDENCIES looks good, but removing `if(NOT TARGET nnpack)` check does not. It is needed to avoid double-importing NNPACK when Caffe2 is built from within another project via `add_subdirectory`

Yangqing:
@Maratyszcza good point. Reverted the change to nnpack.cmake.

Maratyszcza (approved):
LGTM

Yangqing:
Thanks - merged to #2045.

alphilly2013:
X

PR #2045, opened 2018-02-24
Remove the use of EXTERNAL_DEPENDENCIES

Yangqing:
Background: Caffe2_EXTERNAL_DEPENDENCIES was a stop-gap solution for us when we did not do careful cmake dependency tracking via proper imported targets. We are currently moving towards using proper imported targets all the way through, so this PR removes the EXTERNAL_DEPENDENCIES usage completely.

Part of the PR have been tested separately and this is a merge of #2044 #2043 #2042 #2041 .


caffe2/contrib/aten/CMakeLists.txt
```diff
@@ -18,8 +18,9 @@ if(USE_ATEN)
   ${CMAKE_CURRENT_SOURCE_DIR}/gen_op.py
   ${CMAKE_CURRENT_SOURCE_DIR}/aten_op_template.h)
 
-  add_custom_target(aten_build
-  DEPENDS aten_op.h)
+  add_custom_target(__aten_op_header_gen DEPENDS aten_op.h)
```
Maratyszcza, line 21: IMO, this should be inside ATen CMakeLists

cmake/Dependencies.cmake
```diff
@@ -70,11 +69,12 @@ endif()
 if(USE_NNPACK)
   include("cmake/External/nnpack.cmake")
   if(NNPACK_FOUND)
-    caffe2_include_directories(${NNPACK_INCLUDE_DIRS})
-    list(APPEND Caffe2_DEPENDENCY_LIBS ${NNPACK_LIBRARIES})
     if(TARGET nnpack)
       # ---[ NNPACK is being built together with Caffe2: explicitly specify dependency
-      list(APPEND Caffe2_EXTERNAL_DEPENDENCIES nnpack)
+      list(APPEND Caffe2_DEPENDENCY_LIBS nnpack pthreadpool cpuinfo)
```
Maratyszcza: Adding just `nnpack` alone should work. NNPACK CMakeLists properly specifies its own dependencies.

cmake/Dependencies.cmake
```diff
@@ -84,17 +84,12 @@ endif()
 
 # ---[ On Android, Caffe2 uses cpufeatures library in the thread pool
 if (ANDROID)
-  # ---[ Check if cpufeatures was already imported
-  if (NOT TARGET cpufeatures)
-    add_library(cpufeatures STATIC
+  add_library(__caffe2_cpufeatures STATIC
```
Maratyszcza: This will create problems if Caffe2 is included using `add_subdirectory` and the parent project also uses libcpufeatures (even if under a different target name) and links Caffe2 statically. If suggest to leave the old snippet, just remove `Caffe2_EXTERNAL_DEPENDENCIES` reference.

caffe2/contrib/aten/CMakeLists.txt
Yangqing: This is actually generated by a separate build script (gen_op.py). ATen itself is a bit entangled at the moment... cc @zdevito @dzhulgakov

dzhulgakov (approved):

cmake/Dependencies.cmake
```diff
@@ -84,17 +84,14 @@ endif()
...
-  # ---[ Check if cpufeatures was already imported
   if (NOT TARGET cpufeatures)
     add_library(cpufeatures STATIC
-      "${ANDROID_NDK}/sources/android/cpufeatures/cpu-features.c")
+        "${ANDROID_NDK}/sources/android/cpufeatures/cpu-features.c")
     target_include_directories(cpufeatures
-      PUBLIC "${ANDROID_NDK}/sources/android/cpufeatures")
-    target_link_libraries(cpufeatures PRIVATE dl)
+        PUBLIC "${ANDROID_NDK}/sources/android/cpufeatures")
+    target_link_libraries(cpufeatures PUBLIC dl)
```
Maratyszcza, line 92: No need to change to `PUBLIC`, dl interface is not exposed in cpufeatures interface.

Maratyszcza (approved):
LGTM

cmake/Dependencies.cmake
Yangqing: Actually for static lib it will be public anyway - as the linkage flags are transitive for static libs regardless of its setting. I put a “public” here to explicitly note this :)

PR #2063, opened 2018-02-27
Remove onnx-caffe2 reference

yinghai:
Since we have moved `onnx-caffe2` into `caffe2`. Update the code reference here.


houseroad (approved):
LGTM, thanks!

PR #2067, opened 2018-02-27
Cleanup CMake files and build scripts for Android

Maratyszcza:
- Remove USE_ARM64 option because it doesn't do what is expected
- Disable ARM ComputeLibrary for non-ARM/ARM64 builds
- Remove analysis of CMake options from scripts/build_android.sh
- Add user-specified CMake options at the end of command line to allow overriding defaults
- Update README for ARM ComputeLibrary integration and do not require to disable NNPACK for ARM64 build with ARM ComputeLibrary

I checked that Caffe2 with ARM ComputeLibrary builds for both ARMv7 and ARM64


Yangqing:
It looks all good. @jerryzh168 do you mind taking a look?

Also, contbuild is having hiccups so we might want to wait and retest. Let me try. @caffe2bot  retest this please.

caffe2/CMakeLists.txt
```diff
@@ -97,14 +97,16 @@ set(Caffe2_MAIN_LIBS)
 # Compile exposed libraries.
 add_library(caffe2 ${Caffe2_CPU_SRCS} $<TARGET_OBJECTS:Caffe_PROTO> $<TARGET_OBJECTS:Caffe2_PROTO>)
 if(USE_ACL)
-  if(NOT USE_ARM64)
+  # Check if 32-bit ARM: armv7, armv7-a, armv7l, etc
```
Yangqing, line 100: Do you mind moving this part to cmake/Dependencies.cmake?

scripts/build_android.sh
```diff
@@ -75,38 +75,16 @@ if [ "${VERBOSE:-}" == '1' ]; then
...
 CMAKE_ARGS+=("-DANDROID_NATIVE_API_LEVEL=21")
 CMAKE_ARGS+=("-DANDROID_CPP_FEATURES=rtti exceptions")
 # TODO: As the toolchain file doesn't support NEON-FP16 extension,
 # we disable USE_MOBILE_OPENGL for now, it will be re-enabled in the future.
 CMAKE_ARGS+=("-DUSE_MOBILE_OPENGL=OFF")
 
-# Compiler flags
-CMAKE_ARGS+=("-DCMAKE_C_FLAGS=")
-if $USE_GCC;then
-  CMAKE_ARGS+=("-DCMAKE_CXX_FLAGS=-s")
```
jerryzh168, line 108: Are we agreed to remove this, if so, probably write down the reasons here as well for reference. Thanks!

jerryzh168 (approved):
Thanks for fixing the build scripts!

Yangqing:
Android is passing so we should probably be good to go :)

Yangqing (approved):

PR #2072, opened 2018-02-28
Mean Op

zem7:


caffe2/operators/mean_op.cc
```diff
@@ -0,0 +1,39 @@
...
+    .SetDoc(R"DOC(
+Element-wise mean of each of the input tensors. The first input tensor can be
+used in-place as the output tensor, in which case the mean will be done in
+place and results will be accumulated in input0. All inputs and outputs must
+have the same shape and data type.
+)DOC")
+    .Input(0, "data_0", "First of the input tensors. Can be inplace.")
+    .Output(0, "mean", "Output tensor. Same dimension as inputs.");
+
+GRADIENT_NOT_IMPLEMENTED_YET(Mean);
```
Yangqing: Do we want to implement the gradients? It is relatively easy (one simply need to do a scale)

caffe2/operators/mean_op.cu
```diff
@@ -0,0 +1,34 @@
...
+
+#include "caffe2/core/context_gpu.h"
+#include "caffe2/operators/mean_op.h"
+
+namespace caffe2 {
+
+REGISTER_CUDA_OPERATOR(Mean, MeanOp<CUDAContext>);
+
+template <>
+bool MeanOp<CUDAContext>::RunOnDevice() {
```
Yangqing: Should this function be in the .h file directly? It seems to be identical with the cpu version.

caffe2/python/operator_test/mean_op_test.py
```diff
@@ -0,0 +1,54 @@
...
+import hypothesis.strategies as st
+import caffe2.python.hypothesis_test_util as hu
+import numpy as np
+
+import unittest
+
+
+class TestMean(hu.HypothesisTestCase):
+    @given(
+        inputs=hu.tensors(n=2, min_dim=2, max_dim=2),
```
Yangqing: We should also test inplace computation since mean supports inplace. Also, we might want to test multiple number of inputs: one way is to generate a maximum number of inputs (say 5), and then generate a random variable ranging between 1 and 5, and test the corresponding number of inputs.

caffe2/python/operator_test/mean_op_test.py
```diff
@@ -0,0 +1,54 @@
...
+        engine=st.sampled_from(["", "CUDNN"]),
+        **hu.gcs)
+    def test_mean(self, inputs, gc, dc, engine):
+        X, Y = inputs
+
+        def mean_ref(X, Y):
+            return np.mean([X, Y], axis=0)
+
+        op = core.CreateOperator("Mean", ["X", "Y"], ["Result"], engine=engine)
+        self.ws.create_blob("X").feed(X)
```
Yangqing: line 42-48: You don't need do to this. self.assertReferenceChecks is the function that does this for you. Also, current implementation does not check non-cpu build cases, which is why one is recommended to use assertReferenceChecks.

Yangqing (changes requested):

Yangqing:
@zem7 friendly ping.

Yangqing:
@zem7 friendly ping again. Updates?

zem7:
@Yangqing: Parts of the PR has been updated, as per the feedbacks. Debugging some test cases for the  hypothesis that considers all the suggestions.

caffe2/operators/mean_op.cc
```diff
@@ -0,0 +1,51 @@
...
+
+class GetMeanGradient : public GradientMakerBase {
+  using GradientMakerBase::GradientMakerBase;
+  vector<OperatorDef> GetGradientDefs() override {
+    Argument scale = MakeArgument<float>("scale", 1.0f / def_.input_size());
+    return SingleGradientDef(
+        "Scale",
+        "",
+        std::vector<string>{GO(0)},
+        std::vector<string>{GI(0)},
```
Yangqing: This only creates gradient for the first input - will need to do that for all inputs.

caffe2/python/operator_test/mean_op_test.py
```diff
@@ -0,0 +1,79 @@
...
+        **hu.gcs
+    )
+    def test_mean(self, k, n, m, in_place, engine, gc, dc):
+        input_names = []
+        input_vars = []
+
+        for i in range(k):
+            X_name = 'X' + str(i)
+            input_names.append(X_name)
+            var = np.random.randn(n, m).astype(np.float32)
```
Yangqing, line 47: Since you do np.random.randn here, it is probably better to have a seed generated by hypothesis, and set np random seed before running the generation.

Yangqing (changes requested):

caffe2/python/operator_test/mean_op_test.py
```diff
@@ -29,14 +29,16 @@
 
 class TestMean(hu.HypothesisTestCase):
     @given(
-        k=st.integers(1, 5),
-        n=st.integers(1, 10),
-        m=st.integers(1, 10),
+        k=st.integers(2, 2),
```
Yangqing: Wait, why did these change? k, n amd m should stay the previous way.

caffe2/python/operator_test/mean_op_test.py
```diff
@@ -29,14 +29,16 @@
...
+        n=st.integers(1, 1),
+        m=st.integers(1, 3),
         in_place=st.booleans(),
         engine=st.sampled_from(["", "CUDNN"]),
```
Yangqing: We shouldn't need engine since there is no CUDNN implementation.

Yangqing (changes requested):

caffe2/python/operator_test/mean_op_test.py
zem7: These changes were accidental; specific test cases were tested for validating k gradient outputs. Reverting these.

Yangqing:

Yangqing (approved):
CI appears to be unrelated error that @pjh5 is fixing. This should be good to go.

houseroad:
@caffe2bot retest this please

PR #2078, opened 2018-02-28
Fix cmake dependency error in static library case. Peer coded with @bddppq

Yangqing:
Manually verified with local build on Linux docker.


caffe2/CMakeLists.txt
```diff
@@ -127,12 +127,7 @@ if (BUILD_TEST)
     # result, we will explicitly link the test against the Caffe2 dependency
     # libs.
     target_link_libraries(
-          ${test_name} ${Caffe2_MAIN_LIBS} ${Caffe2_DEPENDENCY_LIBS}
-          gtest_main)
-    if (USE_CUDA)
-      target_link_libraries(
-          ${test_name} ${Caffe2_CUDA_DEPENDENCY_LIBS})     
-    endif()
+          ${test_name} ${Caffe2_MAIN_LIBS} gtest_main)
```
Yangqing: As a future reference, target_link_libraries does not guarantee the order of the linkage; to really ensure the order, one need to make it in different target_link_libraries calls.

In this case, we actually do not need ${Caffe2_DEPENDENCY_LIBS} because  ${Caffe2_MAIN_LIBS} will already contain it. What caused error was that, because we included ${Caffe2_DEPENDENCY_LIBS} here, it went ahead of ${Caffe2_MAIN_LIBS} and caused the libcaffe2.a file to not find symbols correctly (linking order problem).

bddppq (approved):

PR #2087, opened 2018-03-01
Fix the pybin11_state_gpu.so linking issue

houseroad:
Problem: not really do the static linking on libcaffe2.a and libcaffe2_gpu.a against pybind11_state_gpu.so when we enable static linking.

Solution: Use double quotes to enforce the order.

Thanks @Yangqing


yinghai (approved):
Awesome! I had the same issue!

bddppq (approved):

PR #2062, opened 2018-02-27
Fix caffe_add_whole_archive_flag in cmake

yinghai:
During link time, `-Wl,...` will pass the arguments down to system specific linker. It doesn't matter whether we call `clang` or `g++`. However, the flags will passed to different linker depending on the platform. See https://cxwangyi.wordpress.com/2012/01/20/using-clang-linker-on-different-platforms/.

PR #2091, opened 2018-03-01
Refix the linkage condition

yinghai:
This was fixed in https://github.com/caffe2/caffe2/pull/2062 but overwritten recently. Resubmit the fix.


bddppq (approved):

Yangqing:
Thanks for catching this - sorry about the original diff.

PR #2093, opened 2018-03-01
Fixing mkl builds not using mkl

pjh5:


bddppq (approved):

pjh5:
All of these failures are unfortunately expected so I will merge now

PR #2094, opened 2018-03-01
[cmake] Building rocksdb as a module

Yangqing:
After the many cmake refactors now we should be able to properly modularize the optional components. This pull request provides a scaffolding example for modularizing rocksdb - one optional dependency that we have in the toolchain.

Note: for fb internal build system (buck) this will require us to change the build targets. @orionr what's the recommended way of doing so at this moment?


Yangqing:
@caffe2bot retest this please

pjh5 (approved):

PR #2105, opened 2018-03-01
Copy some outputs in order to decouple storage

anderspapitto:
so that mutating one of them does not mutate the others

### DIFF
--- a/caffe2/python/onnx/backend.py
+++ b/caffe2/python/onnx/backend.py
@@ -467,9 +467,15 @@ class Caffe2Backend(Backend):
 
         if direction == 'forward':
             hidden_t_all, hidden_t_last = make_rnn(0)
+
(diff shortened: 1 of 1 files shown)


caffe2/python/onnx/backend.py
```diff
@@ -582,11 +588,16 @@ def make_lstm(direction_offset):
 
         if direction == 'forward':
             hidden_t_all, hidden_t_last, cell_last = make_lstm(0)
+
+            # in the forward case, storage is shared between the three
```
dzhulgakov, line 592: what do you mean storage is shared? Slice in caffe2 copies memory afaiu.

dzhulgakov (commented):
Wouldn't it be just easier to copy in VariableLength then?

PR #2106, opened 2018-03-02
Warning if engine is not available

jspark1105:


caffe2/core/operator.cc
```diff
@@ -170,7 +170,8 @@ unique_ptr<OperatorBase> _CreateOperator(
     } else {
       // If the above fails, we will just return the normal case with the
       // default implementation.
-      VLOG(1) << "Operator with engine " << engine << " is not available.";
+      LOG(WARNING) << "Operator with engine " << engine
```
bddppq: For preferred engines (e.g. cudnn for gpu operators), if it's not found, I don't think this should be a warning.

bddppq (changes requested):

caffe2/core/operator.cc
jspark1105: Then how about INFO ?

bddppq (approved):

PR #2114, opened 2018-03-02
Fix Out of Memory failure in test TensorTest.Tensor64BitDimension

mingzhe09088:


bddppq:
@caffe2bot test this please

caffe2/core/blob_test.cc
```diff
@@ -538,7 +538,13 @@ TEST(TensorTest, Tensor64BitDimension) {
   EXPECT_EQ(tensor.ndim(), 1);
   EXPECT_EQ(tensor.dim(0), large_number);
   EXPECT_EQ(tensor.size(), large_number);
-  EXPECT_TRUE(tensor.mutable_data<char>() != nullptr);
+  try {
+    EXPECT_TRUE(tensor.mutable_data<char>() != nullptr);
```
Yangqing, line 542: Thanks! You might want to wrap line 552 as well.

caffe2/core/blob_test.cc
```diff
@@ -538,7 +538,13 @@ TEST(TensorTest, Tensor64BitDimension) {
...
+  } catch (const EnforceNotMet& e) {
+    string msg = e.what();
+    msg = msg.substr(0, msg.find('\n'));
+    LOG(WARNING) << msg;
```
Yangqing: You may want to make sure that the error message contains "aligned_malloc" - for all others, we should still not swallow the exception.

Yangqing (changes requested):

caffe2/core/blob_test.cc
mingzhe09088: Thanks! I've updated that.

caffe2/core/blob_test.cc
mingzhe09088: Updated as well.

bddppq:
@caffe2bot add to whitelist

caffe2/core/blob_test.cc
```diff
@@ -538,12 +538,19 @@ TEST(TensorTest, Tensor64BitDimension) {
...
-  // Try to go even larger, but this time we will not do mutable_data because we
-  // do not have a large enough memory.
-  tensor.Resize(large_number, 100);
+  try {
+    EXPECT_TRUE(tensor.mutable_data<char>() != nullptr);
+    // Try to go even larger, but this time we will not do mutable_data because we
+    // do not have a large enough memory.
+    tensor.Resize(large_number, 100);
+  } catch (const EnforceNotMet& e) {
+    string msg = e.what();
```
Yangqing: As mentioned in the above comment: you should do a string search for aligned_malloc, and only when the exception is caused by aligned malloc should we ignore it.

Yangqing (changes requested):

caffe2/core/blob_test.cc
bddppq: @Yangqing looks to me `Resize` doesn't allocate memory

bddppq:
@mingzhe09088 We should only wrap the line(s) that throw the OOM exception in the test.

pjh5:
I am not sure why the conda build is still failing, but it looks like this error was fixed on it. I would still check this in despite the conda failure

mingzhe09088:
@bddppq From the previous testing result, when the OOM error happened, all the following lines in that function also failed. So I protect the following lines.

bddppq:
@mingzhe09088 In case OOM error happened the rest of the test should be skipped (just early return).

mingzhe09088:
@bddppq Thanks! I've pushed a fix.

Yangqing (approved):
Looks pretty good! :)

bddppq (approved):
Looks good. Thanks

PR #2116, opened 2018-03-02
Fail fast in pytest

yinghai:
Pytest usage: https://docs.pytest.org/en/latest/usage.html#stopping-after-the-first-or-n-failures

Trying to mark the test as failed upon first pytest failure. Benefits:
- Same computation resource.
- Failed test always show at the bottom so that you don't need to scroll up to find them. 
Cons:
- If there are multiple failed tests and you rely on CI to flush them out, it will take more iterations.

### DIFF
--- a/.jenkins/test.sh
+++ b/.jenkins/test.sh
@@ -105,6 +105,7 @@ fi
 echo "Running Python tests.."
 "$PYTHON" \
   -m pytest \
+  -x \
   -v \
   --junit-xml="$TEST_DIR/python/result.xml" \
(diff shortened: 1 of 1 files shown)


pjh5:
Without this flag I don't see what the pytest error actually is. e.g. if pytest is failing because of a missing import statement then all that appears in the output is "Aborted", which is completely worthless. If I add this flag, then I get the full error message of the failed test. I am further in support of adding this.

yinghai:
@bddppq More input?

pjh5 (approved):

PR #2122, opened 2018-03-03
Move main reshape logic for easier reuse

kevinbchen:
We'll want to reuse this logic for Int8 Reshape, but currently the code assumes
Input(0) and Output(0) are TensorCPUs, which may not be the case for a
subclass.


Yangqing:
Could you elaborate the use case? I

kevinbchen:
This mirrors the phabricator diff D7133671 (some more context in there and D7105622). I'm not quite familiar with the github sync process, let me know if I should be doing something different.

Yangqing:
(Sorry, the close was a misclick of yubikey...)

PR #2123, opened 2018-03-03
Fixing conda

pjh5:


Yangqing (approved):

pjh5:
This change will now encompass fixes needed for the conda_full build (which will no longer run in CI, as it must work in a special environment that can not be emulated there).

The conda-tests are still not working, due to opencv requiring specific protobuf versions that are hard to find without introducing gcc ABI issues.

pjh5:
Looks like this is working for conda3, but conda2 still failing with a segfault during conda install P59201060

pjh5:
\o/ conda2 works too! @caffe2bot test this please

.jenkins/build.sh
```diff
@@ -51,9 +51,26 @@ if [[ "${BUILD_ENVIRONMENT}" == *-android* ]]; then
   exit 0
 fi
 if [[ "${BUILD_ENVIRONMENT}" == conda* ]]; then
+
+  # click (required by onnx) wants these set
+  export LANG=C.UTF-8
+  export LC_ALL=C.UTF-8
```
pjh5, line 57: This was needed to fix an exception thrown by 'click' (used by onnx) that complained about detection incorrect locale configuration on the current machine. These same lines appear https://github.com/caffe2/caffe2/blob/master/docker/jenkins/common/install_anaconda.sh#L6 but I guess the values don't persist.

caffe2/core/predictor.h
```diff
@@ -19,7 +19,7 @@
 #include <unordered_set>
 #include "caffe2/core/net.h"
 #include "caffe2/core/tensor.h"
-#include "caffe2/proto/metanet.pb.h"
+#include "caffe2/proto/caffe2.pb.h"
```
pjh5: see comment in caffe2/proto/caffe2.proto

.jenkins/test.sh
```diff
@@ -91,7 +94,7 @@ for test in ./test/*; do
 done
 
 # Get the relative path to where the caffe2 python module was installed
-CAFFE2_PYPATH="$SITE_DIR/caffe2"
+CAFFE2_PYPATH="$INSTALL_SITE_DIR/caffe2"
```
pjh5, line 97: These changes needed to work with conda's different prefix in the site-packages name

caffe2/python/predictor/predictor_exporter_test.py
```diff
@@ -29,6 +29,23 @@
...
+        caffe2_pb2.NetsMap(key="test_key", value=caffe2_pb2.NetDef())
+
+    def test_adding_net(self):
+        '''
+        Tests that NetDefs can be added to MetaNetDefs
+        '''
+        meta_net_def = caffe2_pb2.MetaNetDef()
+        net_def = caffe2_pb2.NetDef()
+        meta_net_def.nets.add(key="test_key", value=net_def)
+
```
pjh5: These tests both fail when the messages in metanet_pb2 are in their own file that imports caffe2_pb2

conda/cuda_full/build.sh
```diff
@@ -52,10 +57,7 @@ CMAKE_ARGS+=("-DUSE_ROCKSDB=OFF")
...
 CMAKE_ARGS+=("-DCMAKE_PREFIX_PATH=$PREFIX")
 
-# No rpaths will work for anaconda?
-# -DCMAKE_SKIP_RPATH=ON \
-
-
+# Build
 mkdir -p build
 cd build
 cmake "${CMAKE_ARGS[@]}"  $CONDA_CMAKE_ARGS $PYTHON_ARGS ..
```
pjh5, line 63: This file is custom for the GPU cluster environment. It is not run in CI and will not work there, nor on "normal" setups. This was a special ask

scripts/read_conda_versions.sh
```diff
@@ -0,0 +1,181 @@
...
+    MENTIONS_CXX11="$(nm "$CONDA_ROOT/lib/$PKG_INSTALLED_LIB" | grep cxx11 | wc -l)"
+    if [ $MENTIONS_CXX11 -gt 0 ]; then
+      echo "  This package is built against the recent gcc ABI ($MENTIONS_CXX11 mentions of cxx11)"
+      echo "$CONDA_ROOT/lib/$PKG_INSTALLED_LIB mentions cxx11 $MENTIONS_CXX11 times" >> $VERBOSE_LOG
+    fi
+  else
+    echo "Error installing $PKG_SPEC , continuing"
+    echo "Error installing $PKG_SPEC , continuing" >> $VERBOSE_LOG
+  fi
+done
```
pjh5, line 181: This is the script I used to scrape all the conda packages for mentions of cxx11 to guess which gcc ABI they had. It is not called by anything.

pjh5:
This review fixes the conda tests for CI. Before now, conda builds were only tested by running 'from caffe2.python import core' and seeing that it worked. This enables all c++ and python tests to work on the conda builds too. After this lands, some conda builds will be moved from build-only to build-and-test in CI

pjh5:
@Yangqing just an FYI this has changed significantly since you reviewed it
@salexspb this moves the metanet_pb2 into caffe2_pb2 b/c an error mentioned in a comment in that file. This change (and the associated renaming metanet_pb2.MetaNetDef() -> caffe2_pb2.MetaNetDef()) is the only change in the diff that will affect non-conda-build code

pjh5:
Okay actually it still breaks on conda3 because of locale issues. But I'd like to check this in anyways to unblock conda2 tests

PR #2131, opened 2018-03-05
Add C++ preprocessor define CAFFE2_USE_EXCEPTION_PTR

kuttas:
Add C++ preprocessor define CAFFE2_USE_EXCEPTION_PTR to guard use of std::exception_ptr on platforms where it is supported.

Tested this out locally by checking "#ifndef CAFFE2_USE_EXCEPTION_PTR / #error foo / #endif". Will land this first, then separate change to use std::exception_ptr for cross-thread exception marshalling.


bddppq:
@caffe2bot retest this please

Yangqing (approved):

Issue #2127, opened 2018-03-03
CMake problem


FrancoisFayard:
Hi. I am trying to build caffe2 on Ubuntu 16.04 and link it to the tools that come from the latest version of Anaconda 3 which comes with Python 3.6.4. Even though I went as far as giving the Python executable, CMake still recognize the Python that comes with Ubuntu which is 2.7. The CMake message is even more strange as it recognizes Python 3.6.4 from Anaconda but takes the Python library 2.7.

What can I do to solve this issue?

```
### System information

* Operating system: Ubuntu 16.04
* Compiler version: Gcc 5.4.0
* CMake version: 3.5.1
* CMake arguments: cmake ../caffe2  -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/opt/caffe-2 -DUSE_ROCKSDB=OFF -DNUMPY_INCLUDE_DIR=/opt/anaconda-3/lib/python3.6/site-packages/numpy/core/include -DGFLAGS_INCLUDE_DIR=/opt/anaconda-3/include -DGLOG_INCLUDE_DIR=/opt/anaconda-3/include  -DGFLAGS_LIBRARY=/opt/anaconda-3/lib/libgflags.so -DGLOG_LIBRARY=/opt/anaconda-3/lib/libglog.so -DPYTHON_EXECUTABLE=/opt/anaconda-3/bin/python -DPYTHON_INCLUDE_DIR=/opt/anaconda-3/include/python3.6m -DSNAPPY_INCLUDE_DIR=/opt/anaconda-3/include -DSNAPPY_LIBRARY=/opt/anaconda-3/lib/libsnappy.so -DLEVELDB_INCLUDE_DIR=/opt/anaconda-3/include -DLEVELDB_LIBRARY=/opt/anaconda-3/lib/libleveldb.so
* Relevant libraries/versions (e.g. CUDA): Cuda 9.1, CuDNN 7.0.5

### CMake summary output

-- ******** Summary ********
-- General:
--   CMake version         : 3.5.1
--   CMake command         : /usr/bin/cmake
--   Git version           : v0.8.1-1163-g9c60bcc
--   System                : Linux
--   C++ compiler          : /usr/bin/c++
--   C++ compiler version  : 5.4.0
--   Protobuf compiler     : /opt/anaconda-3/bin/protoc
--   Protobuf include path : /opt/anaconda-3/include
[...]

PR #2132, opened 2018-03-05
Update Python information shown in CMake summary

bddppq:
Do not show "Python library" as we no longer link with libpython.
Show "Python includes" because this is useful for debugging whether the python headers are from the expected Python.


bddppq:
https://github.com/caffe2/caffe2/issues/2127

pjh5 (approved):

PR #2135, opened 2018-03-05
Use prepare_input_override in derived RNNCells to rectify input_blob

pooyadavoodi:
This helps with rectifying blobs. We don't need to worry about input_blob getting scoped in case it's a string rather than a BlobReference.

### DIFF
--- a/caffe2/python/operator_test/rnn_cell_test.py
+++ b/caffe2/python/operator_test/rnn_cell_test.py
@@ -1074,9 +1074,10 @@ class RNNCellTest(hu.HypothesisTestCase):
         forget_bias=st.floats(-10.0, 10.0),
         forward_only=st.booleans(),
         drop_states=st.booleans(),
+        num_layers=st.integers(1, 9),
     )
     @ht_settings(max_examples=10)
-    def test_layered_lstm(self, input_tensor, **kwargs):
+    def test_layered_lstm(self, input_tensor, num_layers, **kwargs):
         for outputs_with_grads in [[0], [1], [0, 1, 2, 3]]:
             for memory_optim in [False, True]:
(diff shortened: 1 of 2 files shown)


salexspb:
@pooyadavoodi , sorry for a delay. I am confused here. Does it matter if input to prepare_input is a string or not? if user implements custom prepare_input function, it is going to use either net. or brew. interface. In either case a string would be rectified automatically by Caffe2. 

The reason we had to rectify manually in _apply was that we introduced a new scope. But if user passes in a string we did not want that to get into a new scope before it is used. 

I don't understand how this is an issue for the prepare_input function. Also looks like for the MultiRNNCell it is also not an issue. 

Please correct me if I am wrong!

salexspb (changes requested):
Please see my previous comment

pooyadavoodi:
@salexspb Sorry for the delay. I think I finally fixed the notification problem on my github account.

I think the problem was in MultiRNNCell, where we had to rectify the input: https://github.com/caffe2/caffe2/blob/master/caffe2/python/rnn_cell.py#L980

This occurs for example when `input_blob` of `rnn_cell._LSTM` is a string like here: https://github.com/caffe2/caffe2/blob/master/caffe2/python/rnn/rnn_cell_test_util.py#L54

salexspb:
I am confused. AddScopedExternalInputs doesn't return a string, it returns a ScopedBlobReference, not a string.

pooyadavoodi:
hmm, I see. I don't remember now why I added https://github.com/caffe2/caffe2/blob/master/caffe2/python/rnn_cell.py#L980

Let me go through my old tests. I will let you know.

orionr:

PR #2138, opened 2018-03-05
[tiny] make debug info optional, CAFFE2_DEBUG env variable driven

bwasti:
This will cause 1) debug info to normally be dropped 2) to allow users to attach debug info with and env variable.
In the future there may be other constructors and this will allow framework authors to have a
consistent mechanism for determining if debug info should be attached.

### DIFF
--- a/caffe2/python/core.py
+++ b/caffe2/python/core.py
@@ -36,6 +36,7 @@ import pickle
 import numpy as np
 import sys
 import traceback
+import os
 
 # Mac os specific message
 if (sys.platform == 'darwin' and 'leveldb' in C.registered_dbs()):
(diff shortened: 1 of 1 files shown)


salexspb:
Should we actually remove this all together? Looks like folks are leaning towards going a completely different route now. So we don't encourage people storing things there.

bwasti:
if a new route is decided we can always update this.  this is an extremely simple and easy to use solution to a relatively simple problem, I don't see a need to complicate it

Yangqing (approved):

salexspb:
@bwasti , I actually think this can cause issues in certain cases. In case somebody is running with this debug option and then you just want to print protobufs, they would become very convoluted.

salexspb:
of course this diff itself is making things better, not worse, feel free to merge. Still my point is that we might want to get rid of this feature (which is very new) all together.

PR #2145, opened 2018-03-06
Caffe2 module update: move observers as well as binaries.

Yangqing:
This does 2 things:

(1) for the observers library under contrib/ we move it as an explicit module.
(2) for the binaries, we put it at a top level binaries/ folder and build it after the main caffe2 library and any modules being built. All modules built with the main library is also going to be linked by the binaries automatically.

Firing a diff to make sure that things build. Verified on local macbook.


Yangqing:
Rebased past #2150 .

Yangqing:
Checked with @ezyang offline and this is good to go, so merging.

Note the facebook internal: this will change the build target locations for binaries.

PR #2146, opened 2018-03-06
Add ClipByGlobalNorm operator

pooyadavoodi:
- Improve `hu.tensors` arguments.
- Use ClipByGlobalNormOp in optimizer.py
- Support FP16 in ClipByGlobalNorm
- Use scratch space in SumSqr to improve perf for large sizes.
- Add arg clip_using_lr_multiplier to optimizers so that we can use the old method as well.


orionr:

PR #2147, opened 2018-03-06
FP16 native compute in Conv & ConvGradient op (with CuDNN)

kuttas:
Conv/ConvGradient now support FP16 native compute with CuDNN engine. Previously FP16 inputs would only work with FP32 math (aka pseudo fp16 compute).

Expanded CuDNN exhaustive search to allow finding optimal algorithms for these new compute modes. Will also compare FP32 algos in case they are faster than FP16 for a given input size.

Allow all parts of the Conv Op (forward and backwards pass) to have their own descriptor, allowing each to choose their own optimal algorithm/compute mode.

Migrate some features of ConvOp to ConvTransposeOp for consistency (force_algo - ability to specify a particular algorithm, and tensor core support).

Refactor some shared code used by ConvOp & ConvTransposeOp to new op_utils_cudnn.h file.
Update algorithm cache class to support differentiating on compute type (required for fp16 vs fp32 algo search).

Reviewed & approved by ataei (D5841338)

### DIFF
--- a/caffe2/operators/conv_op_cache_cudnn.h
+++ b/caffe2/operators/conv_op_cache_cudnn.h
@@ -25,46 +25,58 @@
 #include "caffe2/core/tensor.h"
 
 namespace caffe2 {
-template <typename T>
+template <typename TAlgorithm>
 class AlgorithmsCache {
  public:
-  T getAlgorithm(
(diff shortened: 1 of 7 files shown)


bddppq:
@caffe2bot retest this please

PR #2148, opened 2018-03-06
[debug_info] Revert #2008

bwasti:


kuttas:
https://github.com/caffe2/caffe2/commit/989b11fdaed94a5104e89b807e567bda4599666e needs to be reverted with this I believe - tests are failing due to debug_info references

PR #2160, opened 2018-03-06
Use Ninja build system when available

Maratyszcza:
When Ninja is installed, use it instead of Make for native builds and for Android cross-builds.
Ninja is faster and generates less output in the terminal on successful builds.

### DIFF
--- a/scripts/build_android.sh
+++ b/scripts/build_android.sh
@@ -48,6 +48,11 @@ cd $BUILD_ROOT
 
 CMAKE_ARGS=()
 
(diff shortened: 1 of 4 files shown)


orionr (approved):
LGTM

Maratyszcza:
@caffe2bot retest this please

bddppq (changes requested):
setup.py needs to be changed as well

Maratyszcza:
@bddppq PTAL

bddppq (approved):
Looks good

PR #2161, opened 2018-03-07
Add outputs_info as parameter in run_node

houseroad:
To address the changes in onnx repo.

### DIFF
--- a/caffe2/python/onnx/backend.py
+++ b/caffe2/python/onnx/backend.py
@@ -205,8 +205,8 @@ class Caffe2Backend(Backend):
     # so this interface MAY make BC-breaking changes.  Specify an
     # opset_version if you don't want this to version.
     @classmethod
-    def run_node(cls, node, inputs, device='CPU', opset_version=_known_opset_version):
-        super(Caffe2Backend, cls).run_node(node, inputs, device)
+    def run_node(cls, node, inputs, device='CPU', opset_version=_known_opset_version, outputs_info=None):
+        super(Caffe2Backend, cls).run_node(node, inputs, device=device, outputs_info=outputs_info)
 
         device_option = get_device_option(Device(device))
         with Workspace(), core.DeviceScope(device_option):  # temporary!


caffe2/python/onnx/backend.py
```diff
@@ -205,7 +205,7 @@ class Caffe2Backend(Backend):
     # so this interface MAY make BC-breaking changes.  Specify an
     # opset_version if you don't want this to version.
     @classmethod
-    def run_node(cls, node, inputs, device='CPU', opset_version=_known_opset_version):
+    def run_node(cls, node, inputs, device='CPU', outputs_info=None, opset_version=_known_opset_version):
```
bddppq: new arg should be put at the end (in case there are callers invoking this function with all positional args).

bddppq: onnx backend test suit should pass this new arg as keyword arg.

houseroad: Sure, that sounds good to me. Just switched the order.

caffe2/python/onnx/backend.py
```diff
@@ -205,7 +205,7 @@ class Caffe2Backend(Backend):
     # so this interface MAY make BC-breaking changes.  Specify an
     # opset_version if you don't want this to version.
     @classmethod
-    def run_node(cls, node, inputs, device='CPU', opset_version=_known_opset_version):
+    def run_node(cls, node, inputs, device='CPU', opset_version=_known_opset_version, outputs_info=None):
         super(Caffe2Backend, cls).run_node(node, inputs, device)
```
bddppq: (I know this is not introduced in this diff) we should pass both device and outputs_info back to the parent class here (as keyword arg).

bddppq (approved):

caffe2/python/onnx/backend.py
houseroad: Good catch. Changed.

houseroad:
@caffe2bot retest this please

houseroad:
@caffe2bot test this please

houseroad:
@caffe2bot retest this please

smessmer (approved):

bddppq:
@caffe2bot retest this please

PR #2162, opened 2018-03-07
Bump minimum CMake version to 3.2

Maratyszcza:
CMake 3.2 is required to properly track dependencies in projects imported as `ExternalProject_Add` (`BUILD_BYPRODUCTS` parameter).
Users on Ubuntu 14.04 LTS would need to install and use `cmake3` package for configurations. Users of other popular distributions generally have a recent enough CMake package.


Yangqing (approved):

Maratyszcza:
@Yangqing I replaced `-std=c++11` with CMake magic, merging after the tests pass.

PR #2164, opened 2018-03-07
Use cpuinfo instead of Android's libcpufeatures in Android build

Maratyszcza:
This makes cpuinfo a required dependency of Caffe2

### DIFF
--- a/caffe2/utils/threadpool/ThreadPool.cc
+++ b/caffe2/utils/threadpool/ThreadPool.cc
@@ -18,9 +18,7 @@
 #include "WorkersPool.h"
 #include "caffe2/core/logging.h"
 
-#if CAFFE2_ANDROID
-#include <cpu-features.h>
-#endif
+#include <cpuinfo.h>
 
 CAFFE2_DEFINE_bool(caffe2_threadpool_force_inline, false,
                    "Force to always run jobs on the calling thread");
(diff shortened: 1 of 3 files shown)


ajtulloch:
Can we make this an optional dependency, gated behind CAFFE2_USE_CPUINFO or similar? I'm concerned about adding this as a non-optional dependency that some users would prefer not to use.

Maratyszcza:
@ajtulloch What are your concerns about making cpuinfo a non-optional dependency? cpuinfo now supports all Caffe2 platforms, including Windows. The plan is to switch to cpuinfo for ISA detection in Caffe2, and remove code in `caffe2/utils/cpuid.cc` as the next step. That, of course, wouldn't work with an optional dependency.

ajtulloch:
@Maratyszcza the concern is having this substantial, non-optional addition of code size (~50kb on ARMv7, right? That's about the size of the whole int8 backend) for a relatively small piece of functionality (just #cores), which is essentially already provided by libcpufeatures 'for free' (since everyone already depends on that in various context).

Maratyszcza:
@ajtulloch When we switched to `cpuinfo` for NNPACK, it added only 13K to Caffe2 on Android. After this change Caffe2 would not depend on `libcpufeatures`, so it will be a net size reduction for apps which don't use `libcpufeatures`, or can replace it with `cpuinfo`. In this PR `cpuinfo` is only used to get number of cores and MIDR, but adding it as a non-optional dependency opens way for other features which need chipset information.

ajtulloch (approved):
sgtm, 15kb of code seems reasonable.

Maratyszcza:
@caffe2bot retest this please

PR #2166, opened 2018-03-07
fix comment on the location of scale and bias (offset) in each fused rowwise 8bit

jspark1105:

### DIFF
--- a/caffe2/operators/fused_rowwise_8bit_conversion_ops.cc
+++ b/caffe2/operators/fused_rowwise_8bit_conversion_ops.cc
@@ -47,9 +47,9 @@ OPERATOR_SCHEMA(Fused8BitRowwiseQuantizedToFloat)
     .SetDoc(R"DOC(
 De-quantizes the result of the
 FloatToFused8BitRowwiseQuantized operator. The input is expected to
-encode the scale as a 32-bit float in the first 4 bytes of each row,
-followed by the bias as a 32-bit float in the next 4 bytes, followed by
-the quantized values in the remaining bytes of the row. The output is a
+encode the scale as a 32-bit float in the second to the last 4 bytes of each
+row, followed by the bias as a 32-bit float in the next 4 bytes, and the
+quantized values in the preceding remaining bytes of the row. The output is a
 matrix containing only the values, but de-quantized. De-quantization is
 performed by multiplying each value by its row's scale and bias
 parameters. The de-quantized values will thus not be exactly equal to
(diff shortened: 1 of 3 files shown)


caffe2/operators/fused_rowwise_8bit_conversion_ops.cc
```diff
@@ -47,9 +47,9 @@ OPERATOR_SCHEMA(Fused8BitRowwiseQuantizedToFloat)
     .SetDoc(R"DOC(
 De-quantizes the result of the
 FloatToFused8BitRowwiseQuantized operator. The input is expected to
-encode the scale as a 32-bit float in the first 4 bytes of each row,
-followed by the bias as a 32-bit float in the next 4 bytes, followed by
-the quantized values in the remaining bytes of the row. The output is a
+encode the scale as a 32-bit float in the second to the last 4 bytes of each
+row, followed by the bias as a 32-bit float in the next 4 bytes, and the
+quantized values in the preceding remaining bytes of the row. The output is a
```
kuttas: "preceding remaining" is a little confusing - do you just mean "preceding" here?

caffe2/operators/lengths_reducer_fused_8bit_rowwise_ops.cc
```diff
@@ -28,7 +28,7 @@ OPERATOR_SCHEMA(SparseLengthsSumFused8BitRowwise)
     .SetDoc(R"DOC(
 Performs the same operation as SparseLengthsSum, but operating on
 8-bit rowwise quantized matrices with fused storage (where each row
-stores the scale, bias and then quantized values).
+stores quantized values, and then the scale and bias).
```
kuttas: Is it worth going into detail, e.g. "and then 8 bytes: 4 bytes for scale and 4 bytes for bias" or is that obvious given the context? (same tweak could apply for the other comment corrections below)

kuttas (approved):
Had 2 clarifying questions on your changes - submit if you think it's good as-is or fix nits and then submit

caffe2/operators/fused_rowwise_8bit_conversion_ops.cc
jspark1105: Thanks for catching the typo.

caffe2/operators/lengths_reducer_fused_8bit_rowwise_ops.cc
jspark1105: Good point. Will do.

PR #2186, opened 2018-03-08
Adding rewrite_net for ACL backend

jerryzh168:


caffe2/mobile/contrib/arm-compute/core/rewrite_net.cc
```diff
@@ -62,7 +62,7 @@ static void insertCopyFromGLOp(NetDef& predictNet, const std::string& cpu_blob)
   auto* op = predictNet.add_op();
   op->set_name("CopyFromGL");
   op->set_type("CopyFromGL");
-  op->add_input(cpu_blob);
+  op->add_input(cpu_blob + "_M");
```
Maratyszcza: Why?

jerryzh168: We can use an temporary variable to first copy out the gpu memory, and then reuse the blob to store the cpu result, but I feel that might be less efficient comparing to copy from the gpu memory directly to cpu memory(which is contained in another blob)?

PR #2189, opened 2018-03-08
Reduce Sum and Reduce Mean

zem7:
ReduceSum(https://github.com/onnx/onnx/blob/master/docs/Operators.md#ReduceSum) and ReduceMean (https://github.com/onnx/onnx/blob/master/docs/Operators.md#ReduceMean) operations for the forward pass, so users are unblocked on more common inference scenario.


caffe2/operators/reduce_sum_op.h
```diff
@@ -0,0 +1,116 @@
...
+          "ReduceSum operator only supports 32-bit float, but",
+          " input was of type ",
+          Input(0).meta().name());
+    }
+  }
+
+ private:
+  std::vector<int> axes_;
+  int keepdims_;
+  Tensor<Context> buffer_;
```
houseroad: Do we really need buffer here?

caffe2/operators/reduce_mean_op.h
```diff
@@ -0,0 +1,116 @@
...
+      T* output_data = output->template mutable_data<T>();
+      T* temp_buf = buffer_.template mutable_data<T>();
+
+      for (int i = 0; i < front; i++) {
+        for (int k = 0; k < back; k++) {
+          T sum = 0;
+          for (int j = 0; j < reduced_dim; j++) {
+            sum += output_data[i * reduced_dim * back + j * back + k];
+          }
+          temp_buf[i * back + k] = sum / reduced_dim;
```
houseroad: Please directly write to the output, unless we need to support in place. But no inplace is allowed in this op.

caffe2/operators/reduce_mean_op.h
```diff
@@ -0,0 +1,116 @@
...
+      int front = size_to_dim_(reduced_axis, dims);
+      int back = size_from_dim_(reduced_axis + 1, dims);
+
+      T* output_data = output->template mutable_data<T>();
+      T* temp_buf = buffer_.template mutable_data<T>();
+
+      for (int i = 0; i < front; i++) {
+        for (int k = 0; k < back; k++) {
+          T sum = 0;
+          for (int j = 0; j < reduced_dim; j++) {
```
houseroad: Please consider the cache miss case.

caffe2/operators/reduce_mean_op.h
```diff
@@ -0,0 +1,116 @@
...
+          "ReduceMean operator only supports 32-bit float, but",
+          " input was of type ",
+          Input(0).meta().name());
+    }
+  }
+
+ private:
+  std::vector<int> axes_;
+  int keepdims_;
+  Tensor<Context> buffer_;
```
houseroad: Why we need this?

caffe2/operators/reduce_sum_op.cc
```diff
@@ -0,0 +1,41 @@
...
+
+REGISTER_CPU_OPERATOR(ReduceSum, ReduceSumOp<CPUContext>);
+
+OPERATOR_SCHEMA(ReduceSum)
+    .NumInputs(1)
+    .NumOutputs(1)
+    .SetDoc(R"DOC(
+      Computes the sum of the input tensor's element along the provided axes.
+      The resulted tensor has the same rank as the input if keepdims equal 1.
+      If keepdims equal 0, then the resulted tensor have the reduced dimension pruned.
```
houseroad: Specify the behavior when list is empty.

caffe2/operators/reduce_sum_op.h
```diff
@@ -0,0 +1,116 @@
...
+    buffer_.ResizeLike(input);
+
+    for (int id = axes_.size() - 1; id >= 0; id--) {
+      int reduced_axis = axes_[id];
+      int reduced_dim = dims[reduced_axis];
+      int front = size_to_dim_(reduced_axis, dims);
+      int back = size_from_dim_(reduced_axis + 1, dims);
+
+      T* output_data = output->template mutable_data<T>();
+      T* temp_buf = buffer_.template mutable_data<T>();
```
houseroad: Why buffer...

caffe2/operators/reduce_sum_op.h
```diff
@@ -0,0 +1,116 @@
...
+      std::sort(axes_.begin(), axes_.end());
+      CAFFE_ENFORCE(axes_.front() >= 0, "Axes ids must be non-negative.");
+      CAFFE_ENFORCE(
+          axes_.back() < ndim,
+          "Axes ids must be smaller than the dimensions of input.");
+    }
+
+    auto* output = Output(0);
+    output->ResizeLike(input);
+    output->CopyFrom(input, &context_);
```
houseroad: Why copy?

caffe2/operators/reduce_sum_op.h
```diff
@@ -0,0 +1,116 @@
...
+      int back = size_from_dim_(reduced_axis + 1, dims);
+
+      T* output_data = output->template mutable_data<T>();
+      T* temp_buf = buffer_.template mutable_data<T>();
+
+      for (int i = 0; i < front; i++) {
+        for (int k = 0; k < back; k++) {
+          T sum = 0;
+          for (int j = 0; j < reduced_dim; j++) {
+            sum += output_data[i * reduced_dim * back + j * back + k];
```
houseroad: Directly write to the output please.

caffe2/python/operator_test/reduce_sum_op_test.py
```diff
@@ -0,0 +1,79 @@
...
+        def reduce_sum_ref(data, axis, keepdims):
+            return [np.sum(data, axis=axis, keepdims=keepdims)]
+
+        # all combinations of reduced axes
+        for n in range(1, 5):
+            for axes in it.combinations(range(4), n):
+                data = np.random.randn(d0, d1, d2, d3).astype(np.float32)
+
+                args = []
+                arg = pb2.Argument()
```
houseroad: no need to create pb2 version.
Check case here: https://github.com/caffe2/caffe2/blob/master/caffe2/python/operator_test/elementwise_op_broadcast_test.py#L50

caffe2/python/operator_test/reduce_mean_op_test.py
```diff
@@ -0,0 +1,79 @@
...
+        **hu.gcs_cpu_only)
+    def test_reduce_mean(self, d0, d1, d2, d3, keepdims, seed, gc, dc):
+        np.random.seed(seed)
+
+        def reduce_mean_ref(data, axis, keepdims):
+            return [np.mean(data, axis=axis, keepdims=keepdims)]
+
+        # all combinations of reduced axes
+        for n in range(1, 5):
+            for axes in it.combinations(range(4), n):
```
houseroad: This is probably overkill, (too many cases are generated), please reduce the number of tests.

caffe2/python/operator_test/reduce_mean_op_test.py
```diff
@@ -0,0 +1,79 @@
...
+                data = np.random.randn(d0, d1, d2, d3).astype(np.float32)
+
+                args = []
+                arg = pb2.Argument()
```
houseroad: We have better API to use. https://github.com/caffe2/caffe2/blob/master/caffe2/python/operator_test/elementwise_op_broadcast_test.py#L50

houseroad (changes requested):
We should lift the common parts out into separate class/file.

caffe2/python/operator_test/reduce_mean_op_test.py
zem7: For an input with 4 dimensions, the number of test is not too many (< 20). Do you recommend a smaller dimension, or just consider picking combination for specific value of "n" (wrt. line 49-50)?

houseroad:
Because this is a reduce op, you can map each element input to the corresponding element in the output.
For example, input[i][j][k] ==> output[i'][j'][k']
t'=0 if the t's axis is in the axes,
t'=t otherwise.

You should directly operate on the output, no need to have any buffer or copy input to the output.

At the beginning, you can compute the output shape, and how many input elements are reduced into one output element.

houseroad:
Also we should have Gradient ops for these two operators.

houseroad:
Further optimization on CPU implementation:

In some cases, you may try to call Eigen function:
https://eigen.tuxfamily.org/dox/group__TutorialReductionsVisitorsBroadcasting.html

houseroad:
Although, we may not targeting GPU implementation in this PR, I think it's still worth to check this function:
http://docs.nvidia.com/deeplearning/sdk/cudnn-developer-guide/index.html#cudnnReduceTensor

caffe2/operators/reduce_ops.cc
```diff
@@ -0,0 +1,146 @@
...
+  vector<TIndex> nd_idx(ndim);
+
+  for (TIndex i = ndim - 1; i >= 0 && index > 0; i--) {
+    nd_idx[i] = index % dims[i];
+    index /= dims[i];
+  }
+  return nd_idx;
+}
+
+TIndex ConvertToOutputIndex(
```
houseroad: Please write down how you convert the input index ==> output index. (In comments)

caffe2/operators/reduce_ops.cc
```diff
@@ -0,0 +1,146 @@
...
+    } else {
+      index += nd_idx[i] * mul;
+      mul *= dims[i];
+    }
+  }
+  return index;
+}
+
+template <typename T, class Context>
+void ComputeSum(
```
houseroad: Let's make computeSum more general, so it can support max/min/log/... later you don't need to refactor the code

caffe2/operators/reduce_ops.h
```diff
@@ -0,0 +1,119 @@
...
+      std::sort(axes_.begin(), axes_.end());
+      CAFFE_ENFORCE(axes_.front() >= 0, "Axes ids must be non-negative.");
+      CAFFE_ENFORCE(
+          axes_.back() < ndim,
+          "Axes ids must be smaller than the dimensions of input.");
+    }
+
+    auto& X = Input(0);
+    auto* Y = Output(0);
+    Y->ResizeLike(X);
```
houseroad: We can compute the size of Y, not directly set its size as X

caffe2/python/operator_test/reduce_ops_test.py
```diff
@@ -24,6 +24,54 @@
 import caffe2.python.hypothesis_test_util as hu
 import hypothesis.strategies as st
 import numpy as np
+import functools
+import itertools as it
+
+
+class TestReduceOps(hu.HypothesisTestCase):
+
+    def reduce_op_test(self, op_name, op_ref, data, axes, keepdims, device):
```
houseroad: this function can be defined inside test_reduce_sum_mean

houseroad (changes requested):
This one is much better, let's keep polishing it :-)

jamesr66a:
Can we fast track this? Not having this in C2 is blocking an important internal use case

caffe2/operators/reduce_ops.cc
```diff
@@ -91,12 +93,11 @@ bool ReduceMeanOp<T, Context>::Compute(
...
     vector<int>& axes,
     int keepdims) {
-  math::Set<T, Context>(X_size, 0.f, Y_data, &context_);
-  ComputeSum<T, Context>(X_data, X_size, dims, Y_data, axes, keepdims);
-  Output(0)->Resize(dims);
-  TIndex Y_size = Output(0)->size();
+  math::Set<T, Context>(Y_size, 0.f, Y_data, &context_);
+  ComputeOp<T, Context>(X_data, X_size, dims, Y_data, axes, keepdims, Add);
   math::Scale(
       Y_size, static_cast<T>(Y_size) / X_size, Y_data, Y_data, &context_);
```
houseroad: Just checked the code, you should not cast the second parameter of math::Scale to T (although we may only support float case), alpha must be float.

houseroad (commented):
One more thing to fix (see the inline comment), after fixing that (and also CI passes) then we are good land this PR.

caffe2/python/operator_test/reduce_ops_test.py
```diff
@@ -24,10 +24,53 @@
...
+class TestReduceOps(hu.HypothesisTestCase):
+    @given(
+        d0=st.integers(1, 5),
+        d1=st.integers(1, 5),
+        d2=st.integers(1, 5),
+        d3=st.integers(1, 5),
+        keepdims=st.integers(0, 1),
+        seed=st.integers(0, 2**32 - 1),
+        **hu.gcs_cpu_only)
+    def test_reduce_sum_mean(self, d0, d1, d2, d3, keepdims, seed, gc, dc):
```
houseroad, line 40: Later we should split this into two tests. But we can land this PR first.

houseroad (approved):
Let's land it first to unblock others. :-)

caffe2/operators/reduce_ops.cc
```diff
@@ -0,0 +1,147 @@
...
+ */
+
+#include "caffe2/operators/reduce_ops.h"
+
+namespace caffe2 {
+
+// For a Tensor X of n dimensions (dims[0], ..., dims[ndim-1]), given index
+// is converted to corresponding n-dimensional index, e.g. for X.shape = (2,
+// 3, 4) the linear index 12 maps to 3-dimensional index (1, 0, 0).
+vector<TIndex> ConvertFromInputIndex(TIndex index, vector<TIndex>& dims) {
```
dzhulgakov, line 24: Hey folks, this is pretty terrible performance to call the entire conversion (with vector allocation even, dang) on each element.

I highly doubt it makes sense to implement operators if their performance is that bad. Please do benchmark your ops and try to write more optimal implementation.

In this case we could actually proxy to Eigen - it seems to have an implementation for multi-axis reduce (thanks @jamesr66a for pointing it out)

houseroad: @dzhulgakov thanks for pointing out the problem. Yeah, we didn't do appropriate performance testing. Hi @zem7, @jamesr66a is working on this issue.

PR #2192, opened 2018-03-08
Fix ninja build in setuptools

bddppq:
setup.py directly invokes cmake


Maratyszcza (approved):
LGTM, thanks for the fix

PR #2202, opened 2018-03-08
Fix license header for GenerateProposalsOp.

orionr:
* Seems like this didn't get adjusted when the ops were open sourced.


orionr:
@caffe2bot retest this please

yinghai (approved):
Thanks.

PR #2207, opened 2018-03-08
update comments in segment_reduction_op

jspark1105:


caffe2/operators/segment_reduction_op.cc
```diff
@@ -19,6 +19,7 @@
 namespace caffe2 {
 
 // registering 4 input gradient with main output
```
kuttas: NumInputs is 5 here, but the comment says 4?

caffe2/operators/segment_reduction_op.h
```diff
@@ -1615,7 +1615,7 @@ class AbstractLengthsGradientOp : public Operator<Context> {
   }
 
   // Input layout:
-  //   orig_arg1, orig_arg2, ..., orig_argN, SEGMENT_GRADS, SEGMENT_IDS
+  //   orig_arg1, orig_arg2, ..., orig_argN, SEGMENT_GRADS, LENGTHS, INDICES
```
kuttas, line 1618: is INDICES optional/conditional? Should it be "[INDICES]" in the comment like it is further below in the file?

caffe2/operators/segment_reduction_op.h
```diff
@@ -1731,8 +1731,8 @@ class AbstractLengthsWithMainInputGradientOp : public Operator<Context> {
   }
 
   // Input layout:
-  //   orig_arg1, orig_arg2, ..., orig_argN, DATA_INPUT, SEGMENT_GRADS,
-  //      SEGMENT_LEGNTHS, [INDICES]
+  //   orig_arg1, orig_arg2, ..., orig_argN, SEGMENT_GRADS, LENGTHS,
+  //      DATA_INPUT, [INDICES]
   // orig_argXs represent original op's inputs and will be passed to the reducer
   // directly
   static constexpr int kNumInputs = ReducerGradient::originalInputs().size() +
```
kuttas, line 1738: Looks like if SparseFused and GradientNeedIndices are both set, there are actually 5 extra inputs?

kuttas (approved):
Minor questions/nits in the comment but looks good otherwise

caffe2/operators/segment_reduction_op.h
jspark1105: INDICES is not optional (required for GPU async version) even though it's not actually used in CPU version.

jspark1105: INDICES is not optional (required for GPU async version) even though it's not used in CPU version.

caffe2/operators/segment_reduction_op.h
jspark1105: Strangely, GradientNeedIndices is not used anywhere as far as I can tell. Let me merge this for now and look at this further.

PR #2213, opened 2018-03-09
Fix typo in CMakeLists build fix for Ninja

Maratyszcza:
The comment suggests that a special case should apply to either Ninja or Visual Studio, but the condition checks for both


houseroad (approved):
Thanks for the fix.

PR #2214, opened 2018-03-09
Avoid duplicated log when explicitly specified engine is not available

jspark1105:


bddppq (approved):
good catch

bddppq:
@caffe2bot test this please

PR #2216, opened 2018-03-09
Fix variable shadow warning

orionr:
title


ilia-cher (approved):

orionr:
Given that this is a *very* localized change and I'm not sure how tests are doing right now, I'm going to merge.

PR #2222, opened 2018-03-09
Fixes for build errors on Windows with GPU

pjh5:
- Macros and #ifs don't play well with Visual Studio 2015.
- That algoIndex_t isn't actually used anywhere and causes a "defined twice" error on Windows
- Windows complains about the automatic cast without the 'f'


pjh5:
Thanks to @kuttas for telling me how to fix these

kuttas (approved):
Change looks good to me. No idea why the old #if/endif code was breaking on windows, but the change preserves the original intent.

Issue #2151, opened 2018-03-06
Python dedent failed


MabinogiX:
### System information

* Operating system: win10
* Compiler version: VS 2015
* CMake version: 3.10.1
* CMake arguments: 
* Relevant libraries/versions (e.g. CUDA): cuda 8.0

### CMake summary output

I am installing caffe2 on win10 and it takes me almost a whole day. There are many missing argvs and errors in `build_windows.bat`. I doubt that the script has not been tested before.
Now I have another problem afer executing `build_windows.bat`:
ERROR Failed to remove indentation from:
"""

      from distutils import sysconfig
      print(sysconfig.get_python_lib(prefix=''))

"""
    Python dedent failed with error code: 系统找不到指定的文件。(system can not find specified file)
CMake Error at cmake/Utils.cmake:177 (message):
   Python dedent failed with error code: 系统找不到指定的文件。(system can not find specified file)
Call Stack (most recent call first):
  cmake/Utils.cmake:200 (dedent)
  caffe2/CMakeLists.txt:145 (pycmd)

Is anybody can help?

pjh5:
@Mabinogiysk what file can it not find? I've never seen this error on Windows before. It should be possible to build GPU on Windows (we've done it), but our coverage of different Windows configurations isn't that great right now.

Kasdision:
Hi, I have the same problem to you.
Are you chagning " -DBUILD_PYTHON=OFF^" to " -DBUILD_PYTHON=ON^", and then the error is coming?

bengaspar:
I have submitted a pull request to modify the utility function for dedent to fix this issue
https://github.com/caffe2/caffe2/pull/2228

Kasdision:
@bengaspar  Thanks to your help, the dedent utility function error has been solved, but it comes to a new problem:
```
“E:\caffe2\build\ALL_BUILD.vcxproj”(Default target) (1) ->
“E:\caffe2\build\binaries\caffe2_benchmark.vcxproj”(Default target) (10) ->
“E:\caffe2\build\caffe2\caffe2_gpu.vcxproj”(Default target) (11) ->
(ClCompile 目标) ->
  E:\caffe2\caffe2\operators\conv_op_cudnn.cc(199): error C2121: “#”: Invalid character: may be the result of macro expansion[E:\caffe2\build\caffe2\caffe2_g
pu.vcxproj]
[...]

PR #2228, opened 2018-03-10
Fixed issue with dedent utility function when building with python win10

bengaspar:
Fix for build issue in dedent utility function when building with python
https://github.com/caffe2/caffe2/issues/2151
It isn't entirely clear to me why we would need to pipe that through std in via the command line vs using the first arg.  Please review and let me know if this can cause any side effects.  Thank you for the review.

### DIFF
--- a/cmake/Utils.cmake
+++ b/cmake/Utils.cmake
@@ -163,12 +163,10 @@ function(dedent outvar text)
   else()
     set(_python_exe "${PYTHON_EXECUTABLE}")
   endif()
-  set(_fixup_cmd "import sys; from textwrap import dedent; print(dedent(sys.stdin.read()))")
-  # Use echo to pipe the text to python's stdinput. This prevents us from
-  # needing to worry about any sort of special escaping.
+  set(_fixup_cmd "import sys; from textwrap import dedent; print(dedent(sys.argv[1]));")
+  # Send script through command line arg
   execute_process(
-    COMMAND echo "${text}"
-    COMMAND "${_python_exe}" -c "${_fixup_cmd}"
+	COMMAND "${_python_exe}" -c "${_fixup_cmd} " "${text}"
     RESULT_VARIABLE _dedent_exitcode
     OUTPUT_VARIABLE _dedent_text)
   if(NOT ${_dedent_exitcode} EQUAL 0)


pjh5:
Thanks for doing this. Can you rebase off of master to remove all the changes in caffe2/operators/conv_op_cudnn.cc  , which have been checked in in #2222 ?

We don't have CI set up yet to test this, but I'll validate that this works tomorrow morning when I'm back at my Windows desktop.

scripts/build_windows.bat
```diff
@@ -8,14 +8,15 @@
 setlocal
 
 SET ORIGINAL_DIR=%cd%
-SET CAFFE2_ROOT=%~dp0%..
+SET CAFFE2_ROOT=%~dp0..
+SET PYBIND_DIR="%CAFFE2_ROOT%\third_party\pybind11\include"
```
pjh5: Is this needed for the dedent error is this an unrelated change?

pjh5:
I think you'll have to add https://github.com/caffe2/caffe2.git as an upstream remote of your fork and then fetch that and rebase off of it's master in order to update this diff's base commit. I think it
`git remote add upstream https://github.com/caffe2/caffe2.git`
`git fetch upstream`
`git rebase upstream/master`

scripts/build_windows.bat
bengaspar: No, this was just a pybind11 issue I was having, seemed to fix it but I'm not sure if its just my environment.

bengaspar:
Sorry about the mess, first PR.  The diff should only be my change at this point.  Going forward I will be sure  to isolate the PR changes into a separate branch away from my other changes.  Thanks for helping me learn this process.

orionr:
@caffe2bot test this please

Erotemic:
@bengaspar as the original comment indicates, the use of stdin/stdout was to prevent needing to escape special characters if you were to place the text directly in the command. However, I think your solution also addresses those issues. So, it shouldn't cause any side effects. I simply didn't consider using `sys.argv`. I think your solution is cleaner better than my original code. Thanks for submitting the patch.  

I still would like to know what the actual issue was. I was never able to reproduce it on my windows box. Under what circumstances does windows have trouble piping stdout?

yinghai:
@caffe2bot retest this please

orionr:

PR #2248, opened 2018-03-13
Caffe2-onnx exporter

yinghai:
This PR implements the caffe2->onnx exporter in C++ for majority of the convnet related ops. We register the `InheritOnnxNode` and its conversion function in each of the c2 op schema. 

Test Plan:
Passing `with-proxy pytest -s c2_ref_test.py::TestCaffe2End2End::test_resnet50`


yinghai:
This PR is ready for review. 
- Split out the `InheritOnnxNode` schema change (https://github.com/caffe2/caffe2/pull/2366)
- Remove the hard dependency of Caffe2 on ONNX

Maratyszcza:
@yinghai Is it WIP, or ready for review?

yinghai:
@Maratyszcza It's ready. :)

caffe2/onnx/helper.cc
```diff
@@ -44,4 +44,27 @@ void DummyName::Reset(const std::unordered_set<std::string> &used_names) {
...
+  node.set_op_type(type);
+  for (const auto& input: inputs) {
+    node.add_input(input);
+  }
+  for (const auto& output: outputs) {
+    node.add_output(output);
+  }
+  for (const auto& attr: attributes) {
+    auto* tmp = node.add_attribute();
+    tmp->CopyFrom(attr);
```
Maratyszcza: Why do you need the temporary? If you want to check it was successfully created, you can do e.g.
```c++
auto* new_node = node.add_attribute();
CHECK_NOTNULL(new_node);
new_node->CopyFrom(attr);
```
Otherwise, you can just do one-liner `node.add_attribute()->CopyFrom(attr);`

yinghai: I'll do one liner.

caffe2/onnx/helper.h
```diff
@@ -16,24 +16,77 @@
...
 #include <unordered_set>
 
-namespace caffe2 { namespace onnx {
+namespace caffe2 {
+namespace onnx {
+
+using ::ONNX_NAMESPACE::AttributeProto;
+using ::ONNX_NAMESPACE::NodeProto;
+
 class DummyName {
```
Maratyszcza: Without documentation, a class with such a name looks strange

caffe2/onnx/helper.h
```diff
@@ -16,24 +16,77 @@
...
-   private:
-     static std::unordered_set<std::string>& get_used_names();
-     static size_t counter_;
+ private:
+  static std::unordered_set<std::string>& get_used_names();
+  static size_t counter_;
 };
 
-}}
+inline AttributeProto MakeAttribute(
```
Maratyszcza, line 47: This functionality is similar to `utils/proto_utils.h`, albeit works with ONNX messages instead of Caffe2 messages. Maybe they could be unified?

caffe2/onnx/onnx_exporter.cc
```diff
@@ -0,0 +1,585 @@
...
+#include "caffe2/proto/caffe2_legacy.pb.h"
+#include "caffe2/utils/map_utils.h"
+
+#include <unordered_set>
+
+namespace caffe2 {
+namespace onnx {
+
+namespace {
+void ApplyTrans(
```
Maratyszcza: This function needs documentation: it is totally unclear what it does

caffe2/onnx/onnx_exporter.cc
```diff
@@ -0,0 +1,585 @@
...
+  }
+
+  if (!vals.empty() && !global) {
+    attrs->emplace(ks2, MakeAttribute(ks2, vals));
+  }
+}
+
+int64_t DimProd(const caffe2::TensorShape& shape, int start, int end) {
+  int64_t acc = 1;
+  for (int i = start; i < end; ++i) {
```
Maratyszcza, line 87: We target C++11, so can also do
```
for (int dim : shape.dims()) {
  acc *= dim;
}
```

caffe2/onnx/helper.h
yinghai: I'll add document. On the other hand, I'll change this class in another PR as current design is using static variables which is not good.

caffe2/onnx/onnx_exporter.cc
yinghai: `TensorShape` supports iterators?

caffe2/onnx/onnx_exporter.cc
```diff
@@ -0,0 +1,585 @@
...
+  return acc;
+}
+
+TensorProto CreateOnnxShapeTensor(const std::vector<int64_t>& shape) {
+  TensorProto tensor;
+  tensor.set_name(DummyName::NewDummyName());
+  tensor.set_data_type(TensorProto::INT64);
+  tensor.add_dims(shape.size());
+  tensor.mutable_raw_data()->assign(
+      reinterpret_cast<const char*>(shape.data()), sizeof(int64_t) * shape.size());
```
Maratyszcza, line 99: The prototype for the method is `void set_raw_data(const void* value, size_t size)`, so a better implementation would be
```
tensor.mutable_raw_data()->assign(
      static_cast<const void*>(shape.data()),
      static_cast<size_t>(sizeof(int64_t) * shape.size()));
```
Cast to `const char*` breaks strict aliasing rules, and can disable some compiler optimizations.

caffe2/onnx/onnx_exporter.cc
yinghai: I'll ping @houseroad. :)

caffe2/onnx/helper.h
yinghai: Let me check. I still prefer to have onnx part separated in case there is subtle mismatch between `caffe2::Argument` and `onnx::Attribute`.

caffe2/onnx/onnx_exporter.cc
```diff
@@ -0,0 +1,585 @@
...
+  return tensor;
+}
+} // namespace
+
+const std::unordered_map<std::string, std::string>&
```
Maratyszcza, line 104: What is the purpose of having functions to get all these maps instead of making them global static objects? Every time you call these functions, they create copies of map + all strings in keys and values. Thats a lot of memory allocations.

caffe2/onnx/onnx_exporter.cc
```diff
@@ -0,0 +1,585 @@
...
+      };
+  return kSpecialOperators;
+}
+
+void OnnxExporter::CopyCaffe2ArgToOnnxAttr(
+    AttributeProto* attr,
+    const std::string& op_type,
+    const caffe2::Argument& arg) {
+  std::string name;
+  if (get_per_op_renamed_attrs().count(op_type)) {
```
Maratyszcza: If you use the value when its present, it is better to do `std::map<T>.find()` and compare the result to `std::map<T>.end()` to find if the key is present in the map. If it is, you can rename the iterator returned by `std::map<T>.find()`. Thus you will search the map only once, not twice.

yinghai: Got it. I noticed this but since the map is very small I didn't bother. I can change it back.

caffe2/onnx/onnx_exporter.cc
```diff
@@ -0,0 +1,585 @@
...
+  } else if (arg.strings_size()) {
+    attr->mutable_strings()->CopyFrom(arg.strings());
+    attr->set_type(AttributeProto::STRINGS);
+  } else {
+    CAFFE_THROW(
+        caffe2::MakeString("Unsupported Caffe2 argument: ", arg.name()));
+  }
+}
+
+bool OnnxExporter::IsBlaskListed(const caffe2::Argument& arg) {
```
Maratyszcza: Typo IsBla**s**kListed -> IsBlackListed

caffe2/onnx/onnx_exporter.cc
yinghai: Why would it create a copy?

caffe2/onnx/onnx_exporter.cc
```diff
@@ -0,0 +1,585 @@
...
+    }
+  }
+
+  return false;
+}
+
+ConvertedResult OnnxExporter::Caffe2OpToOnnxNodes(
+    const caffe2::OperatorDef& def,
+    const std::unordered_map<std::string, caffe2::TensorShape>& shapes) {
+  std::string type = def.type();
```
Maratyszcza, line 225: You create a copy of the string here, but it doesn't seem necessary. Please use `const string& type = def.type()` instead.

caffe2/onnx/onnx_exporter.cc
```diff
@@ -0,0 +1,585 @@
...
+  const auto& renamed_op_lut = get_renamed_operators();
+  const auto it = renamed_op_lut.find(type);
+  if (it != renamed_op_lut.end()) {
+    type = it->second;
+  }
+  if (get_special_operators().count(type)) {
+    return (this->*get_special_operators().at(type))(def, shapes);
```
Maratyszcza: This took a while to parse. Would be better to split into separate table lookup and method call statements.

caffe2/onnx/onnx_exporter.cc
yinghai: I see. It's protobuf. :)

caffe2/onnx/onnx_exporter.cc
yinghai: I think by using the iterator with 1 lookup, we can break this down naturally. :)

caffe2/onnx/onnx_exporter.cc
Maratyszcza: Right, I didn't notice they return a reference.

caffe2/operators/channel_shuffle_op.cc
```diff
@@ -37,7 +37,8 @@ REGISTER_GRADIENT(ChannelShuffle, GetChannelShuffleGradient);
 OPERATOR_SCHEMA(ChannelShuffle)
     .IdenticalTypeAndShape()
     .NumInputs(1)
-    .NumOutputs(1);
+    .NumOutputs(1)
+    .InheritOnnxSchema("ChannelShuffle");
```
Maratyszcza, line 41: Does it create a non-optional dependency on ONNX schema in Caffe2 schema?

caffe2/onnx/onnx_exporter.cc
```diff
@@ -0,0 +1,585 @@
...
+    const caffe2::OperatorDef& def) {
+  ConvertedResult result;
+  auto& nodes = result.first;
+  nodes.emplace_back();
+  NodeProto& node = nodes.back();
+  node.set_name(def.name());
+  node.set_op_type(
+      caffe2::get_default(get_renamed_operators(), def.type(), def.type()));
+  for (const auto& i : def.input()) {
+    node.add_input(i);
```
Maratyszcza, line 250: IIRC, Caffe2 operators can overwrite previous outputs, but ONNX operators can not. You may need extra logic to handle these situations.

caffe2/onnx/onnx_exporter.cc
```diff
@@ -0,0 +1,585 @@
...
+      // to replace legacy_pad. pad[end] = output_size[start + 2] *
+      // stride[start] - pad[start] - 1 + kernel[start] - input[start + 2] end =
+      // start + len(pad) / 2
+      std::cerr << "Warning: Converting legacy padding to explicit padding."
+                << std::endl;
+      auto* pads_attr = attrs.at("pads").mutable_ints();
+      auto& strides_attr = attrs.at("strides").ints();
+      auto& kernel_shape_attr = attrs.at("kernel_shape").ints();
+      for (int i = 0; i < 2; ++i) {
+        int64_t tmp = output_size.dims(i + 2) * strides_attr.Get(i) -
```
Maratyszcza: Please use a more descriptive name than `tmp`

caffe2/onnx/onnx_exporter.cc
```diff
@@ -0,0 +1,585 @@
...
+      attrs.emplace("auto_pad", MakeAttribute("auto_pad", "SAME_UPPER"));
+    } else if (
+        legacy_pad_attr.i() ==
+        static_cast<int64_t>(caffe2::LegacyPadding::CAFFE_LEGACY_POOLING)) {
+      // The problem here is that, Pool op in Caffe may add an additional pixel,
+      // if the last part is smaller than stride. So we use the explicit padding
+      // to replace legacy_pad. pad[end] = output_size[start + 2] *
+      // stride[start] - pad[start] - 1 + kernel[start] - input[start + 2] end =
+      // start + len(pad) / 2
+      std::cerr << "Warning: Converting legacy padding to explicit padding."
```
Maratyszcza: Should be `LOG(WARNING) << "Converting legacy padding to explicit padding."`;

caffe2/onnx/onnx_exporter.cc
```diff
@@ -0,0 +1,585 @@
...
+      g = arg.i();
+      break;
+    }
+  }
+  CAFFE_ENFORCE(g && c % g == 0);
+  ConvertedResult result;
+  auto& nodes = result.first;
+  auto& const_tensors = result.second;
+
+  const auto tmp1 = DummyName::NewDummyName();
```
Maratyszcza: More descriptive names would be great!

caffe2/onnx/onnx_exporter.cc
```diff
@@ -0,0 +1,585 @@
...
+  auto x = def.input(0);
+  auto w = def.input(1);
+  const auto& b = def.input(2);
+  const auto& y = def.output(0);
+  const auto& x_shape = shapes.at(x);
+
+  ConvertedResult result;
+  auto& nodes = result.first;
+  auto& const_tensors = result.second;
+  std::unordered_map<std::string, const caffe2::Argument*> args;
```
Maratyszcza, line 535: See `ArgumentHelper` in `utils/proto_utils.h`

caffe2/onnx/onnx_exporter.h
```diff
@@ -0,0 +1,102 @@
...
+
+ public:
+  ConvertedResult Caffe2OpToOnnxNodes(
+      const caffe2::OperatorDef& def,
+      const std::unordered_map<std::string, caffe2::TensorShape>& shapes);
+
+ private:
+  ConvertedResult CommonCaffe2OpToOnnxNodes(const caffe2::OperatorDef& def);
+
+  ConvertedResult CreateConvPoolNodes(
```
Maratyszcza, line 53: why these methods are not static?

Maratyszcza (approved):
LGTM, great work! Few fixes, and ready to ship

caffe2/onnx/onnx_exporter.cc
yinghai: Yes. That's why I added the comment that the input NetDef should be SSA rewritten form already. Currently, the SSA rewriting part is in Python, I plan to move that into C++.

caffe2/operators/channel_shuffle_op.cc
yinghai: It doesn't. It's just the string property. Caffe2 can still be built without ONNX.

caffe2/onnx/onnx_exporter.h
yinghai: Hmm, no particular reason.

caffe2/onnx/onnx_exporter.cc
```diff
@@ -0,0 +1,585 @@
...
+  if (node.op_type() == "MaxPool" || node.op_type() == "AveragePool") {
+    auto it = attrs.find("global_pooling");
+    if (it != attrs.end() && it->second.has_i() && it->second.i()) {
+      node.set_op_type("Global" + node.op_type());
+      attrs.erase(it);
+    }
+  }
+
+  bool global = (node.op_type().find("Global") == 0);
+  ApplyTrans(&attrs, global, "kernel", 2, "kernel_shape");
```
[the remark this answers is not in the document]
yinghai: Right now, it doesn't.

caffe2/onnx/onnx_exporter.cc
yinghai: Actually, I cannot do this because `type` may be written later.

caffe2/onnx/onnx_exporter.cc
yinghai: `tensor.mutable_raw_data()` returns `std:string*` and its `assign` method takes `const char*`. Passing `const void*` will compile....

caffe2/onnx/onnx_exporter.cc
yinghai: I see. I actually don't want to make an copy of the `Argument`s here.

caffe2/onnx/onnx_exporter.cc
yinghai: I cannot do this right here because it doesn't iterate the whole range of `shape.dims()`.

PR #2250, opened 2018-03-13
[C2] Support optional lengths input to ReduceFront/Back operators

enosair:
If ops are specified a second input, that would be tensor of ints specifying length of each sample. Ignore entries beyond the lengths when computing the reduction.

### DIFF
--- a/caffe2/operators/reduction_front_back_ops.cc
+++ b/caffe2/operators/reduction_front_back_ops.cc
@@ -30,10 +30,12 @@ void SumReduceDimsOp<CPUContext, true, false>::Compute(
     int rows,
     int cols,
     const T* in_data,
+    const int32_t* lengths_data,
     T* out_data) {
   for (int j = 0; j < cols; j++) {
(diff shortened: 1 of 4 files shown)


orionr (approved):
Nicely done!

caffe2/operators/reduction_front_back_ops.cc
```diff
@@ -65,9 +69,16 @@ void SumReduceDimsGradientOp<CPUContext, true, false>::Compute(
     int rows,
     int cols,
     const T* dYdata,
+    const int* lengths_data,
     T* dXdata) {
   for (int i = 0; i < rows * cols; i++) {
-    dXdata[i] = dYdata[i % cols];
+    int row = i / cols;
+    int col = i % cols;
+    if (lengths_data == nullptr or row < lengths_data[col]) {
```
yinghai, line 77: Please use old style `||`, as `or` is not supported by MSVC.

Issue #2191, opened 2018-03-08
Compile custom operator failed in Detectron


7oud:
### Detailed steps to reproduce
1. Compile caffe2 source code and install it into conda environment
2. Test caffe2 OK, when running `python2 -c 'from caffe2.python import workspace; print(workspace.NumCudaDevices())'`
3. Install other dependencies and detectron
4. Detectron tests pass
```
I0307 21:53:58.443390 23369 operator.cc:173] Operator with engine CUDNN is not available for operator SpatialNarrowAsGradient.
----------------------------------------------------------------------
Ran 3 tests in 0.805s

OK
```
5. Complie custom op **FAILED**, I am not sure it is the problem of my env or a bug in detectron or caffe2, it seems some similar situation [cocodataset/cocoapi#94 (comment)](https://github.com/cocodataset/cocoapi/issues/94#issuecomment-350152998)
```
(caffe2) zoud@ubu1604:~/Workspace/Github/facebookresearch/detectron/lib$ make ops
mkdir -p build && cd build && cmake .. && make -j4
-- Caffe2: Found gflags with new-style gflags target.
-- Caffe2: Cannot find glog automatically. Using legacy find.
...
-- CUDA detected: 8.0
CMake Error at cmake/Cuda.cmake:32 (message):
  Invalid CUDA_ARCH_NAME, supported values: Kepler, Maxwell, Pascal, Volta,
  All, Manual.  Got Auto
Call Stack (most recent call first):
  cmake/Cuda.cmake:183 (caffe2_select_nvcc_arch_flags)
  cmake/Dependencies.cmake:11 (include)
  CMakeLists.txt:14 (include)

-- Configuring incomplete, errors occurred!
See also "/home/zoud/Workspace/Github/facebookresearch/detectron/lib/build/CMakeFiles/CMakeOutput.log".
See also "/home/zoud/Workspace/Github/facebookresearch/detectron/lib/build/CMakeFiles/CMakeError.log".
Makefile:13: recipe for target 'ops' failed
make: *** [ops] Error 1

```
**ERROR** info in CMakeError.log, it seems pthread can not be found
```
Linking C executable cmTC_ec809
/usr/bin/cmake -E cmake_link_script CMakeFiles/cmTC_ec809.dir/link.txt --verbose=1
/usr/bin/cc   -std=c11     CMakeFiles/cmTC_ec809.dir/CheckSymbolExists.c.o  -o cmTC_ec809 -rdynamic 
[...]

PR #2251, opened 2018-03-13
bump version to 0.8.2

Yangqing:
This is in order to update the detectron package that is showing dependency issues, e.g. https://github.com/caffe2/caffe2/issues/2191

Caffe2 now ships a proper set of CMake files for dependent libraries' CMake build scripts. This version bump can make things easier for dependent libraries to check if it is having supported caffe2 cmake files.

cc @orionr


orionr (approved):

orionr:
@caffe2bot retest this please

PR #2269, opened 2018-03-14
Caffe2 Profiles

Maratyszcza:
Enable dynamic modification of OperatorDef attributes based on platform or SoC name

### DIFF
--- a/caffe2/core/profiles.cc
+++ b/caffe2/core/profiles.cc
@@ -1,0 +1,207 @@
+/**
+ * Copyright (c) 2016-present, Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <string>
+#include <map>
+
+#ifdef __APPLE__
+#include <TargetConditionals.h>
+#endif
+
+#include "caffe2/core/profiles.h"
+#include "caffe2/core/logging.h"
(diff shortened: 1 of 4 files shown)


Maratyszcza:
@caffe2bot retest this please

caffe2/core/profiles.cc
```diff
@@ -0,0 +1,220 @@
...
+    const cpuinfo_package* package = cpuinfo_get_package(0);
+    return std::string(package->name,
+      strnlen(package->name, CPUINFO_PACKAGE_NAME_MAX));
+  }
+};
+
+/**
+ * Checks if @a string starts with @a prefix.
+ */
+inline bool startsWith(
```
yinghai: Move to `utils/string_utils.h`?

Maratyszcza: Done!

Maratyszcza:
@caffe2bot retest this please

Yangqing (changes requested):
I think in offline discussions folks are having quite nontrivial concerns about this design, so we probably want to iterate on this before landing. cc @salexspb @ajtulloch @bwasti .

caffe2/proto/caffe2.proto
```diff
@@ -219,6 +270,11 @@ message NetDef {
   // blobs' contents may be overwritten.
   repeated string external_input = 7;
   repeated string external_output = 8;
+
+  // List of profiles that modify operator arguments based on characteristics
+  // of the target system. Profiles are applied in the order in which they
+  // are listed.
+  repeated Profile profile = 9;
```
salexspb, line 277: I personally feel this is a hack. Protobufs are meant to be very rarely updated. The Predicate concept is very vague. We will have to keep adding stuff there if we decide to go with protobuf changing approach. So I would discourage emulating predicates on the protobuf level. 

I wonder what are technical difficulties you see when you implement this without any changes to the protobuf. So far engines + runtime dispatch based on the available hardware worked. You just need to write dispatch logic on the library code level, but not on the serialization level. For example, check out perfkernels code, this is dispatch Dima and Yangqing added last year to support fast embedding lookup kernels from Misha. 

https://github.com/caffe2/caffe2/tree/01827c153db96349745a544148c1ff0386c5ef9e/caffe2/perfkernels

You could also think about this as a technical challenge - write the code in the way that dispatch works at runtime based on the hardware you have. But without such heavy changes to the protobuf.

PR #2270, opened 2018-03-14
implement ReshapeDynamic

anderspapitto:


caffe2/python/onnx/backend.py
```diff
@@ -938,6 +939,9 @@ def _create_reshape(cls, init_model, pred_model, n, opset_version):
         c2_op.output.append(dummy_name())
         return c2_op
 
+    def _create_reshape_dynamic(cls, init_model, pred_model, n, opset_version):
```
bddppq: why can not use `_create_reshape`?

anderspapitto: well, it does something different (provide shape as input or as argument). it could have also been done with some `if` statements inside `_create_reshape`

bddppq: But the differences are in the NodeProto, not in the conversion right? Looks to me both are doing same special handling (i.e. adding an extra output in the c2 op).

anderspapitto: you're right

PR #2273, opened 2018-03-15
Fix ONNX backend for MatMul

jamesr66a:
The Caffe2 `BatchMatMul` op requires `broadcast=1` to match ONNX semantics

### DIFF
--- a/caffe2/python/onnx/backend.py
+++ b/caffe2/python/onnx/backend.py
@@ -159,7 +159,6 @@ class Caffe2Backend(Backend):
         'Neg':                   'Negative',
         'BatchNormalization':    'SpatialBN',
         'InstanceNormalization': 'InstanceNorm',
-        'MatMul':                'BatchMatMul',
         'Upsample':              'ResizeNearest',
         'Identity':              'Copy',
         'InstanceNormalization': 'InstanceNorm',
@@ -201,6 +200,7 @@ class Caffe2Backend(Backend):
         'RNN': '_create_rnn',
         'Sqrt': '_create_sqrt',
         'Reciprocal': '_create_reciprocal',
+        'MatMul': '_create_matmul',
     }
 
     # NB: By default, you will use the LATEST definition of the operator,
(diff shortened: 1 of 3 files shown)


jamesr66a:
@onnxbot retest this please

yinghai (commented):
Sorry, here is the awkward news. You may need to change the python part of the code too. Right now we cannot completely get rid of the Python flow because of dependencies. I'll make a PR to do so after next week....

jamesr66a:
@yinghai Ported changes to Python code

caffe2/python/onnx/backend.py
```diff
@@ -960,6 +960,11 @@ def _create_reciprocal(cls, init_model, pred_model, n, opset_version):
             exponent=-1.0,
         )
 
+    @classmethod
+    def _create_matmul(cls, init_model, pred_model, n, opset_version):
+        n.attrs['broadcast'] = 1
+        return cls._common_onnx_node_to_caffe2_op(init_model, pred_model, n, opset_version)
```
yinghai: We have `CommonOnnxNodeToCaffe2Ops` in C++ implementation, too. It might save you some code. 

https://github.com/caffe2/caffe2/blob/master/caffe2/onnx/backend.cc#L437

yinghai (approved):
I don't see blockers but see if you can use `CommonOnnxNodeToCaffe2Ops` to simplify your c++ code. 

And I will remove the python version soon. It's not good to have two implementations. Thanks a lot!

caffe2/python/onnx/backend.py
```diff
@@ -960,6 +960,11 @@ def _create_reciprocal(cls, init_model, pred_model, n, opset_version):
             exponent=-1.0,
         )
 
+    @classmethod
+    def _create_matmul(cls, init_model, pred_model, n, opset_version):
+        n.attrs['broadcast'] = 1
```
bddppq: This is mutating the original onnx node. Should get the c2 op first and then mutate the c2 op.

jamesr66a:
Triggering re-test

PR #2304, opened 2018-03-18
Add support for several auto-created observers

salexspb:
This allows to support several auto-attached observers.

This is a sync of an internal diff.


salexspb:
Going to merge as this was already reviewed internally.

PR #2309, opened 2018-03-19
Use nth_element in GenerateProposals for performance

orionr:
I'd like to find a way to reliably test that this improves performance, but when I run `make test` it at least doesn't seem to degrade.

### DIFF
--- a/caffe2/operators/generate_proposals_op.cc
+++ b/caffe2/operators/generate_proposals_op.cc
@@ -171,16 +171,19 @@ void GenerateProposalsOp<CPUContext>::ProposalsForOneImage(
   // 3. remove predicted boxes with either height or width < min_size
   auto keep = utils::filter_boxes(proposals, min_size, im_info);
   DCHECK_LE(keep.size(), scores.size());
-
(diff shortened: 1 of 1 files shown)


caffe2/operators/generate_proposals_op.cc
```diff
@@ -171,17 +171,20 @@ void GenerateProposalsOp<CPUContext>::ProposalsForOneImage(
   // 3. remove predicted boxes with either height or width < min_size
   auto keep = utils::filter_boxes(proposals, min_size, im_info);
   DCHECK_LE(keep.size(), scores.size());
-
-  // 4. sort all (proposal, score) pairs by score from highest to lowest
-  // 5. take top pre_nms_topN (e.g. 6000)
-  std::sort(keep.begin(), keep.end(), [&scores](int lhs, int rhs) {
```
yinghai, line 177: Maybe you can use `std::partial_sort` to replace `std::nth_element` + `std::sort`. 

http://en.cppreference.com/w/cpp/algorithm/partial_sort

PR #2313, opened 2018-03-19
Make InputSize and OutputSize const member function

jspark1105:
Make InputSize and OutputSize const member function


bddppq (approved):
LG

bddppq:
@onnxbot test this please

PR #2316, opened 2018-03-19
[C2] Fix the check of current scope in optimizer

enosair:
scope.CurrentDeviceScope() can return a None type, which was not considered.


ffjiang (approved):
Looks good!

PR #2324, opened 2018-03-20
Skip the test average pool same mode tests

houseroad:
Let's skip all the average pool tests evolved same mode.

I have added support in another PR. Let's land this first, and enable same_upper in the other one.

### DIFF
--- a/caffe2/python/onnx/tests/onnx_backend_test.py
+++ b/caffe2/python/onnx/tests/onnx_backend_test.py
@@ -35,7 +35,8 @@ backend_test = onnx.backend.test.BackendTest(c2, __name__)
 
 backend_test.exclude(r'(test_hardsigmoid'  # Does not support Hardsigmoid.
                      '|test_mean|test_hardmax'  # Does not support Mean and Hardmax.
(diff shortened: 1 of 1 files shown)


houseroad:
Please land this PR before the https://github.com/onnx/onnx/pull/572 is merged into ONNX repo. Otherwise, our CI will break.

bddppq (approved):

caffe2/python/onnx/tests/onnx_backend_test.py
```diff
@@ -35,7 +35,8 @@
 
 backend_test.exclude(r'(test_hardsigmoid'  # Does not support Hardsigmoid.
                      '|test_mean|test_hardmax'  # Does not support Mean and Hardmax.
-                     '|test_cast.*FLOAT16.*)')  # Does not support Cast on Float16.
+                     '|test_cast.*FLOAT16.*'  # Does not support Cast on Float16.
+                     '|test_averagepool_.*same.*)')  # Does not support Cast on Float16.
```
bddppq, line 39: tbh initially when I added these `include` `exclude` thing to the backend test suite I imagined people will have multiple invocations with each only contains one pattern, but turns out you folks are too good at using regular expression 😂

PR #2325, opened 2018-03-20
Support ONNX SAME_UPPER

houseroad:
1) Add support for SAME_UPPER. Looks like Caffe2 does not support SAME_LOWER by default.

2) Fix the bug in C++ ONNX C2 backend. If an attribute is renamed to empty string, we should delete it.

3) ~~Looks like we have some CuDNN fall back issue, will fix it.~~

### DIFF
--- a/caffe2/onnx/backend.cc
+++ b/caffe2/onnx/backend.cc
@@ -5,6 +5,8 @@
 #include "caffe2/onnx/helper.h"
 #include "caffe2/utils/map_utils.h"
 #include "caffe2/utils/proto_utils.h"
+#include "caffe2/proto/caffe2_legacy.pb.h"
+#include "onnx/defs/schema.h"
 
 #if !CAFFE2_MOBILE
 #include "onnx/checker.h"
@@ -190,6 +192,17 @@ float OnnxAttributes::get(const std::string& key) const {
   return value;
 }
 
+template <> const std::string*
+OnnxAttributes::get(const std::string &key) const {
+  const std::string* value;
+  const auto it = onnx_attrs_.find(key);
+  if (it != onnx_attrs_.end()) {
+    const AttributeProto &attr = *it->second;
+    value = &attr.s();
+  }
+  return value;
+}
+
 template <>
 ::google::protobuf::RepeatedPtrField<std::string> OnnxAttributes::get(
     const std::string& key) const {
@@ -235,6 +248,9 @@ OnnxAttributes::OnnxAttrToCaffe2Arg(
     const auto& attr = rewritten_onnx_attrs_.count(kv.first)
         ? rewritten_onnx_attrs_.at(kv.first)
         : (*kv.second);
+    if (mapper(attr.name()).empty()) {
+      continue;
+    }
     auto* arg = args.Add();
     arg->set_name(mapper(attr.name()));
     CopyOnnxAttrValueToCaffe2Arg(arg, attr);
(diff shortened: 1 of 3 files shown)


caffe2/onnx/backend.cc
```diff
@@ -304,7 +319,7 @@ Caffe2Backend::get_renamed_operators() const {
 const std::unordered_map<std::string, std::string>&
 Caffe2Backend::get_renamed_attrs() const {
   const static std::unordered_map<std::string, std::string> kRenamedAttrs{
-      {"kernel_shape", "kernels"}};
+      {"kernel_shape", "kernels"}, {"auto_pad", ""}};
```
bddppq: I think you don't need this, in the `CreateConvePoolOpBase` you can remove the `auto_pad` attribute before calling `CommonOnnxNodeToCaffe2Ops`.

caffe2/onnx/backend.cc
```diff
@@ -199,6 +200,17 @@ template <> float OnnxAttributes::get(const std::string &key) const {
   return value;
 }
 
+template <> const std::string*
+OnnxAttributes::get(const std::string &key) const {
+  const std::string* value;
```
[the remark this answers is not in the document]
houseroad: Good catch

caffe2/onnx/backend.cc
houseroad: Sure, done :-)

caffe2/onnx/backend.cc
```diff
@@ -244,6 +256,9 @@ OnnxAttributes::OnnxAttrToCaffe2Arg(
     const auto &attr = rewritten_onnx_attrs_.count(kv.first)
                            ? rewritten_onnx_attrs_.at(kv.first)
                            : (*kv.second);
+    if (mapper(attr.name()).empty()) {
```
bddppq: this is not needed anymore

bddppq (commented):
Unfortunately right now this change also needs to mirrored to the python side. @yinghai to confirm.

caffe2/onnx/backend.cc
houseroad: For Upsample, we still need it: https://github.com/caffe2/caffe2/blob/master/caffe2/onnx/backend.cc#L319

If we just want to remove an attribute during onnx to caffe2 conversion, probably this handling is the easiest.

caffe2/onnx/backend.cc
```diff
@@ -199,6 +200,17 @@ template <> float OnnxAttributes::get(const std::string &key) const {
   return value;
 }
 
+template <> const std::string*
```
yinghai, line 195: Any reason why we return a pointer instead of a copy or a const reference?

houseroad: For the performance purpose, I avoid returning copy. 
To be consistent with TensorProto* case at line 239, use const pointer here.

caffe2/onnx/backend.cc
```diff
@@ -411,6 +426,25 @@ Caffe2Ops Caffe2Backend::CreateConvePoolOpBase(const ModelProto &init_model,
     attr->set_i(1);
   }
 
+  if (attributes.HasAttribute("auto_pad")) {
+    auto auto_pad = attributes.get<const std::string*>("auto_pad");
```
yinghai, line 419: `CHECK(auto_pad);`?

yinghai (approved):
No real blockers. Just note the comments.

yinghai:
https://github.com/caffe2/caffe2/pull/2362 removed the python code. We can just focus on C++ code. :)

caffe2/onnx/backend.cc
bddppq: alright add a TODO to create another explicit list of attributes that we want to remove, like `_blacklist_caffe2_args` in the frontend

caffe2/onnx/backend.cc
houseroad: Do we have a `CHECK` macro? I didn't quite get this.

caffe2/onnx/backend.cc
houseroad: Sure, I will do that.

PR #2306, opened 2018-03-19
[wip] Add an option for Caffe2 to link with local protobuf.

Yangqing:
Continuation of the protobuf hiding work following #2302 . Starting a PR to test build.

PR #2326, opened 2018-03-20
Add hidden visibility to protobuf build when we do custom protobuf

Yangqing:
(continuation of #2306 )


cmake/ProtoBuf.cmake
```diff
@@ -160,6 +174,10 @@ function(caffe2_protobuf_generate_cpp_py srcs_var hdrs_var python_var)
       COMMAND ${CMAKE_COMMAND} -E make_directory "${CMAKE_CURRENT_BINARY_DIR}"
       COMMAND ${CAFFE2_PROTOC_EXECUTABLE} -I${PROJECT_SOURCE_DIR} --cpp_out=${DLLEXPORT_STR}${PROJECT_BINARY_DIR} ${abs_fil}
       COMMAND ${CAFFE2_PROTOC_EXECUTABLE} -I${PROJECT_SOURCE_DIR} --python_out "${PROJECT_BINARY_DIR}" ${abs_fil}
+      COMMAND sed -i "s/::google::protobuf::internal::GetEmptyStringAlreadyInited/GetEmptyStringAlreadyInited/" "${CMAKE_CURRENT_BINARY_DIR}/${fil_we}.pb.h"
```
yinghai, line 177: Do we have `sed` in windows?

Yangqing: Actually not, so @orionr is taking a shot over there. We will probably put windows at a lower priority.

orionr: Yup - already got that figured out at https://github.com/caffe2/caffe2/pull/2384 . Unfortunately other things are failing. :) But, yes, Windows will be a lower priority for static linking.

orionr:
This is now rolled into https://github.com/caffe2/caffe2/pull/2384

PR #2332, opened 2018-03-20
Limit the number of LOG(INFO) for unavailable engine to 64

jspark1105:
Inlining what glog's LOG_FIRST_N does because not every platform has glog.

### DIFF
--- a/caffe2/core/operator.cc
+++ b/caffe2/core/operator.cc
@@ -175,8 +175,12 @@ unique_ptr<OperatorBase> _CreateOperator(
     }
   }
   if (operator_def.engine().size() && !VLOG_IS_ON(1)) {
-    LOG(INFO) << "Engine " << operator_def.engine()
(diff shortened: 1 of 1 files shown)


caffe2/core/operator.cc
```diff
@@ -175,8 +175,12 @@ unique_ptr<OperatorBase> _CreateOperator(
     }
   }
   if (operator_def.engine().size() && !VLOG_IS_ON(1)) {
-    LOG(INFO) << "Engine " << operator_def.engine()
-              << " is not available for operator " << op_type << ".";
+    static int log_occurrences = 0;
```
Maratyszcza, line 178: This is not thread-safe!

Maratyszcza: Nevermind, we can tolerate race condition here

Maratyszcza (approved):

PR #2334, opened 2018-03-20
[debug_info] Revert #2008

bwasti:


kuttas:
Does this change also need to revert the commit that added the 'CAFFE2_DEBUG' env var? I don't see that in the change to core.py
https://github.com/caffe2/caffe2/commit/efeae29f10b09f9f848889cc95f50c1aa1949385

PR #2348, opened 2018-03-21
Remove more protobuf APIs.

Yangqing:
This is continuing work to remove protobuf API from Caffe2 - we should only use caffe2 specific APIs. This PR hides ShutdownProtobufLibrary and TextFormat::ParseFromString().


orionr (approved):
Nice!

pjh5 (approved):

PR #2351, opened 2018-03-21
Log NNPACK profile to std::cout instead of LOG(INFO)

Maratyszcza:
Similar to #2333, but for NNPACK bindings

### DIFF
--- a/caffe2/share/contrib/nnpack/conv_op.cc
+++ b/caffe2/share/contrib/nnpack/conv_op.cc
@@ -14,6 +14,9 @@
  * limitations under the License.
  */
 
+
+#include <iostream>
+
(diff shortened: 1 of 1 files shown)


Yangqing:
Wait, for core libraries we never want to do std::cout because this will print stuff during execution time, so I thin we should keep LOG(INFO) in this case.

Maratyszcza:
@Yangqing It will log only if you run the binary with `--caffe2_profile_nnpack`, which is off by default, and practically only used for performance analysis & tuning.

Yangqing:
Ah got it, makes sense.

Yangqing (approved):

PR #2358, opened 2018-03-21
Test hidden visibility for protobuf builds

orionr:


Yangqing:
I assume this is superceded by #2384 ?

PR #2362, opened 2018-03-22
Remove Python onnx-caffe2 conversion code

yinghai:
Now we have two sets of onnx-caffe2 conversion code, one in C++, the other in Python, which is not good. This PR merge the truth to C++ implementation with Python binding. And we will only have one code path to maintain. Note that for ops are not supported in C++ currently, we fall back to the Python implementation.


yinghai:
@caffe2bot retest this please

caffe2/onnx/backend.cc
```diff
@@ -971,24 +972,27 @@ Caffe2Ops Caffe2Backend::ConvertNode(const std::string& node_str, int opset_vers
   return OnnxNodeToCaffe2Ops(init_model, pred_model, &onnx_node, opset_version);
 }
 
-Caffe2Ops Caffe2Backend::OnnxNodeToCaffe2Ops(const ModelProto &init_model,
-                                            const ModelProto &pred_model,
-                                            OnnxNode* onnx_node,
-                                            int opset_version) {
+Caffe2Ops Caffe2Backend::OnnxNodeToCaffe2Ops(
+    const ModelProto& init_model,
```
houseroad, line 976: Do we really need init_model and pred_model here? I guess later RNN may need, probably for now, we can remove it, Since you don't even use it.

caffe2/onnx/backend.h
```diff
@@ -133,80 +129,33 @@ class Caffe2Backend {
       const TensorProto& onnx_tensor,
       const std::string& name = "");
 
-  Caffe2Ops CommonOnnxNodeToCaffe2Ops(
-      const ModelProto& init_model,
-      const ModelProto& pred_model,
-      OnnxNode* onnx_node,
-      int opset_version);
+  Caffe2Ops CommonOnnxNodeToCaffe2Ops(OnnxNode* onnx_node, int opset_version);
```
houseroad, line 135: This looks much better.

caffe2/python/onnx/backend.py
```diff
@@ -1018,7 +726,7 @@ def prepare(cls, model, device='CPU', **kwargs):
         # Build the C++ backend
         # TODO: build a predictor that supports GPU
         #       And for RNN nets, we need to avoid adding init_net
-        if device == 'CPU' and not rnn_nodes:
+        if False: #device == 'CPU' and not rnn_nodes:
```
houseroad: Is this too hacky? Do we want to directly remove this branch?

houseroad (commented):
ConvertNode is very confusing. Looks like it returns two vectors of ops, one for init ops, and another for pred ops...

caffe2/python/onnx/backend.py
yinghai: I forgot to remove this debug string...

caffe2/onnx/backend.cc
yinghai: For now we don't but generally speaking if we want to support RNN ops, we need it.

yinghai:
> ConvertNode is very confusing. Looks like it returns two vectors of ops, one for init ops, and another for pred ops...

Yes. That's to unify the interface for RNN conversion (which we don't have in the C++ side yet).

yinghai:
@onnxbot restest this please

yinghai:
@onnxbot retest this please

houseroad (approved):
LGTM, thanks!

caffe2/python/onnx/backend.py
```diff
@@ -1036,7 +744,7 @@ def prepare(cls, model, device='CPU', **kwargs):
             cbackend = C.Caffe2Backend()
             rep = cbackend.prepare(model.SerializeToString(), device, c2_rnn_ops)
             # For testing
-            # Dump the net descritpions to file for comparison with the Python ones
+            # Dump the net descripions to file for comparison with the Python ones
```
houseroad: typo?

yinghai: Cool. I fixed a typo with a typo!

PR #2366, opened 2018-03-22
Add InheritOnnxSchema property to c2 op schema

yinghai:
This PR is a split off of https://github.com/caffe2/caffe2/pull/2248. It adds `InheritOnnxSchema` property to c2 op schema and removes the onnx conversion function.


houseroad (approved):
Add the ONNX mapping into Caffe2 schema. Looks good to me. Thanks!

PR #2367, opened 2018-03-22
Add argmax and argmin ops

xiaomengy:


houseroad:
Looks like this PR also touches TopK code.

PR #2341, opened 2018-03-20
[nomnigraph] Move nomnigraph to OSS

bwasti:
This commit moves nomnigraph into caffe2/core and sets up basic cmake files for building the library.

PR #2380, opened 2018-03-22
[nomnigraph] Move nomnigraph to OSS 

bwasti:
This commit moves nomnigraph into caffe2/core and sets up basic cmake files for building the library.


bwasti:
this is a clone of #2341 which was closed by an accidental misuse of the internal export script and github's automatic closing policy.  letting builds run now

PR #2384, opened 2018-03-22
Work on protobuf static and hidden

orionr:
Continuation of https://github.com/caffe2/caffe2/pull/2306 and based on Yangqing's PR at https://github.com/caffe2/caffe2/pull/2326


cmake/ProtoBuf.cmake
```diff
@@ -94,7 +108,7 @@ endif()
 # convention we will use SYSTEM inclusion path.
 get_target_property(__tmp protobuf::libprotobuf INTERFACE_INCLUDE_DIRECTORIES)
 message(STATUS "Caffe2 protobuf include directory: " ${__tmp})
-include_directories(SYSTEM ${__tmp})
+include_directories(BEFORE SYSTEM ${__tmp})
```
Yangqing, line 111: For the error message that you are encountering on cuda, I suspect this is the cause, but I don't know the underlying reason. Maybe this is related:

https://github.com/facebookresearch/Detectron/blob/master/lib/CMakeLists.txt#L37-L39

orionr: Might have been caused by my use of

`target_link_libraries(caffe2 PUBLIC protobuf::libprotobuf)` 

instead of 

`target_link_libraries(caffe2 INTERFACE protobuf::libprotobuf)`

like you were, so I've switched it back. Also added the use of the FindCUDA fix, since I didn't think that was required, but looks like it might be. Just looking to get green builds and tests first across the board.

cmake/ProtoBuf.cmake
```diff
@@ -152,22 +166,44 @@ function(caffe2_protobuf_generate_cpp_py srcs_var hdrs_var python_var)
...
+               "${CMAKE_CURRENT_BINARY_DIR}/${fil_we}.pb.h"
+               "${CMAKE_CURRENT_BINARY_DIR}/${fil_we}_pb2.py"
+        WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
+        COMMAND ${CMAKE_COMMAND} -E make_directory "${CMAKE_CURRENT_BINARY_DIR}"
+        COMMAND ${CAFFE2_PROTOC_EXECUTABLE} -I${PROJECT_SOURCE_DIR} --cpp_out=${DLLEXPORT_STR}${PROJECT_BINARY_DIR} ${abs_fil}
+        COMMAND ${CAFFE2_PROTOC_EXECUTABLE} -I${PROJECT_SOURCE_DIR} --python_out "${PROJECT_BINARY_DIR}" ${abs_fil}
+
+        # If we remove all reference to these pb.h files from external
+        # libraries and binaries this rewrite can be removed.
+        COMMAND sed -i.bk -e "s/::google::protobuf::internal::GetEmptyStringAlreadyInited/GetEmptyStringAlreadyInited/g" "${CMAKE_CURRENT_BINARY_DIR}/${fil_we}.pb.h"
```
yinghai: Just curious why do we have this special treatment of `::google::protobuf::internal::GetEmptyStringAlreadyInited`?

orionr: Good question. It's because it uses a global object to check empty strings. Calling that across library boundaries will fail and create seg faults, so we need to isolate it to one .so file by wrapping it.

Yangqing:
We are close! :)

orionr:
Looks like there's a protobuf compatibility error that starts with ONNX. Not sure why this is an issue in conda specifically, but I'll talk with @pjh5 tomorrow.

Yangqing:
Conda one seems to be just inclusion path error - I think it's probably some flags that we need to pass.

pjh5:
On conda, the "/Applications/Xcode9.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: ../../../lib/libprotobuf.a(atomicops_internals_x86_msvc.cc.o) has no symbols" errors on Mac seem to be https://stackoverflow.com/questions/23603355/ranlib-ar-in-mac-os-gives-me-has-no-symbols-warning , where Windows specific protobuf files are empty on Mac and the static linker doesn't like that

orionr:
Thanks to @pjh5 for the suggested changes above. We'll see how the tests go and then we'll need to decide when we'll turn this static linking on. For binary installs, but maybe not for default source builds and installs?

orionr:
I want to land these changes, so I'm going to (a) roll back the conda changes and (b) disable static linking for now. We can look at enabling it again in some cases next week, but hopefully we can get this code into master.

pjh5 (approved):

PR #2385, opened 2018-03-22
Update speed_benchmark binary

Maratyszcza:
- Support specifying type (float or uint8_t) for inputs
- Create input blobs if they don't exist

### DIFF
--- a/binaries/speed_benchmark.cc
+++ b/binaries/speed_benchmark.cc
@@ -49,6 +49,9 @@ CAFFE2_DEFINE_string(
     "semicolon to separate the dimension of different "
     "tensors.");
 CAFFE2_DEFINE_string(
+    input_type,
+    "", "Input type (uint8_t/float)");
+CAFFE2_DEFINE_string(
     output,
     "",
     "Output that should be dumped after the execution "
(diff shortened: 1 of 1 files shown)


binaries/speed_benchmark.cc
```diff
@@ -99,23 +102,44 @@ int main(int argc, char** argv) {
...
           input_dims.push_back(caffe2::stoi(s));
         }
-        caffe2::TensorCPU* tensor =
-            workspace->GetBlob(input_names[i])->GetMutable<caffe2::TensorCPU>();
+        caffe2::Blob* blob = workspace->GetBlob(input_names[i]);
+        if (blob == nullptr) {
+          blob = workspace->CreateBlob(input_names[i]);
+        }
+        caffe2::TensorCPU* tensor = blob->GetMutable<caffe2::TensorCPU>();
+        CHECK_NOTNULL(tensor);
```
bddppq, line 134: I'm curious why do we have `CHECK_NOTNULL`? Doesn't `CAFFE_ENFORCE` cover it?

binaries/speed_benchmark.cc
```diff
@@ -99,23 +102,44 @@ int main(int argc, char** argv) {
         CAFFE_ENFORCE(caffe2::ReadProtoFromFile(input_files[i], &blob_proto));
         workspace->CreateBlob(input_names[i])->Deserialize(blob_proto);
       }
-    } else if (caffe2::FLAGS_input_dims.size()) {
+    } else if (caffe2::FLAGS_input_dims.size() || caffe2::FLAGS_input_type.size()) {
+      CAFFE_ENFORCE_NE(0, caffe2::FLAGS_input_dims.size(),
+          "Input dims must be specified when input tensors are used.");
+      CAFFE_ENFORCE_NE(0, caffe2::FLAGS_input_type.size(),
```
bddppq, line 108: nit: consider allowing user to not specify types and default to `float `?

bddppq (approved):
LG

binaries/speed_benchmark.cc
Maratyszcza: Many models use `uint8_t` input with `PackedInt8BGRANHWCToNCHWCStylizerPreprocess` as the first operation, so I'd rather have users think about the data type explicitly.

PR #2394, opened 2018-03-23
Remove legacy onnx optimizer tests

bddppq:
has been moved to here https://github.com/onnx/onnx/blob/5f69c37628002efa8c03d70d652db03c9d5ffca7/onnx/test/optimizer_test.py


houseroad (approved):
LGTM

PR #2398, opened 2018-03-23
Commit the part of protobuf that is already safe to patch

Yangqing:
This helps the effort to hide protobuf visibility. Kicking a contbuild. Instead of sed, cmake would make it more platform independent.


orionr:
This is now rolled into https://github.com/caffe2/caffe2/pull/2384

PR #2414, opened 2018-03-23
[nomnigraph] Expose transformations to python

bwasti:
Adding a python interface to the transformations


Maratyszcza (approved):
Thanks, looks good. I've also seen NNPACK enabled on layers where it is suboptimal, so would be nice to not only add NNPACK where it helps, but also remove it in other cases.

caffe2/opt/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+# ---[ CPU files.
```
yinghai, line 1: Is this directory for all the transformation code?

orionr:

PR #2421, opened 2018-03-26
Initialize cpuinfo in the thread pool

Maratyszcza:
Thread pool called cpuinfo_get_processors_count() without initializing cpuinfo. Only by chance it didn't regress performance: threadpool is initialized after NNPACK, and NNPACK initializes cpuinfo itself.

This commit also updates cpuinfo to a version that aborts with a fatal error if its used uninitialized.


ezyang:
Appveyor failure looks legit

Maratyszcza:
@caffe2bot retest this please

ajtulloch (approved):

PR #2426, opened 2018-03-26
Strip down onnx to only pb definitions in mobile build

bddppq:


caffe2/onnx/backend.cc
```diff
@@ -941,9 +939,14 @@ void Caffe2Backend::OnnxToCaffe2(
     const std::vector<Caffe2Ops>& extras) {
   auto device_option = GetDeviceOption(Device(device));
 
+#if !CAFFE2_MOBILE
```
anderspapitto, line 942: well, rather than having crippled functionality, should we just say that OnnxToCaffe2 does not exist at all on mobile?

anderspapitto (approved):

yinghai:
Another note is that we don't need to build with ONNX at all. It's also an option. :)

PR #2428, opened 2018-03-26
Moving conda/ to caffe2/conda

pjh5:


orionr (approved):

pjh5:
Ignoring failures because they are caused by registry problems and unrelated

PR #2435, opened 2018-03-26
Move doc generation code into docs/caffe2.

orionr:
Moving docs generation code up a level.

FYI @ezyang and @pjh5


pjh5 (approved):

orionr:
The failure doesn't look related, so merging.

PR #2437, opened 2018-03-26
Make error messages in net_dag more clear

kuttas:
The logging in net_dag.cc was a little misleading because it conflated the first operator in an execution chain with the operator actually being run. This change attempts to make the error a little more clear. The current interface between DAGNetBase and its child classes (which implement the virtual "RunAt()") doesn't make it clear which specific operator failed except by deep inspection of the exception thrown (if any is thrown). If an operator *does* thrown an exception, the exception message will indicate which exact operator caused it, but if it merely returns 'false' from Run, then the various implementations of DAGNetBase will swallow the failure and return after the entire chain is executed. Therefore, all we can deduce from DAGNetBase::WorkerFunction() is the operator at the head of the execution chain that failed, so we print that out. If an exception was thrown, we also print the exception message which narrows in on the specific operator that caused the error.

### DIFF
--- a/caffe2/core/net_dag.cc
+++ b/caffe2/core/net_dag.cc
@@ -193,12 +193,12 @@ void DAGNetBase::HandleException(
       operator_nodes_[operator_idx].operator_->debug_def().name();
   const std::string& operator_type =
       operator_nodes_[operator_idx].operator_->debug_def().type();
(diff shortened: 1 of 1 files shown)


romain-intel (approved):
LGTM

PR #2438, opened 2018-03-26
Make data_parallel_model Synchronize match behavior of _AllReduceBlobs, _SyncAllParams for single host case

kuttas:
Instead of asserting when calling this function on a single shard use case, early-out if there is only 1 shard. This makes the calling code cleaner because we do not need to branch on the number of shards or the nullity of model._rendezvous; the same API call works for both single host and multi-host cases.


kuttas:
This change was already approved in D7388995.

PR #2454, opened 2018-03-27
Quick and dirty raw value substitution from zip file import for ONNX

jamesr66a:
Since the backend implementation is split between C++ and python and is in flux, I implemented substitution of raw values from an ONNX zip archive as a separate pass that substitutes them in before passing to `prepare()`. This is slower but it works for now

### DIFF
--- a/caffe2/python/onnx/backend.py
+++ b/caffe2/python/onnx/backend.py
@@ -13,6 +13,7 @@ from __future__ import unicode_literals
 import os
 import collections
 from subprocess import Popen, PIPE
+import zipfile
 
 import caffe2
 from caffe2.python import core, workspace, rnn_cell, gru_cell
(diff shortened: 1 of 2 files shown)


yinghai (approved):
LGTM. What's the long term plan for this BTW?

jamesr66a:
@caffe2bot retest this please

jamesr66a:
@yinghai  we should probably sync up on the long-term plan for this. I'd like to integrate this more deeply in to the code (and into the caffe2 runtime itself) s.t. we can take full advantage of the perf benefits this format gives us

caffe2/python/onnx/backend.py
```diff
@@ -648,6 +649,35 @@ def make_gru(direction_offset):
...
+    @classmethod
+    def _substitute_raw_value(cls, tp, raw_values_dict):
+        if tp.HasField('raw_data') and tp.raw_data == bytes(b'__EXTERNAL'):
+            if tp.name not in raw_values_dict:
+                raise RuntimeError('TensorProto for value {} referenced raw data but it was not found!'.format(tp.name))
+            else:
+                tp.raw_data = raw_values_dict[tp.name]
+
+    @classmethod
+    def _visit_and_substitute_raw_values(cls, nodes, raw_values_dict):
```
dzhulgakov, line 661: oh wow, we even substitute it in Constant nodes, that's fancy

caffe2/python/onnx/backend.py
```diff
@@ -673,6 +703,21 @@ def optimize_onnx(input, init=False, predict=False):
         out = onnx.optimizer.optimize(input, passes)
         return out
 
+    @classmethod
+    def prepare_zip_archive(cls, file, device='CPU', **kwargs):
```
dzhulgakov, line 707: we should probably move this logic (everything but prepare) to onnx.load() instead. But since it's 'quick-and-hacky' - still - maybe just have a standalone function load_zip_archive?

PR #2468, opened 2018-03-28
Move setup.py and version into caffe2 folder.

orionr:
* Prep for repository movement.


pjh5 (approved):

caffe2/setup.py
```diff
@@ -20,8 +20,8 @@
 import sys
 from textwrap import dedent
 
-TOP_DIR = os.path.realpath(os.path.dirname(__file__))
-SRC_DIR = os.path.join(TOP_DIR, 'caffe2')
+SRC_DIR = os.path.realpath(os.path.dirname(__file__))
+TOP_DIR = os.path.join(SRC_DIR, '..')
```
bddppq: 1. '..' is not truly cross-platform, there `os.pardir`
2. We should resolve the realpath instead of leaving `..` in it.

bddppq (changes requested):
`find_packages` should also need changes

bddppq:
To test whether this will work, you should create a PR in onnx-fb-universe repo with `install.sh` modified and `repos/caffe2` points to this caffe2 PR's commit.

orionr:
Good catch on `find_packages()` - changing to `setuptools.find_packages(TOP_DIR)`

orionr:
Thanks @bddppq - universe PR up at https://github.com/onnxbot/onnx-fb-universe/pull/1358

pjh5 (approved):

ezyang:
Comments should call out that using absolute paths is important (relative paths don't work.) Does this work with `python setup.py develop`?

orionr:
We decided to rename instead - closing.

PR #2470, opened 2018-03-28
Correctly handle missing hidden state in ONNX import

anderspapitto:
This goes along with https://github.com/pytorch/pytorch/pull/6080

When the hidden state is not provided, use the existing support for
initializing it in caffe2


anderspapitto:
this doesn't work, because the caffe2 support isn't capable of handling dynamic batch sizes

PR #2478, opened 2018-03-29
Checks a default op mapping onnx -> caffe2 bears the same arguments

huitseeker:
When adding an extra arg to an input ONNX op, if it's not supported in Caffe2, the exporter would just silently pass it to NetDef and ignore it in the implementation. It's pretty error-prone. Caffe2 also has an OpSchema description and we can enforce that all arguments explicitly appear in schema or listed explicitly in Caffe2. 

- make arguments check work as universally as possible in the c2 conversion
- Add test for argument-checking


huitseeker:
Removing in favor of https://github.com/pytorch/pytorch/pull/6317

PR #2486, opened 2018-03-30
strip some python dependencies

bddppq:
those are only used in non-core part


orionr (approved):

houseroad (commented):
No setuptools !?

bddppq:
@houseroad it's not counted as install dependencies of caffe2, without it you won't even be able to run our setup.py :-)

pjh5 (commented):
You can axe matplotlib too

houseroad:
I would like to include setuptools... to me, the experience of using setup.py is way better than pure cmake/make installation.

I remember half a year ago, I had suffered from cmake/make installation... with support of setuptools, it was so easy to install caffe2, and even my skin got better. :-)

bddppq:
@houseroad it's unrelated to setup.py vs. cmake. my point is you don't need to specify setuptools as dependencies of caffe2 in setup.py, because in order to run our setup.py you must already have setuptools installed.

houseroad:
@bddppq cool, I got your point, thanks!

houseroad (approved):

PR #2491, opened 2018-04-02
simplify the calculation of top_diff_this_bin

xmyqsh:
simplify the calculation of top_diff_this_bin


orionr:
Hi @xmyqsh - do you mind opening a PR on https://github.com/pytorch/pytorch instead? The code has moved. Much appreciated.

PR #2495, opened 2018-04-02
Fix observer logic in RNN executor. Remove dynamic casts

mdschatz:
Test case failed that triggered PR#2427 as a bandaid.  As part of this, we discussed redesigning how RNN-Capable observers were treated and landed on specifying a boolean for this purpose.  Updated code to support this.


houseroad (approved):
LGTM, let's move it to pytorch repo :-)

PR #2542, opened 2018-08-06
Update .gitmodules

yanxiangyi:
Update source of the `eigen` repo.


orionr:
Hi @yanxiangyi - thank you for the update. However, we've actually moved all of the source code over to https://github.com/pytorch/pytorch. Things should be updated over there.

PR #2568, opened 2018-11-21
Add code of conduct

JoelMarcey:


orionr:
Actually, we've removed all code on caffe2/caffe2 so people go to pytorch/pytorch and don't accidentally install this. We really only have it here as historical info since we can't do a real redirect.

JoelMarcey:
Ugh - I must have been working on an outdated branch. Sorry about that. Thanks for catching that.

